PR #2, opened 2021-07-08
fix make test-dependencies by only perform 'mv' if kubebuilder doesn't exist already

mikeshng:
Addresses https://github.com/open-cluster-management-io/community/issues/58

Signed-off-by: Mike Ng <ming@redhat.com>


dhaiducek (approved):
I didn't know there was a flag for this! Nice!

PR #3, opened 2021-07-08
Fix the deploy make target

mprahl:
See each individual commit for an explanation of each change.

Fixes https://github.com/open-cluster-management-io/community/issues/61.


dhaiducek (commented):
Thanks for investigating this! I don't even see the `policy.open-cluster-management.io_v1alpha1_configurationpolicy_cr.yaml` file in use anywhere (and it has annotations referencing IBM). Do you think we could just delete the file? I'm not sure we need to be deploying a policy alongside the controller...

mprahl:
@dhaiducek sure, I'll go ahead and delete it. Also, was the intention to also include the `deploy/crds/v1` directory?

dhaiducek:
Good catch--yeah it looks like we want the `v1` (that'd match our instructions on the [community website](https://open-cluster-management.io/getting-started/integration/policy-controllers/)).

mprahl:
Thanks for the review @dhaiducek! The last push should address your feedback.

dhaiducek (approved):
Nice--thanks, @mprahl!

PR #5, opened 2021-07-12
Download the build harness bootstrap Makefile only when it's not present

mprahl:
This will reduce the number of downloads of the file and reduce the chances of hitting GitHub's rate limit.

This resolves part of https://github.com/open-cluster-management-io/community/issues/57.


JustinKuli (approved):
/lgtm

PR #16, opened 2021-11-08
Merge updates from ocm-plus repo

JustinKuli:
Brings in commits up to the current "release-2.4" branch from the other repo.


cmd/manager/main.go
```diff
@@ -142,12 +146,22 @@ func main() {
 				os.Exit(1)
 			}
 		} else {
+
 			log.Info("Starting lease controller to report status")
 			leaseUpdater := lease.NewLeaseUpdater(
 				generatedClient,
-				"config-policy-controller",
+				"policy-controller",
```
dhaiducek, line 153: This change was reverted in
- https://github.com/open-cluster-management/config-policy-controller/pull/175

So maybe pull some more PRs from midstream in? I'm not 100% sure of the effect here, but keeping `config-policy-controller` seems clearer to me...

dhaiducek (commented):
Wow! Thanks for tackling this! I looked pretty quickly and noted one thing that stuck out. I'm hoping the tests are capturing anything I might have missed.

cmd/manager/main.go
JustinKuli: Nice catch

dhaiducek (approved):

PR #34, opened 2022-06-14
Move to go 1.18

gparvin:
Refs:
 - https://github.com/stolostron/backlog/issues/22476

Signed-off-by: Gus Parvin <gparvin@redhat.com>


mprahl (approved):

dhaiducek:
The branch protection updates run periodically (every 6 hours or so), so the status list should update when the Prow job runs. In the meantime, the `KinD tests` workflow should be updated to `1.18`:
- https://github.com/open-cluster-management-io/config-policy-controller/blob/6b86f4a87e52b927ccb0b4cc78ed6960df5d01e0/.github/workflows/kind.yml#L26

mprahl (approved):

PR #36, opened 2022-06-22
Evaluate policies concurrently

mprahl:
This adds concurrency to evaluating configuration policies. This
defaults to two Go routines to provide a performance increase while not
overwhelming the Kubernetes API in resource constrained deployments.
Additionally, it means that the E2E tests will be testing the
concurrency on every test, which is desired.

The number of Go routines can be configured with the
`--evaluation-concurrency` flag.

Relates:
https://github.com/stolostron/backlog/issues/23546

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -105,6 +105,8 @@ type ConfigurationPolicyReconciler struct {
 	// that reads objects from the cache and writes to the apiserver
 	client.Client
 	DecryptionConcurrency uint8
+	// Determines the number of Go routines that can evaluate policies concurrently.
+	EvaluationConcurrency uint8
 	Scheme                *runtime.Scheme
 	Recorder              record.EventRecorder
 	discoveryInfo
@@ -158,9 +160,20 @@ func (r *ConfigurationPolicyReconciler) PeriodicallyExecConfigPolicies(freq uint
 			skipLoop = true
 		}
 
+		// This is done every loop cycle since the channel needs to be variable in size to account for the number of
+		// policies changing.
+		policyQueue := make(chan *policyv1.ConfigurationPolicy, len(policiesList.Items))
+		var wg sync.WaitGroup
+
 		if !skipLoop {
 			log.Info("Processing the policies", "count", len(policiesList.Items))
-			// flattenedpolicylist only contains 1 of each policy instance
+
+			for i := 0; i < int(r.EvaluationConcurrency); i++ {
+				wg.Add(1)
+
+				go r.handlePolicyWorker(policyQueue, &wg)
+			}
+
 			for i := range policiesList.Items {
 				policy := policiesList.Items[i]
 				if !shouldEvaluatePolicy(&policy) {
@@ -168,9 +181,12 @@ func (r *ConfigurationPolicyReconciler) PeriodicallyExecConfigPolicies(freq uint
 				}
 
 				// handle each template in each policy
-				r.handleObjectTemplates(policy)
+				policyQueue <- &policy
 			}
 		}
+
+		close(policyQueue)
+		wg.Wait()
 
 		// making sure that if processing is > freq we don't sleep
 		// if freq > processing we sleep for the remaining duration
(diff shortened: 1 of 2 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -158,20 +160,47 @@ func (r *ConfigurationPolicyReconciler) PeriodicallyExecConfigPolicies(freq uint
 			skipLoop = true
 		}
 
+		scheduledPolicyCount := 0
+
+		// This is done every loop cycle since the channel needs to be variable in size to account for the number of
+		// policies changing.
+		policyQueue := make(chan *policyv1.ConfigurationPolicy, len(policiesList.Items))
+		doneQueue := make(chan struct{}, len(policiesList.Items))
```
JustinKuli, line 168: Did you consider using a [WaitGroup](https://pkg.go.dev/sync#WaitGroup) for this? Counting the number of policies "done" seems wasteful.

controllers/configurationpolicy_controller.go
```diff
@@ -188,6 +217,16 @@ func (r *ConfigurationPolicyReconciler) PeriodicallyExecConfigPolicies(freq uint
 	}
 }
 
+// handlePolicyWorker is meant to be used as a Go routine that wraps handleObjectTemplates.
+func (r *ConfigurationPolicyReconciler) handlePolicyWorker(
+	policyQueue <-chan *policyv1.ConfigurationPolicy, done chan<- struct{},
+) {
+	for policy := range policyQueue {
+		r.handleObjectTemplates(*policy)
```
JustinKuli, line 225: If this panics, the done signal is never sent. This would be another reason to use a WaitGroup - idiomatically you call the `Done` in a `defer` statement to ensure it is always called.

controllers/configurationpolicy_controller.go
```diff
@@ -158,20 +160,47 @@ func (r *ConfigurationPolicyReconciler) PeriodicallyExecConfigPolicies(freq uint
...
 		}
 
+		scheduledPolicyCount := 0
+
+		// This is done every loop cycle since the channel needs to be variable in size to account for the number of
+		// policies changing.
+		policyQueue := make(chan *policyv1.ConfigurationPolicy, len(policiesList.Items))
+		doneQueue := make(chan struct{}, len(policiesList.Items))
+
+		for i := 0; i < int(r.EvaluationConcurrency); i++ {
```
JustinKuli, line 170: I think we need to validate that `r.EvaluationConcurrency` is not 0, otherwise bad things could happen here. A user could think setting it to `0` would maximize the concurrency.

JustinKuli (changes requested):

controllers/configurationpolicy_controller.go
mprahl: I had forgotten about this. Thanks for the suggestion! That's much better.

controllers/configurationpolicy_controller.go
mprahl: We can't use `defer` in this case since we are looping over a channel and thus we need to call `wg.Done()` after **each** `r.handleObjectTemplates(*policy)` call.

I also want a panic in one of the worker Go routines to bring down the entire controller which is keeping the current behavior so a deferred call to `wg.Done()` wouldn't affect that since the entire controller is shutting down. We could reconsider this and start a new worker Go routine upon recovering a panic from one of the Go routine workers so that one policy that hits a nasty bug doesn't prevent the rest from being evaluated, but I think that's out of scope for this PR.

JustinKuli: You're right, a panic in one of them does bring the whole thing down, I mistakenly thought it might lead to a deadlock.

controllers/configurationpolicy_controller.go
```diff
@@ -188,6 +203,16 @@ func (r *ConfigurationPolicyReconciler) PeriodicallyExecConfigPolicies(freq uint
 	}
 }
 
+// handlePolicyWorker is meant to be used as a Go routine that wraps handleObjectTemplates.
+func (r *ConfigurationPolicyReconciler) handlePolicyWorker(
+	policyQueue <-chan *policyv1.ConfigurationPolicy, wg *sync.WaitGroup,
+) {
+	for policy := range policyQueue {
+		r.handleObjectTemplates(*policy)
+		wg.Done()
```
JustinKuli, line 212: This works, but it's still more counting than necessary. I was thinking the WaitGroup would count the workers, not the work items. So the `wg.Done()` would go after the loop here, and the end of the loop occurs when the channel was closed (after all the policies are added to it).

mprahl:
@JustinKuli is this good to merge now?

JustinKuli (approved):

PR #41, opened 2022-06-27
Add `LabelSelector` for `NamespaceSelector`

dhaiducek:
Addresses:
- https://github.com/open-cluster-management-io/community/issues/155
- https://github.com/stolostron/backlog/issues/23523

### DIFF
--- a/test/e2e/case15_event_format_test.go
+++ b/test/e2e/case15_event_format_test.go
@@ -239,5 +239,12 @@ var _ = Describe("Testing compliance event formatting", func() {
 		}
 
 		deleteConfigPolicies(configPolicies)
+
+		utils.Kubectl("delete", "-f", case15PodForNonComplianceYaml)
+		err := clientManaged.CoreV1().Pods("default").Delete(
+			context.TODO(), "case15-becomescompliant", metav1.DeleteOptions{})
+		if !k8serrors.IsNotFound(err) {
+			Expect(err).To(BeNil())
+		}
 	})
 })


test/e2e/case19_ns_selector_test.go
```diff
@@ -0,0 +1,124 @@
...
+	patches := []struct {
+		description string
+		patch       string
+		message     string
+	}{
+		{"no namespaceSelector specified", "{}", "namespaced object has no namespace specified"},
+		{
+			"a non-matching LabelSelector",
+			"{\"matchLabels\":{\"name\":\"not-a-namespace\"}}",
+			"namespaced object has no namespace specified",
```
JustinKuli, line 38: This seems like an odd message. I wonder if we need a new message for the case where no namespaces were matched by the selector - that seems different than "not specified". From this message, a user might expect they have to add the namespace to the object template, which is not correct.

test/e2e/case19_ns_selector_test.go
```diff
@@ -0,0 +1,124 @@
...
+		Eventually(func() interface{} {
+			managedPlc := utils.GetWithTimeout(clientManagedDynamic, gvrConfigPolicy,
+				case19PolicyName, testNamespace, true, defaultTimeoutSeconds)
+
+			return utils.GetStatusMessage(managedPlc)
+		}, defaultTimeoutSeconds, 1).Should(ContainSubstring(
+			"[configmap-selector-e2e] in namespace case19-6-e2e missing"))
+	})
+
+	It("Cleans up", func() {
```
JustinKuli, line 117: If this was implemented as an `AfterAll`, then it would always cleanup after itself. The other tests here are ordered anyway, so it would be good to add. https://onsi.github.io/ginkgo/#setup-in-ordered-containers-beforeall-and-afterall

JustinKuli (commented):
Overall really great! I appreciate the thorough test cases and examples.

api/v1/configurationpolicy_types.go
```diff
@@ -65,10 +65,15 @@ type Condition struct {
...
 }
 
-// Target defines the list of namespaces to include/exclude
-type Target struct {
+// NamespaceSelector defines the list of namespaces to include/exclude
+type NamespaceSelector struct {
+	// +kubebuilder:validation:MinItems=1
 	Include []NonEmptyString `json:"include,omitempty"`
+	// +kubebuilder:validation:MinItems=1
 	Exclude []NonEmptyString `json:"exclude,omitempty"`
```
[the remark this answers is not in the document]
dhaiducek: Correct. It's the kind of thing where "if it's provided, put something in there". I was trying to make it more transparent that an empty array here would be rejected outright.

Now that you mention it though, I do need to explore `omitempty` a bit since I'm wondering if `matchLabels` is provided but empty might not show up, which could be confusing as a user looking back at the object.

controllers/configurationpolicy_controller_test.go
```diff
@@ -648,3 +650,383 @@ func TestShouldEvaluatePolicy(t *testing.T) {
...
+		}
+	}
+
+	// Set of testcases to run
+	tests := []struct {
+		testDescription string
+		selector        policyv1.NamespaceSelector
+		expected        []string
+		errMsg          string
+	}{
```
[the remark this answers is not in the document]
dhaiducek: Thanks for the link! I'll give it a shot!

controllers/configurationpolicy_controller.go
```diff
@@ -579,10 +567,31 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
+		if isNamespaced && len(relevantNamespaces) == 0 {
+			// If relevantNamespaces is empty, retrieve the namespaces based on filters in NamespaceSelector
+			selector := plc.Spec.NamespaceSelector
+			// If MatchLabels/MatchExpressions/Include were not provided, return no namespaces
+			if selector.MatchLabels == nil && selector.MatchExpressions == nil && len(selector.Include) == 0 {
+				log.Info("NamespaceSelector is empty. Skipping namespace retrieval.")
+			} else {
+				relevantNamespaces, err = getPolicyNamespaces(selector)
+				if err != nil {
+					log.Error(err, "Error filtering namespaces with provided NamespaceSelector", "index", indx)
```
mprahl, line 579: It'd be nice to log the invalid namespace selector as well.
```suggestion
					log.Error(err, "Error filtering namespaces with provided namespaceSelector", "index", indx, "namespaceSelector": fmt.Sprintf("#v", selector))
```

test/e2e/case19_ns_selector_test.go
JustinKuli: It shouldn't be necessary to update the message in this PR. I just wanted to flag it while it was more obvious.

controllers/configurationpolicy_controller_suite_test.go
```diff
@@ -24,7 +24,7 @@ var samplePolicy = policiesv1alpha1.ConfigurationPolicy{
 	},
 	Spec: policiesv1alpha1.ConfigurationPolicySpec{
 		Severity: "low",
-		NamespaceSelector: policiesv1alpha1.Target{
+		NamespaceSelector: policiesv1alpha1.NamespaceSelector{
```
mprahl, line 27: Optional: It'd be nice to rename the import of `policiesv1alpha1` to `policyv1`.

controllers/configurationpolicy_controller_test.go
```diff
@@ -648,3 +650,383 @@ func TestShouldEvaluatePolicy(t *testing.T) {
 		)
 	}
 }
+
+func TestGetPolicyNamespaces(t *testing.T) {
+	t.Parallel()
```
mprahl, line 655: This can't be run in parallel because of the call to `Initialize` which affects global variables in the package.

controllers/configurationpolicy_controller_test.go
```diff
@@ -648,3 +650,383 @@ func TestShouldEvaluatePolicy(t *testing.T) {
...
+				Labels: map[string]string{
+					"label" + idx: "works",
+					"all":         "namespaces",
+					"thisns":      "is_" + idx,
+				},
+			},
+		})
+	}
+
+	for _, ns := range namespaces {
```
mprahl, line 679: Optional: You could consolidate this loop with the previous one.

controllers/configurationpolicy_controller_test.go
```diff
@@ -648,3 +650,383 @@ func TestShouldEvaluatePolicy(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range tests {
+		test := test
+
+		t.Run(
+			test.testDescription,
+			func(t *testing.T) {
+				t.Parallel()
```
mprahl, line 822: Same as above.

controllers/configurationpolicy_controller_test.go
```diff
@@ -648,3 +650,383 @@ func TestShouldEvaluatePolicy(t *testing.T) {
...
+				Labels: map[string]string{
+					"label" + idx: "works",
+					"all":         "namespaces",
+					"thisns":      "is_" + idx,
+				},
+			},
+		})
+	}
+
+	for _, ns := range namespaces {
```
mprahl, line 864: Optional: You could consolidate this loop with the previous loop.

api/v1/configurationpolicy_types.go
```diff
@@ -65,10 +65,15 @@ type Condition struct {
...
+
+	MatchLabels      *map[string]string                 `json:"matchLabels,omitempty"`
```
mprahl, line 75: Could you please document `MatchLabels` so that the CRD documents this field?

api/v1/configurationpolicy_types.go
```diff
@@ -65,10 +65,15 @@ type Condition struct {
...
+	MatchExpressions *[]metav1.LabelSelectorRequirement `json:"matchExpressions,omitempty"`
```
mprahl, line 76: Setting `// +kubebuilder:validation:MinItems=1` would be nice for this.

pkg/common/pattern_util.go
```diff
@@ -4,134 +4,66 @@
...
+func Matches(
+	namespaces []string,
+	includeList []policyv1.NonEmptyString,
+	excludeList []policyv1.NonEmptyString,
+) ([]string, error) {
+	matchingNamespaces := make([]string, 0)
+
+	for _, namespace := range namespaces {
+		// If the includeList is empty include all namespaces for the NamespaceSelector LabelSelector
+		include := false || len(includeList) == 0
```
mprahl, line 23:
```suggestion
		include := len(includeList) == 0
```

mprahl (changes requested):

controllers/configurationpolicy_controller.go
dhaiducek: Made the change now, but making sure I'm not crazy. This should use a comma and `%`, correct? Like so?
```
"namespaceSelector", fmt.Sprintf("%#v", selector)
```

api/v1/configurationpolicy_types.go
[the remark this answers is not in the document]
JustinKuli: > An empty label selector matches all objects. A null label selector matches no objects

https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelector

controllers/configurationpolicy_controller.go
mprahl: Sorry, that was a typo. I mean `%v`.

api/v1/configurationpolicy_types.go
dhaiducek: True, but technically providing it empty returns all namespaces, so I don't think we necessarily want to prevent users from providing an empty MatchExpressions?

mprahl: Fair point!

controllers/configurationpolicy_controller_test.go
dhaiducek: Ah, I didn't think of that. Thanks! Though I hadn't been running into issues (perhaps because it only creates namespaces and then retrieves them)? Regardless, probably safer to keep them isolated.

api/v1/configurationpolicy_types.go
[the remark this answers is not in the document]
dhaiducek: Okay, I've poked at it just in case, and providing `matchLabels: {}` and `matchExpressions: []` does still appear in the object. Setting `matchLabels: null` deletes the key.

mprahl:
@dhaiducek could you please squash these commits?
* Add MatchLabels/MatchExpressions 
* Update unit tests for LabelSelector 
* Update E2E tests for LabelSelector

controllers/configurationpolicy_controller.go
```diff
@@ -579,10 +567,33 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
 		}
 
+		if isNamespaced && len(relevantNamespaces) == 0 {
+			// If relevantNamespaces is empty, retrieve the namespaces based on filters in NamespaceSelector
+			selector := plc.Spec.NamespaceSelector
+			// If MatchLabels/MatchExpressions/Include were not provided, return no namespaces
+			if selector.MatchLabels == nil && selector.MatchExpressions == nil && len(selector.Include) == 0 {
+				log.Info("NamespaceSelector is empty. Skipping namespace retrieval.")
+			} else {
+				relevantNamespaces, err = getPolicyNamespaces(selector)
```
mprahl, line 577: This is being called per object template array entry, but this can actually be called only once per `ConfigurationPolicy`. What I suggest is that you determine the namespaces once near the beginning of the `handleObjectTemplates` method as it was before. Optionally, you can only perform the query if at least one object template array entry is of a namespaced object and doesn't specify `metadata.namespace`.

Additionally, it'd be good to add the error to the policy status in the event the namespace selector is invalid. You can do this in a similar way to how the policy template resolve issues get bubbled up.

dhaiducek: Hopefully the latest changes aren't too sweeping. I've moved the loop that grabs the namespace/name/isNamespaced information up to the top so that the status can be updated if the `NamespaceSelector` fails (if the NamespaceSelector is valid but returns no namespaces, I currently let this fall through to the original "no namespaces" message--maybe it should fail there also?). In the process, the loop stores the namespace/name/isNamespaced information and uses it later in the function rather than re-retrieving it.

api/v1/configurationpolicy_types.go
```diff
@@ -120,8 +125,7 @@ func (e EvaluationInterval) GetNonCompliantInterval() (time.Duration, error) {
 type ConfigurationPolicySpec struct {
 	Severity           Severity           `json:"severity,omitempty"`          // low, medium, high
 	RemediationAction  RemediationAction  `json:"remediationAction,omitempty"` // enforce, inform
-	NamespaceSelector  Target             `json:"namespaceSelector,omitempty"`
-	LabelSelector      map[string]string  `json:"labelSelector,omitempty"`
```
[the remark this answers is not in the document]
dhaiducek: So, it looks like it might be in use for `cert-policy-controller` in order to select certificates by label. (Though even that isn't documented--I'll verify and open an issue to document it better if we need it.) 
That said, should the `LabelSelector` field be left in place here for now or write it off as being specific to `cert-policy-controller`? I'm leaning toward the latter.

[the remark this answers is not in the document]
JustinKuli: Since it's neither documented nor used by the configuration-policy-controller, I feel even better about just removing it. Leave it as specific to cert-policy. And we won't need to include it in the nucleus 😄 

I think it's also present in the iampolicy spec, but not used.

pkg/common/namespace_selection.go
```diff
@@ -1,48 +0,0 @@
...
-package common
-
-import (
-	"context"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-)
-
-// GetSelectedNamespaces returns the list of filtered namespaces according to the policy namespace selector
-func GetSelectedNamespaces(included, excluded, allNamespaces []string) []string {
```
dhaiducek, line 13: One other thing I noticed while poking at this code. I deleted `namespace_selection.go` since it wasn't being used, but maybe I should have migrated the functions there instead and deleted the functions from the controller? It'd be a straightforward change (and would make the controller file a bit shorter). Thoughts?

JustinKuli: +1 to smaller files.

api/v1/configurationpolicy_types.go
```diff
@@ -65,10 +66,40 @@ type Condition struct {
 	Message string `json:"message,omitempty" protobuf:"bytes,5,opt,name=message"`
 }
 
-// Target defines the list of namespaces to include/exclude
-type Target struct {
+// namespaceSelector defines the list of namespaces to include/exclude for objects defined in
+// spec.objectTemplates. All selector rules are ANDed.
+type NamespaceSelector struct {
+	// include is an array of filepath expressions to include namespaces by name.
```
mprahl, line 72: Could you also document that if it's empty but a label selector is specified, this acts as being set to `*`?

dhaiducek: Ah, good point. Yes, I'll add it.

controllers/configurationpolicy_controller.go
```diff
@@ -364,6 +361,104 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
+			if objName, ok := metadata["name"]; ok {
+				templateObjs[idx].name = objName.(string)
+			} else {
+				log.V(2).Info(
+					"metadata.name not found or could not be retrieved from the object template", "index", idx,
+				)
+			}
+		}
+
+		if queryNamespaces ||
```
mprahl, line 413: I find this easier to read:
```suggestion
		if !queryNamespaces &&
```

mprahl: You could also just remove the `queryNamespaces` check all together. I wasn't sure if you included that for a slight performance improvement.

controllers/configurationpolicy_controller.go
```diff
@@ -364,6 +361,104 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
 	parentStatusUpdateNeeded := false
 
+	// Collect template information for processing:
+	// - Object name, if specified
+	// - Object namespace, if specified
+	// - Whether object is namespaced
+	//
+	// Also determine whether a namespace query is needed for this policy:
+	//   If a namespaceSelector is present AND objects are namespaced without a namespace specified
+	templateObjs := make([]struct {
```
mprahl, line 371: Optional, consider moving this new block of code to a separate function that returns `templateObjs`. That would make it easier to read and potentially unit testable if that's needed.

controllers/configurationpolicy_controller.go
dhaiducek: Ah, right. That would be clearest. Yeah, I think I was but I think the performance impact would be minimal.

controllers/configurationpolicy_controller.go
dhaiducek: Good idea!

dhaiducek: I might chalk this up to technical debt. I started migrating it, but I found `GetDetails` which pretty much does the same thing but differenter and then wondering if I could combine them. There are already a bunch of functions in this gigantic file but moving the code to a different package would require passing the reconciler object over to it for the `isObjectNamespaced` function, which seemed heavy for a util function. I might poke at it tomorrow when I'm clearer headed, but in this case we might need to prioritize getting the code in so QE can start testing it and consider this again when ConfigurationPolicy is (hopefully!) refactored. Or is this simpler than I'm making it out to be?

mprahl: @dhaiducek I was thinking something like this (untested and undocumented and unlinted):
```diff
diff --git a/controllers/configurationpolicy_controller.go b/controllers/configurationpolicy_controller.go
index 949fb65..69bac0b 100644
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -335,32 +335,13 @@ func getTemplateConfigErrorMsg(err error) string {
        return fmt.Sprintf("An unexpected error occurred when configuring the template resolver: %v", err)
 }
 
-// handleObjectTemplates iterates through all policy templates in a given policy and processes them
-func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.ConfigurationPolicy) {
-       log := log.WithValues("policy", plc.GetName())
-       log.V(1).Info("Processing object templates")
-
-       // error if no remediationAction is specified
-       if plc.Spec.RemediationAction == "" {
-               message := "Policy does not have a RemediationAction specified"
-               log.Info(message)
-               statusChanged := addConditionToStatus(&plc, 0, false, "No RemediationAction", message)
-
-               if statusChanged {
-                       r.Recorder.Event(&plc, eventWarning,
-                               fmt.Sprintf(plcFmtStr, plc.GetName()), convertPolicyStatusToString(&plc))
-               }
-
-               r.addForUpdate(&plc, statusChanged)
-
-               return
-       }
-
-       // initialize the RelatedObjects for this Configuration Policy
-       oldRelated := append([]policyv1.RelatedObject{}, plc.Status.RelatedObjects...)
-       relatedObjects := []policyv1.RelatedObject{}
-       parentStatusUpdateNeeded := false
+type objectTemplateDetails struct {
+       name         string
+       namespace    string
+       isNamespaced bool
+}
 
+func (r *ConfigurationPolicyReconciler) getObjectTemplateDetails(plc policyv1.ConfigurationPolicy) ([]objectTemplateDetails, []string, bool, error) {
        // Collect template information for processing:
        // - Object name, if specified
        // - Object namespace, if specified
@@ -368,11 +349,7 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
        //
        // Also determine whether a namespace query is needed for this policy:
        //   If a namespaceSelector is present AND objects are namespaced without a namespace specified
-       templateObjs := make([]struct {
-               name         string
-               namespace    string
-               isNamespaced bool
-       }, len(plc.Spec.ObjectTemplates))
+       templateObjs := make([]objectTemplateDetails, len(plc.Spec.ObjectTemplates))
        selectedNamespaces := []string{}
        queryNamespaces := false
 
@@ -381,7 +358,7 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
                if jsonErr := json.Unmarshal(objectT.ObjectDefinition.Raw, &blob); jsonErr != nil {
                        log.Error(jsonErr, "Could not unmarshal data from JSON")
 
-                       return
+                       return templateObjs, selectedNamespaces, false, jsonErr
                }
 
                var unstruct unstructured.Unstructured
@@ -436,8 +413,6 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
                                        "Error filtering namespaces with provided NamespaceSelector: %s", err.Error())
                                statusChanged := addConditionToStatus(&plc, 0, false, reason, msg)
                                if statusChanged {
-                                       parentStatusUpdateNeeded = true
-
                                        r.Recorder.Event(
                                                &plc,
                                                eventWarning,
@@ -448,7 +423,7 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 
                                r.addForUpdate(&plc, statusChanged)
 
-                               return
+                               return templateObjs, selectedNamespaces, true, err
                        }
 
                        if len(selectedNamespaces) == 0 {
@@ -458,6 +433,44 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
                }
        }
 
+       return templateObjs, selectedNamespaces, false, nil
+}
+
+// handleObjectTemplates iterates through all policy templates in a given policy and processes them
+func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.ConfigurationPolicy) {
+       log := log.WithValues("policy", plc.GetName())
+       log.V(1).Info("Processing object templates")
+
+       // error if no remediationAction is specified
+       if plc.Spec.RemediationAction == "" {
+               message := "Policy does not have a RemediationAction specified"
+               log.Info(message)
+               statusChanged := addConditionToStatus(&plc, 0, false, "No RemediationAction", message)
+
+               if statusChanged {
+                       r.Recorder.Event(&plc, eventWarning,
+                               fmt.Sprintf(plcFmtStr, plc.GetName()), convertPolicyStatusToString(&plc))
+               }
+
+               r.addForUpdate(&plc, statusChanged)
+
+               return
+       }
+
+       // initialize the RelatedObjects for this Configuration Policy
+       oldRelated := append([]policyv1.RelatedObject{}, plc.Status.RelatedObjects...)
+       relatedObjects := []policyv1.RelatedObject{}
+       parentStatusUpdateNeeded := false
+
+       templateObjs, selectedNamespaces, parentStatusUpdateNeeded, err := r.getObjectTemplateDetails(plc)
+       if err != nil {
+               if parentStatusUpdateNeeded {
+                       r.checkRelatedAndUpdate(plc, relatedObjects, oldRelated, parentStatusUpdateNeeded)
+               }
+
+               return
+       }
+
        addTemplateErrorViolation := func(reason, msg string) {
                log.Info("Setting the policy to noncompliant due to a templating error", "error", msg)
```

controllers/configurationpolicy_controller.go
```diff
@@ -364,6 +361,103 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
+						&plc,
+						eventWarning,
+						fmt.Sprintf(plcFmtStr, plc.GetName()),
+						convertPolicyStatusToString(&plc),
+					)
+				}
+
+				r.addForUpdate(&plc, statusChanged)
+
+				return
```
mprahl, line 451: You'll need to also call `r.checkRelatedAndUpdate` before returning for the status update to take effect.

controllers/configurationpolicy_controller.go
dhaiducek: Yeah, yesterday I was trying to get it the heck out of the giant file, but this at least does make it more readable. I'll update it--thanks!

controllers/configurationpolicy_controller.go
mprahl: It's probably okay actually to just call `addForUpdate`.

controllers/configurationpolicy_controller.go
mprahl: @dhaiducek you could move it to another file in the same package if you want though.

controllers/configurationpolicy_controller.go
dhaiducek: I think you're right, though--it's probably clearer to wipe out the related resources if the policy/namespaceSelector is invalidated. I've moved the logic up so that it applies to `remediationAction` also.

controllers/configurationpolicy_controller.go
```diff
@@ -335,15 +335,120 @@ func getTemplateConfigErrorMsg(err error) string {
...
+			var err error
+			selectedNamespaces, err = common.GetSelectedNamespaces(selector)
+			if err != nil {
+				log.Error(
+					err, "Error filtering namespaces with provided NamespaceSelector",
+					"namespaceSelector", fmt.Sprintf("%+v", selector))
+
+				reason := "NamespaceSelector error"
+				msg := fmt.Sprintf(
+					"Error filtering namespaces with provided NamespaceSelector: %s", err.Error())
```
mprahl, line 417: This is user-facing so the YAML casing would be better.
```suggestion
					"Error filtering namespaces with provided namespaceSelector: %s", err.Error())
```

controllers/configurationpolicy_controller.go
```diff
@@ -860,29 +936,34 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 	objNames := []string{}
 	remediation := policy.Spec.RemediationAction
 
-	name, kind, metaNamespace := getDetails(unstruct)
-	if metaNamespace != "" {
-		namespace = metaNamespace
-		log = log.WithValues("namespace", metaNamespace)
+	// Since the object namespace is retrieved in the calling function, this condition should not be
+	// met and is an error
+	if objDetails.namespace != "" && objDetails.namespace != namespace {
```
mprahl, line 941: If this is true, then this is a programming error. Perhaps you could just `panic` in this case since it's impossible to reach unless we introduce a bug.

deploy/crds/policy.open-cluster-management.io_configurationpolicies.yaml
```diff
@@ -54,23 +54,67 @@ spec:
...
                 type: object
-              labelSelector:
-                additionalProperties:
-                  type: string
-                type: object
               namespaceSelector:
-                description: Target defines the list of namespaces to include/exclude
+                description: namespaceSelector defines the list of namespaces to include/exclude
+                  for objects defined in spec.objectTemplates. All selector rules
+                  are ANDed. If Include is not provided but MatchLabels and/or MatchExpressions
```
mprahl, line 60: Could you use the `lower-case` value of `Include` in the comment in the `NamespaceSelector` struct so that the casing matches what the user sees in the CRD?

pkg/common/namespace_selection.go
```diff
@@ -1,48 +1,99 @@
-// Copyright (c) 2020 Red Hat, Inc.
+// Licensed Materials - Property of IBM
```
mprahl, line 1: Yikes!

dhaiducek: Haha, oops. I'll need to make sure it gets taken out of `cert-policy-controller`. 🙂

mprahl (commented):
A few nitpicks, but it looks good to me!

dhaiducek:
Removing `getDetails()` turned out to be problematic because the templates aren't resolved when the object details are fetched initially. I added it back and leveraged it inside of the new `getObjectTemplateDetails()`.

controllers/configurationpolicy_controller.go
```diff
@@ -335,15 +335,117 @@ func getTemplateConfigErrorMsg(err error) string {
...
 }
 
+type objectTemplateDetails struct {
+	kind         string
+	name         string
+	namespace    string
+	isNamespaced bool
+}
+
+func getDetails(templateObj *objectTemplateDetails, unstruct unstructured.Unstructured) {
```
mprahl, line 345: `getDetails` implies that the values are turned.
```suggestion
func addDetails(templateObj *objectTemplateDetails, unstruct unstructured.Unstructured) {
```

controllers/configurationpolicy_controller.go
```diff
@@ -860,29 +933,40 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 	objNames := []string{}
 	remediation := policy.Spec.RemediationAction
 
-	name, kind, metaNamespace := getDetails(unstruct)
-	if metaNamespace != "" {
-		namespace = metaNamespace
-		log = log.WithValues("namespace", metaNamespace)
+	// Re-fetch details in case they were templated
+	getDetails(&objDetails, unstruct)
```
mprahl, line 937: Can you call this in `handleObjectTemplates` right after template processing instead?

Calling it here causes the implementation details of `handleObjectTemplates` to be leaked into `handleObjects`. It also doesn't need to be called for *every* namespace, just every object template.

controllers/configurationpolicy_controller_suite_test.go
```diff
@@ -43,7 +43,7 @@ func TestMain(m *testing.M) {
 		CRDDirectoryPaths: []string{filepath.Join("..", "..", "..", "deploy", "crds")},
 	}
 
-	err := policiesv1alpha1.AddToScheme(scheme.Scheme)
```
mprahl, line 46: These changes should have been part of a different commit but I'll let it slide this time. :slightly_smiling_face:

pkg/common/namespace_selection.go
```diff
@@ -1,48 +1,95 @@
...
+	log.V(2).Info("Returning final filtered namespace list", "namespaces", finalList)
+
+	return finalList, err
 }
 
-// GetAllNamespaces gets the list of all namespaces from k8s
-func GetAllNamespaces() (list []string, err error) {
-	namespaces := KubeClient.CoreV1().Namespaces()
-	namespaceList, err := namespaces.List(context.TODO(), metav1.ListOptions{})
+// GetAllNamespaces gets the list of all namespaces from k8s.
```
mprahl, line 68:
```suggestion
// GetAllNamespaces gets the list of all namespaces from k8s that matches the input label selector.
```

pkg/common/namespace_selection_test.go
```diff
@@ -1,114 +1,358 @@
-// Copyright (c) 2020 Red Hat, Inc.
+// Copyright 2019 The Kubernetes Authors.
```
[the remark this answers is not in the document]
dhaiducek: Oops. Another copy/paste from cert. I'll fix it.

pkg/common/namespace_selection.go
```diff
@@ -1,48 +1,95 @@
-// Copyright (c) 2020 Red Hat, Inc.
+/// Copyright (c) 2020 Red Hat, Inc.
```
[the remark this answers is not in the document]
dhaiducek: I just really wanted to make sure it was a comment? Maybe I need another `/` or three...

controllers/configurationpolicy_controller_suite_test.go
dhaiducek: I was being lazy last night. I'm going to spend a bit more time figuring out commits this morning.

controllers/configurationpolicy_controller.go
dhaiducek: Good point. Yes, I'll move it.

controllers/configurationpolicy_controller_suite_test.go
mprahl: Don't spend too much time on this. Just try to separate your commits in future PRs.

controllers/configurationpolicy_utils.go
```diff
@@ -55,6 +56,23 @@ func addRelatedObjects(
...
+
+		return unstruct, jsonErr
+	}
+
+	unstruct.Object = make(map[string]interface{})
+	unstruct.Object = blob.(map[string]interface{})
+
+	return unstruct, nil
+}
+
```
dhaiducek, line 75: I created this new function since was using this flow a couple times and it seemed better to move it to the `utils` file.

pkg/common/namespace_selection.go
```diff
@@ -5,44 +5,100 @@ package common
...
+		matchExpressions = *selector.MatchExpressions
+	}
+
+	labelSelector = metav1.LabelSelector{
+		MatchLabels:      matchLabels,
+		MatchExpressions: matchExpressions,
+	}
+
+	return labelSelector
+}
```
dhaiducek, line 43: `parseToLabelSelector()` is also new to make the code reusable if `policyv1.Target` is used in the future.

pkg/common/namespace_selection.go
```diff
@@ -5,44 +5,100 @@ package common
...
+		return nil, fmt.Errorf("error parsing namespace LabelSelector: %w", err)
+	}
+
+	listOpt := metav1.ListOptions{
+		LabelSelector: parsedSelector.String(),
+	}
+
+	log.V(2).Info("Retrieving namespaces with LabelSelector", "LabelSelector", parsedSelector.String())
+
+	nsList, err := (KubeClient).CoreV1().Namespaces().List(context.TODO(), listOpt)
```
dhaiducek, line 90: As I was refactoring to abstract `policyv1.Target`, it started to bug me that these functions were namespace-specific, but that can be addressed if `policyv1.Target` does indeed get used for a different API endpoint since I don't think refactoring it at this point is necessary.

dhaiducek (commented):
Re-squashed, refactored, and reviewable! (I hope...)

controllers/configurationpolicy_controller.go
```diff
@@ -335,15 +335,111 @@ func getTemplateConfigErrorMsg(err error) string {
...
+		// override the namespace if specified in objectTemplates
+		if objectns, ok := metadata["namespace"]; ok {
+			templateObj.namespace = strings.TrimSpace(objectns.(string))
+		}
+	}
+
+	if objKind, ok := unstruct.Object["kind"]; ok {
+		templateObj.kind = strings.TrimSpace(objKind.(string))
+	}
+}
```
JustinKuli, line 361: That namespaced bit is annoying - I guess that's the reason this mutates an input, as opposed to constructing the `objectTemplateDetails` based on the `unstruct`?

Is it necessary to be so careful with the accessors and TrimSpace? Does `unstruct.GetName()` not quite work for this?

dhaiducek: The issue is that the values could be templated, so `addDetails` is used to figure out if anything is in those fields to determine whether namespaces need to be fetched. And, later on after templates have been resolved, it's called again to update the values with the resolved values. Does that answer the first question?

For the second question, this was a copy/paste of existing code. I agree--`TrimSpace` probably isn't necessary, and I'll remove it. It'd be much cleaner to use `GetName()`/`GetNamespace()`. There might be a very minor performance improvement by not using `GetName()` since we're fetching two things from `metadata`, but it'd be much cleaner using the built-in functions.

dhaiducek: Are you thinking this could look something like this?

```golang
	templateObj.kind = unstruct.GetKind()
	templateObj.name = unstruct.GetName()
	templateObj.namespace = unstruct.GetNamespace()
```

It is much cleaner. And the fact that we're not just passing around the `unstruct` instead is probably a future refactor that could be done, also? (And get rid of `addDetails()` entirely...)

dhaiducek: Actually, that is also more maintainable too (though not that Kind/Name/Namespace would change much). I'll go ahead and update it.

mprahl (commented):
I'll finish my review this afternoon.

controllers/configurationpolicy_controller.go
dhaiducek: Latest changes have removed the `addDetails()` function in favor of calling `unstruct` functions directly.

mprahl (dismissed):
Nice job!

mprahl:
@dhaiducek could you please fix the CI error?

controllers/configurationpolicy_controller.go
```diff
@@ -335,15 +335,94 @@ func getTemplateConfigErrorMsg(err error) string {
...
+			// If an error occurred in the NamespaceSelector, update the policy status and abort
+			var err error
+			selectedNamespaces, err = common.GetSelectedNamespaces(selector)
+			if err != nil {
+				errMsg := "Error filtering namespaces with provided namespaceSelector"
+				log.Error(
+					err, errMsg,
+					"namespaceSelector", fmt.Sprintf("%+v", selector))
+
+				reason := "NamespaceSelector error"
```
mprahl, line 389: Optional: This would match the case of the YAML for a user-facing message.
```suggestion
				reason := "namespaceSelector error"
```

controllers/configurationpolicy_controller.go
dhaiducek: Another followup to using `unstruct` functions directly. Removing the `strings.TrimSpace()` caused a failure because one of the templates was written with a multi-line value and the `\n` was included in the string, so it turns out we do need to have it.

dhaiducek:
I also got doc feedback, so in addition to code changes, this also updates the namespace message to:
```
namespaced object has no namespace specified from the policy namespaceSelector nor the object metadata
```

mprahl (approved):

PR #42, opened 2022-06-28
Add a metric to record the duration of evaluating all config policies

mprahl:
Relates:
https://github.com/stolostron/backlog/issues/23547

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	gocmp "github.com/google/go-cmp/cmp"
+	"github.com/prometheus/client_golang/prometheus"
 	templates "github.com/stolostron/go-template-utils/v2/pkg/templates"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -33,6 +34,7 @@ import (
 	extpoliciesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	policyv1 "open-cluster-management.io/config-policy-controller/api/v1"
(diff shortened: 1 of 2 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -68,6 +70,19 @@ var config *rest.Config
 // NamespaceWatched defines which namespace we can watch for the GRC policies and ignore others
 var NamespaceWatched string
 
+var evalLoopHistogram = prometheus.NewHistogram(
```
JustinKuli, line 73: I've never made an alert off of a histogram - what will that look like? I suspect this will be better than a gauge, but I'm interested to see how it will work (and if there are any non-obvious drawbacks).

JustinKuli (changes requested):

controllers/configurationpolicy_controller.go
mprahl: Note: These are untested queries.

I think a simple alert might be the following. This just takes the average time it took to evaluate all config policies over the last 5 minutes and if it's over 60 seconds, fire an alert.
```promql
(rate(config_policies_evaluation_duration_seconds_sum[5m]) / rate(config_policies_evaluation_duration_seconds_count[5m])) > 60
```

Having it as a histogram rather than a counter allows for checking the percentiles, so something like the following. This checks that at least 99% of the time, it took 30 seconds or less to process all config policies in the past 60 minutes.

```promql
histogram_quantile(.99, sum(rate(config_policies_evaluation_duration_seconds_bucket[60m]))) > 30
```

controllers/configurationpolicy_controller.go
```diff
@@ -68,6 +70,19 @@ var config *rest.Config
...
+	prometheus.HistogramOpts{
+		Name:    "config_policies_evaluation_duration_seconds",
+		Help:    "The seconds that it takes to evaluate all configuration policies on the cluster",
+		Buckets: []float64{1, 3, 9, 10.5, 15, 30, 60, 90, 120, 180, 300, 450, 600},
```
JustinKuli, line 77: I think these buckets are a good starting point. I think we need to use the metric in some tests and see which ones are useful, or where we might want more data. And I think I'd want us to re-visit this before a release, since changing it will cause incompatible data between versions.

Eg, we might find that the buckets past 60 aren't worth the additional time-series, and we might want more below 10 in order to more accurately estimate how many more policies we could add before hitting saturation (at 10s).

JustinKuli (approved):

berenss:
Excellent news hearing about this metric && the related alert. 
please add a DOC item to expose this information to the customer in our public portal topic: https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.5/html/observability/observing-environments-intro#metric-types

We will also want to make use of this metric/alert so that we can report on the health of the ACM hub itself.

gparvin:
The metric is not pushed to Observability by default.  This was intentionally delayed for some further discussions: see https://github.com/stolostron/backlog/issues/24521
Since it's not in Observability by default, I expect we don't want to add the doc item yet.

berenss:
It can be queried locally - meaning, directly in the prometheus running on the hub. And I see the other 2 GRC metrics in the page, 
`policy_governance_info` and `policyreport_info`
probably fall into that same boat (metric is not pushed to Observability by default).

I think the intent of the table in the doc is to be transparent about the metrics we enable/support so that customers/users can begin to make use of them, even if perhaps the overall integration user story is not yet complete.

I could be wrong about that, perhaps Joydeep would have an opinion to share (perhaps he already did).

PR #44, opened 2022-07-07
add fields for obj deletion to crd

willkutler:

closes https://github.com/stolostron/backlog/issues/24025

### DIFF
--- a/api/v1/configurationpolicy_types.go
+++ b/api/v1/configurationpolicy_types.go
@@ -27,6 +27,10 @@ type RemediationAction string
 // +kubebuilder:validation:Enum=low;Low;medium;Medium;high;High;critical;Critical
 type Severity string
 
+// PruneObjectBehavior : DeleteAll, DeleteIfCreated, or None
+// +kubebuilder:validation:Enum=DeleteAll;DeleteIfCreated;None;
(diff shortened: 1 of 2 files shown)


JustinKuli (approved):

api/v1/configurationpolicy_types.go
```diff
@@ -27,6 +27,10 @@ type RemediationAction string
 // +kubebuilder:validation:Enum=low;Low;medium;Medium;high;High;critical;Critical
 type Severity string
 
+// PruneObjectBehavior : DeleteAll, DeleteIfCreated, or None
+// +kubebuilder:validation:Enum=DeleteAll;DeleteIfCreated;None;
+type PruneObjectBehavior string
```
mprahl, line 32: Optional: What do you think about adding a default value of `None` explicitly with a kubebuilder marker?
https://book.kubebuilder.io/reference/markers/crd-validation.html#crd-validation

I think this does two things:
* It makes it clear what the pruneObjectBehavior is regardless of if it was explicitly set.
* It makes the new field discoverable when existing policies are queried from the API.

willkutler: sure, I can add this too

PR #46, opened 2022-07-11
set related object properties on creation

willkutler:
https://github.com/stolostron/backlog/issues/24026

### DIFF
--- a/api/v1/configurationpolicy_types.go
+++ b/api/v1/configurationpolicy_types.go
@@ -260,8 +260,8 @@ type RelatedObject struct {
 	//
 	Compliant string `json:"compliant,omitempty"`
 	//
-	Reason     string           `json:"reason,omitempty"`
-	Properties ObjectProperties `json:"properties,omitempty"`
+	Reason     string            `json:"reason,omitempty"`
+	Properties *ObjectProperties `json:"properties,omitempty"`
 }
 
 // ObjectResource is an object identified by the policy as a resource that needs to be validated.
@@ -287,7 +287,7 @@ type ObjectMetadata struct {
 
 type ObjectProperties struct {
 	// Whether the object was created by the parent policy
-	CreatedByPolicy bool `json:"createdByPolicy,omitempty"`
+	CreatedByPolicy *bool `json:"createdByPolicy,omitempty"`
 	// Store object UID to help track object ownership for deletion
 	UID string `json:"uid,omitempty"`
 }
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -747,6 +747,18 @@ func sortRelatedObjectsAndUpdate(
 
 	update := false
 
+	// don't set creation info if it has already been set
+	for i, newEntry := range related {
+		for _, oldEntry := range oldRelated {
+			if oldEntry.Object.Kind == newEntry.Object.Kind &&
+				oldEntry.Object.Metadata.Name == newEntry.Object.Metadata.Name &&
+				oldEntry.Object.Metadata.Namespace == newEntry.Object.Metadata.Namespace &&
+				oldEntry.Properties != nil {
+				related[i].Properties = oldEntry.Properties
+			}
+		}
+	}
+
 	if len(oldRelated) == len(related) {
 		for i, entry := range oldRelated {
 			if !gocmp.Equal(entry, related[i]) {
@@ -983,13 +995,23 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 
 		log.V(2).Info("Handling a single object template")
 
-		objNames, compliant, rsrcKind, statusUpdateNeeded = r.handleSingleObj(
+		var creationInfo *policyv1.ObjectProperties
+
+		objNames, compliant, rsrcKind, statusUpdateNeeded, creationInfo = r.handleSingleObj(
 			singObj, remediation, exists, dclient, objectT,
 		)
 		// The message string for single objects is different than for multiple
 		reason = generateSingleObjReason(objShouldExist, compliant, exists)
 		// Enforce could clear the objNames array so use name instead
-		relatedObjects = addRelatedObjects(compliant, rsrc, namespace, objDetails.isNamespaced, []string{name}, reason)
+		relatedObjects = addRelatedObjects(
+			compliant,
+			rsrc,
+			namespace,
+			objDetails.isNamespaced,
+			[]string{name},
+			reason,
+			creationInfo,
+		)
 	} else { // This case only occurs when the desired object is not named
 		if objShouldExist {
 			if exists {
@@ -1009,7 +1031,7 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 			}
 		}
 
-		relatedObjects = addRelatedObjects(compliant, rsrc, namespace, objDetails.isNamespaced, objNames, reason)
+		relatedObjects = addRelatedObjects(compliant, rsrc, namespace, objDetails.isNamespaced, objNames, reason, nil)
 
 		if !statusUpdateNeeded {
 			log.V(2).Info("The status did not change for this object template")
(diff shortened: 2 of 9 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -983,13 +995,21 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 
 		log.V(2).Info("Handling a single object template")
 
-		objNames, compliant, rsrcKind, statusUpdateNeeded = r.handleSingleObj(
+		_, compliant, _, _, creationInfo := r.handleSingleObj(
```
JustinKuli, line 998: I'm not following why some of these are now `_`... previously some of these values were being updated by what was coming from `handleSingleObj`. If they no longer need to be updated, I think this is the only place that function is called, so we could just remove them from the function signature.

Or maybe this is appearing because of variable shadowing? Were there unused variable errors before changing them to `_`? In which case, I think that `creationInfo` should be initialized separately, then this can be remain a `=` statement, instead of `:=`.

controllers/configurationpolicy_controller.go
```diff
@@ -1058,12 +1078,22 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 	exists bool,
 	dclient dynamic.Interface,
 	objectT *policyv1.ObjectTemplate,
-) (objNameList []string, compliance bool, rsrcKind string, statusUpdateNeeded bool) {
+) (
+	objNameList []string,
+	compliance bool,
+	rsrcKind string,
+	statusUpdateNeeded bool,
+	creationInfo map[string]interface{},
```
JustinKuli, line 1086: Optional:
```suggestion
	creationInfo policyv1.ObjectProperties,
```
Using the type might be a bit more performant, and would save some conversion later.

JustinKuli (changes requested):
I think it's on the right track, there could be reasons my suggestions wouldn't work, so just let me know :)

One other comment: I think the only test changes were for the updated function signatures. How much work will it be to implement tests for this functionality? Making sure that it gets populated when an object is created, not populated when it's just updated, etc.

controllers/configurationpolicy_controller.go
willkutler: I thought these variables weren't actually used in that block but I might have been wrong. I will look into it

api/v1/configurationpolicy_types.go
```diff
@@ -287,7 +287,7 @@ type ObjectMetadata struct {
 
 type ObjectProperties struct {
 	// Whether the object was created by the parent policy
-	CreatedByPolicy bool `json:"createdByPolicy,omitempty"`
+	CreatedByPolicy *bool `json:"createdByPolicy,omitempty"`
```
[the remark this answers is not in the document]
willkutler: if createdByPolicy isn't a pointer then `false` won't show up in the yaml - omitempty will leave it out

test/resources/case20_delete_objects/case20_edit_pod.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: policy.open-cluster-management.io/v1
+kind: ConfigurationPolicy
+metadata:
+  name: policy-pod-edit
```
JustinKuli, line 4: Optional nit: I wonder if a more accurate name would be `policy-pod-already-created`.

test/e2e/case20_delete_objects_test.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+const (
+	case20ConfigPolicyNameCreate string = "policy-pod-create"
+	case20ConfigPolicyNameEdit   string = "policy-pod-edit"
+	case20PolicyYamlCreate       string = "../resources/case20_delete_objects/case20_create_pod.yaml"
+	case20PolicyYamlEdit         string = "../resources/case20_delete_objects/case20_edit_pod.yaml"
+)
+
+var _ = Describe("Test status fields being set for object deletion", func() {
```
JustinKuli, line 20: Good start to tests here - I think I will want another test to make sure that editing an object will not populate `createdByPolicy` (the current policy does not make any changes to the object). Based on the code, I believe this will pass, it's just a matter of completeness here.

This test can be added in another PR.

controllers/configurationpolicy_controller.go
```diff
@@ -1058,12 +1080,23 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 	exists bool,
 	dclient dynamic.Interface,
 	objectT *policyv1.ObjectTemplate,
-) (objNameList []string, compliance bool, rsrcKind string, statusUpdateNeeded bool) {
+) (
+	objNameList []string,
+	compliance bool,
+	rsrcKind string,
+	statusUpdateNeeded bool,
+	creationInfo policyv1.ObjectProperties,
```
JustinKuli, line 1088: I'm surprised this doesn't need to return a pointer to identify the difference between nil and the empty values... but I don't think it's worth worrying over it while we're still working on the initial implementation. If it needs changing, I think it will pop up at some point during tests.

test/resources/case20_delete_objects/case20_edit_pod.yaml
```diff
@@ -0,0 +1,23 @@
...
+        kind: Pod
+        metadata:
+          name: nginx-pod-e2e20
+        spec:
+          containers:
+            - image: nginx:1.7.9
+              name: nginx
+              ports:
+                - containerPort: 80
+
```
JustinKuli, line 23: Optional nit: github is complaining about an end of file thing here (and the other yaml)... if you make any other updates to this PR, see if that can be cleaned up.

JustinKuli (dismissed):
/hold
LGTM! A couple optional nits. Unhold if you don't want to address them

controllers/configurationpolicy_controller.go
willkutler: you're right, this did need to be fixed - it was causing inform policies to have the same properties as enforce policies that did not create the object

controllers/configurationpolicy_controller.go
```diff
@@ -1428,6 +1469,13 @@ func enforceByCreatingOrDeleting(obj singleObject, dclient dynamic.Interface) (r
 			log.V(2).Info("Created missing must have object", "resource", obj.gvr.Resource, "name", obj.name)
 			reason = "K8s creation success"
 			msg = fmt.Sprintf("%v %v was missing, and was created successfully", obj.gvr.Resource, idStr)
+
+			var uidIsString bool
+			uid, uidIsString, err = unstructured.NestedString(obj.object.Object, "metadata", "uid")
+
+			if !uidIsString || err != nil {
+				log.Error(err, "Tried to set UID in status but the field is not a string")
```
JustinKuli, line 1477: More accurately, it's either not a string, or some other error happened... You might consider splitting them into separate cases? But I'm pretty sure k8s will guarantee it's a string, so I don't see this really being an issue.

JustinKuli (dismissed):
/hold

Thanks for adding more tests! Looking good, assuming everything passes. And I think you wanted to rebase/squash the commits, so I'm adding the hold.

controllers/configurationpolicy_controller.go
```diff
@@ -983,13 +995,23 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 
 		log.V(2).Info("Handling a single object template")
 
-		objNames, compliant, rsrcKind, statusUpdateNeeded = r.handleSingleObj(
+		var creationInfo *policyv1.ObjectProperties
+
+		objNames, compliant, rsrcKind, statusUpdateNeeded, creationInfo = r.handleSingleObj(
```
JustinKuli, line 1000: Please rebase on top of https://github.com/open-cluster-management-io/config-policy-controller/pull/47, I think there will be a conflict here.

JustinKuli (changes requested):

JustinKuli (approved):
Looks good, sorry for the false alarm about the conflict

PR #47, opened 2022-07-13
Fix bug where parent policy event was not sent

JustinKuli:
The `parentStatusUpdateNeeded` variable is (usually) carefully guarded
to ensure it only ever moves from false->true, and never back to false.
The moved section of code was not obeying that, and it was being run for
every object template. The result was that one object template could
mark the status as needing an update, but then be overriden by a later
run through the loop.

This section can just be outside of the loop, improving performance and
removing the bug. The additional variable to guard the update to
`parentSatusUpdateNeeded` might not be strictly necessary, but this bug
was too annoying to find to not make every effort to prevent it in the
future.

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -511,22 +511,8 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 
 	log.V(2).Info("Processing the object templates", "count", len(plc.Spec.ObjectTemplates))
 
-	for indx, objectT := range plc.Spec.ObjectTemplates {
-		nonCompliantObjects := map[string]map[string]interface{}{}
-		compliantObjects := map[string]map[string]interface{}{}
-		enforce := strings.EqualFold(string(plc.Spec.RemediationAction), string(policyv1.Enforce))
-		kind := ""
-		objShouldExist := !strings.EqualFold(string(objectT.ComplianceType), string(policyv1.MustNotHave))
-
-		// Here appears to be a  good place to hook in template processing
-		// This is at the head of objectemplate processing
-		// ( just before the perNamespace handling of objectDefinitions)
-
-		// check here to determine if the object definition has a template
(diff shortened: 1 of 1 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -511,6 +511,25 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 
 	log.V(2).Info("Processing the object templates", "count", len(plc.Spec.ObjectTemplates))
 
+	// Parse and fetch details from each object in each objectTemplate, and gather namespaces if required
+	var templateObjs []objectTemplateDetails
+	var selectedNamespaces []string
+	var objTmplStatusChangeNeeded bool
+
+	templateObjs, selectedNamespaces, objTmplStatusChangeNeeded, err = r.getObjectTemplateDetails(plc)
```
dhaiducek, line 519: I was starting to wonder if I introduced a bug somewhere... 😞 

So the issue with moving it here is that if one of the object details fetched is templated, it won't be resolved yet when this is called here.

dhaiducek: Ah nevermind, sorry--I found the loop you're referring to. But my template comment about moving this code back to its original location I think still applies.

mprahl: I think the best approach is loop over the object templates once for policy template processing. Then call `getObjectTemplateDetails` and then loop over the object templates again now that the policy templates are processed.

JustinKuli: With the new update it will have a loop for resolving templates, then do the `getObjectTemplateDetails`, then have another loop for the other stuff.

dhaiducek (approved):
LGTM! Thanks for fixing my "feature"! 😄

mprahl (approved):

JustinKuli:
Tide seems stuck... I'm going to try and poke it

/override KinD tests (1.18, latest)

JustinKuli:
Now I might've broken it

/override KinD tests (1.18, latest)

PR #48, opened 2022-07-18
Allow an alternative kubeconfig for evaluating policies

mprahl:
This allows the controller to run on a separate cluster than the one it
is managing.

As part of this, global variables were removed to make it simpler to
determine where the Kubernetes client and configuration is coming from.

Related:
https://github.com/stolostron/backlog/issues/24361


mprahl:
@dhaiducek let me know if you have a cleaner idea for the Makefile changes.

controllers/configurationpolicy_controller.go
```diff
@@ -2123,10 +2114,10 @@ func (r *ConfigurationPolicyReconciler) createParentPolicyEvent(instance *policy
 		eventMsg)
 }
 
-func createParentPolicy(instance *policyv1.ConfigurationPolicy) extpoliciesv1.Policy {
+func createParentPolicy(instance *policyv1.ConfigurationPolicy, defaultNamespace string) extpoliciesv1.Policy {
```
JustinKuli, line 2117: Nit: can we add doc for when/why the defaultNamespace is used?

controllers/configurationpolicy_controller.go
```diff
@@ -122,8 +108,13 @@ type ConfigurationPolicyReconciler struct {
 	DecryptionConcurrency uint8
 	// Determines the number of Go routines that can evaluate policies concurrently.
 	EvaluationConcurrency uint8
+	ParentPolicyNamespace string
```
JustinKuli, line 111: Can we add doc for what this is used for

main.go
```diff
@@ -256,7 +283,7 @@ func main() {
 			log.Info("Starting lease controller to report status")
 
 			leaseUpdater := lease.NewLeaseUpdater(
-				clientset,
+				kubernetes.NewForConfigOrDie(reconciler.TargetK8sConfig),
```
JustinKuli, line 286: Will this lease be on the managed cluster (like shown here), or will it be on the hosting cluster?

I think it would be hosting cluster, based on https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/63-hosted-addon#guidance-for-transforming-default-mode-addon-to-hosted-mode
> 2. Refactor the code of the addon agent to accept a flag like managed-kubeconfig, and use the kubeconfig received with the flag to build the managed cluster kube client, and use this client to process resources on the managed cluster. This kubeconfig should be different from the local kubeconfig or in-cluster kubeconfig which is mainly used for leader election and addon lease.

JustinKuli (changes requested):

controllers/configurationpolicy_controller.go
mprahl: I'm going to refactor this so that we just use the namespace from the `ConfigurationPolicy` object since that's always going to be the watch namespace anyways.

controllers/configurationpolicy_controller.go
mprahl: I'm going to remove this field.

controllers/configurationpolicy_controller.go
```diff
@@ -2124,15 +2114,11 @@ func (r *ConfigurationPolicyReconciler) createParentPolicyEvent(instance *policy
...
-	if ns == "" {
-		ns = NamespaceWatched
-	}
-
 	return extpoliciesv1.Policy{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      instance.OwnerReferences[0].Name,
-			Namespace: ns, // we assume that the parent policy is in the watched-namespace passed as flag
+			Name: instance.OwnerReferences[0].Name,
+			// It's assumed that the parent policy is in the same namespace as the configuration policy
```
JustinKuli, line 2120: BTW, this is a *very* safe assumption. From https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#owner-references-in-object-specifications:
> A valid owner reference consists of the object name and a UID within the same namespace as the dependent object.

JustinKuli (approved):
LGTM!

/hold for Dale (optional)

gparvin (dismissed):
Thanks for these changes Matt.  It looks good to me.

JustinKuli (approved):

PR #49, opened 2022-07-19
delete required related objects when necessary

willkutler:

https://github.com/stolostron/backlog/issues/24028
https://github.com/stolostron/backlog/issues/24027

Should be good to merge - I can add a few extra test cases via a later PR but would like to get this in earlier for QE

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -403,6 +403,88 @@ func (r *ConfigurationPolicyReconciler) getObjectTemplateDetails(
 	return templateObjs, selectedNamespaces, false, nil
 }
 
+func (r *ConfigurationPolicyReconciler) cleanUpChildObjects(plc policyv1.ConfigurationPolicy) []string {
+	deletionFailures := []string{}
+
+	for _, object := range plc.Status.RelatedObjects {
+		// set up client for object deletion
+		gvk := schema.FromAPIVersionAndKind(object.Object.APIVersion, object.Object.Kind)
+
+		log := log.WithValues("policy", plc.GetName(), "groupVersionKind", gvk.String())
+
+		mapper := restmapper.NewDiscoveryRESTMapper(r.apiGroups)
+
+		mapping, err := mapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+		if err != nil {
+			log.Error(err, "Could not get resource mapping for child object")
+
+			deletionFailures = append(deletionFailures, gvk.String()+fmt.Sprintf(` "%s" in namespace %s`,
+				object.Object.Metadata.Name, object.Object.Metadata.Namespace))
+
+			continue
+		}
+
+		dclient, gvr := r.getResourceAndDynamicClient(mapping)
+
+		namespaced := object.Object.Metadata.Namespace != ""
+
+		// determine whether object should be deleted
+		needsDelete := false
+
+		if strings.EqualFold(string(plc.Spec.RemediationAction), "enforce") {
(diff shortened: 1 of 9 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -403,6 +403,78 @@ func (r *ConfigurationPolicyReconciler) getObjectTemplateDetails(
...
+		if needsDelete {
+			var res dynamic.ResourceInterface
+			if namespaced {
+				res = dclient.Resource(gvr).Namespace(object.Object.Metadata.Namespace)
+			} else {
+				res = dclient.Resource(gvr)
+			}
+
+			if completed, err := deleteObject(res, object.Object.Metadata.Name, object.Object.Metadata.Namespace); !completed {
+				deletionFailures = append(deletionFailures, gvk.String()+fmt.Sprintf(` "%s"`, object.Object.Metadata.Name))
```
JustinKuli: Could this also include the namespace?

controllers/configurationpolicy_controller.go
```diff
@@ -429,6 +501,66 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
+		}
+	} else {
+		// if pruneObjectBehavior is none, no finalizer is needed
+		if len(plc.ObjectMeta.Finalizers) != 0 {
+			plc.SetFinalizers([]string{})
+			err := r.Update(context.TODO(), &plc)
+			if err != nil {
+				log.V(1).Error(err, "Error unsetting finalizer for configuration policy", plc)
+			}
+		}
```
JustinKuli: This will clobber any finalizers anything else might add to the ConfigurationPolicy, which could be bad.

controllers/configurationpolicy_utils.go
```diff
@@ -53,7 +54,7 @@ func addRelatedObjects(
 		}
 
 		relatedObject.Object.APIVersion = rsrc.GroupVersion().String()
-		relatedObject.Object.Kind = rsrc.Resource
+		relatedObject.Object.Kind = kind
```
JustinKuli, line 57: What's the purpose of this change? I feel like I'm missing something... was it recording incorrect information before?

JustinKuli (commented):
This is looking good, I think all the hard parts are already ironed out. Just some suggestions and one or two places it could cause problems

controllers/configurationpolicy_utils.go
willkutler: @JustinKuli Sort of - the `resource` and `kind` for an object are pretty similar, but I found that passing the resource in instead of the kind would cause the mapping function to be unable to query the object. For example, for pods the resource is `pods` and the kind is `Pod`. Using the kind makes the relatedObjects array line up better with the actual object (which displays 'kind') and makes the code for object deletion cleaner.

JustinKuli (approved):
LGTM. My understanding is that we will add and improve these tests, but they are a good starting point.

Two new thoughts I had that we should address / consider separately: 
1) since the cleanup is inside the `handleObjectTemplates`, if a policy has a slow evaluationInterval, it will be stuck with its finalizer for a long time. Ideally this will become event-driven at some point, but an intermediate step would just be to update the `shouldEvaluatePolicy` logic so that policies with non-nil deletion timestamps are always evaluated.
2) we'll want to test if the deletion can get stuck - if the config policy controller is trying to delete something with a finalizer that will not be removed (ie one we manually create) then does the evaluation loop hang? Or will the deletion request timeout at a reasonable time? Or does it just do background deletion, and we don't need to worry about it at all?

PR #50, opened 2022-07-21
evaluate policy upon deletion

willkutler:

addresses point 1 here: https://github.com/open-cluster-management-io/config-policy-controller/pull/49#pullrequestreview-1046616393

### DIFF
--- a/controllers/configurationpolicy_controller_test.go
+++ b/controllers/configurationpolicy_controller_test.go
@@ -523,6 +523,7 @@ func TestShouldEvaluatePolicy(t *testing.T) {
 		evaluationInterval      policyv1.EvaluationInterval
 		complianceState         policyv1.ComplianceState
 		expected                bool
+		deletionTimestamp       *metav1.Time
 	}{
 		{
 			"Just evaluated and the generation is unchanged",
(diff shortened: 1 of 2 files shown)


controllers/configurationpolicy_controller_test.go
```diff
@@ -627,6 +640,16 @@ func TestShouldEvaluatePolicy(t *testing.T) {
 			policyv1.EvaluationInterval{NonCompliant: "12h"},
 			policyv1.NonCompliant,
 			false,
+			nil,
+		},
+		{
+			"Deletion timestamp is non nil",
+			time.Now().UTC().Add(-12 * time.Second).Format(time.RFC3339),
```
mprahl, line 647: This will make it so that we know it's not based on the evaluation interval not being reached.
```suggestion
			time.Now().UTC().Add(-13 * time.Hour).Format(time.RFC3339),
```

controllers/configurationpolicy_controller_test.go
```diff
@@ -627,6 +640,16 @@ func TestShouldEvaluatePolicy(t *testing.T) {
...
+			2,
+			policyv1.EvaluationInterval{NonCompliant: "12h"},
+			policyv1.NonCompliant,
+			false,
```
mprahl, line 651: Don't we expect true here?

willkutler: oops! good catch

mprahl (approved):

PR #51, opened 2022-07-21
Add tests to check relatedObject updating

dhaiducek:
Tests for this issue that I thought was a bug but actually wasn't:
- https://github.com/stolostron/backlog/issues/24294

### DIFF
--- a/test/e2e/case14_selection_test.go
+++ b/test/e2e/case14_selection_test.go
@@ -4,130 +4,266 @@
 package e2e
 
 import (
+	"context"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


test/e2e/case14_selection_test.go
```diff
@@ -4,130 +4,266 @@
...
+			By("Deleting " + case14LimitRangeName + " on " + ns)
+			utils.Kubectl("delete", "-f", case14LimitRangeFile, "-n", ns)
+		}
+		for _, ns := range append(testNamespaces, newNs) {
+			By("Deleting namespace " + ns)
+			utils.Kubectl("delete", "namespace", ns, "--ignore-not-found")
+		}
+		for _, policy := range policyTests {
+			By("Deleting " + policy.name + " on managed")
+			utils.Kubectl("delete", "-f", policy.yamlFile, "-n", testNamespace)
```
gparvin, line 80: In some cases we would want to delete the policy first since it might recreate other resources that we want to delete.

gparvin (commented):
Double check the cleanup ordering.

gparvin (approved):
Looks good to me!

PR #53, opened 2022-07-26
Enable foreground deletion

willkutler:
causes config policy deletion to hang when child object deletion hangs, and adds "terminating" to the config policy status when deletion is in progress

### DIFF
--- a/api/v1/configurationpolicy_types.go
+++ b/api/v1/configurationpolicy_types.go
@@ -51,6 +51,9 @@ const (
 
 	// UnknownCompliancy is an ComplianceState
 	UnknownCompliancy ComplianceState = "UnknownCompliancy"
+
+	// Terminating is a ComplianceState
+	Terminating ComplianceState = "Terminating"
 )
 
 // Condition is the base struct for representing resource conditions
@@ -199,6 +202,7 @@ type ResourceState string
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="Compliance state",type="string",JSONPath=".status.compliant"
 
 // ConfigurationPolicy is the Schema for the configurationpolicies API
 type ConfigurationPolicy struct {
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -61,6 +61,7 @@ var (
 	reasonWantFoundDNE       = "Resource not found but should exist"
 	reasonWantNotFoundExists = "Resource found but should not exist"
 	reasonWantNotFoundDNE    = "Resource not found as expected"
+	reasonCleanupError       = "Error cleaning up child objects"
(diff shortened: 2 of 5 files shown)


api/v1/configurationpolicy_types.go
```diff
@@ -51,6 +51,9 @@ const (
 
 	// UnknownCompliancy is an ComplianceState
 	UnknownCompliancy ComplianceState = "UnknownCompliancy"
+
+	// Terminating is an ComplianceState
+	Terminating ComplianceState = "Terminating"
```
JustinKuli, line 56: Since we want this in the `kubectl` output, can we also add compliance state as a printercolumn? Like https://github.com/open-cluster-management-io/governance-policy-propagator/blob/main/api/v1/policy_types.go#L102, for reference

test/resources/case20_delete_objects/case20_createpod_finalizer.yaml
```diff
@@ -0,0 +1,25 @@
...
+  pruneObjectBehavior: DeleteIfCreated
+  object-templates:
+    - complianceType: musthave
+      objectDefinition:
+        apiVersion: v1
+        kind: Pod
+        metadata:
+          name: nginx-pod-cannot-delete
+          finalizers:
+          - policy.open-cluster-management.io/delete-object
```
JustinKuli, line 19: Nit: a name that makes it clear this is for testing could be better. Otherwise if I read this somewhere I might think it's some new undocumented feature thing.

controllers/configurationpolicy_controller.go
```diff
@@ -921,7 +950,10 @@ func addConditionToStatus(
 
 	var complianceState policyv1.ComplianceState
 
-	if compliant {
+	if reason == "Error cleaning up child objects" {
```
JustinKuli, line 953: Nit: I think this string could be better as a `const` somewhere, as opposed to a magic string.

JustinKuli (commented):
Looking good! Some small suggestions.

api/v1/configurationpolicy_types.go
```diff
@@ -51,6 +51,9 @@ const (
 
 	// UnknownCompliancy is an ComplianceState
 	UnknownCompliancy ComplianceState = "UnknownCompliancy"
+
+	// Terminating is an ComplianceState
```
mprahl, line 55:
```suggestion
	// Terminating is a ComplianceState
```

mprahl (dismissed):

JustinKuli (approved):

mprahl (approved):
In the future, please include the GitHub issue that this is resolving in the commit message.

dhaiducek:
@willkutler Please open a PR in the addon-controller to update the CRD there. It's a quick update--you can use the `./build/crd-sync.sh` script in the repo.

PR #57, opened 2022-07-29
Fix: update UID when recreating an object

JustinKuli:
When an object was re-created by the policy after being deleted, the
UID in the object's properties was not updated. Then, the logic for
DeleteIfCreated pruning would see the UID as stale and not remove the
object, even though it *was* created by the policy.

Refs:
 - https://github.com/stolostron/backlog/issues/24677

Co-authored-by: Will Kutler <wkutler@redhat.com>


willkutler (dismissed):

willkutler:
needs a format fix but looks good to me otherwise

PR #60, opened 2022-08-02
Send compliance events in a few more situations

JustinKuli:
Sending an event whenever the generation changes seems like a good idea
in general. Sending an event when the ComplianceState changes is
particularly useful because other components can use this as a
mechanism to ensure a new compliance event is generated.

Refs:
 - https://github.com/stolostron/backlog/issues/24750


JustinKuli:
/hold

I'll triple check this actually accomplishes what I want when I have all the patched images ready

mprahl (approved):

mprahl (approved):

PR #61, opened 2022-08-11
Update named objects in arrays instead of appending

mprahl:
In the case of a Deployment object's container array, if a policy
updates the image of a container, the container should be updated.

Before this commit, a new container with the same name would be
erroneously appended to the containers array and cause the Kubernetes
API update request to fail and make the policy non-compliant.

This assumption is safe due to the Kubernetes API conventions:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

The first commit in the PR is a micro performance improvement.

Relates:
https://github.com/stolostron/backlog/issues/25061
https://bugzilla.redhat.com/show_bug.cgi?id=2117728

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -1917,16 +1917,34 @@ func mergeArrays(newArr []interface{}, old []interface{}, ctype string) (result 
 			}
 
 			var mergedObj interface{}
+			// Stores if val1 and val2 are maps with the same "name" key value. In the case of the containers array
+			// in a Deployment object, the value should be merged and not appended if the name is the same in both.
+			var sameNamedObjects bool
 
 			switch val2 := val2.(type) {
 			case map[string]interface{}:
+				// If the policy value and the current value are different types, use the same logic
+				// as the default case.
+				val1, ok := val1.(map[string]interface{})
+				if !ok {
+					mergedObj = val1
+
+					break
+				}
+
+				if name2, ok := val2["name"].(string); ok && name2 != "" {
+					if name1, ok := val1["name"].(string); ok && name1 == name2 {
+						sameNamedObjects = true
+					}
+				}
+
 				// use map compare helper function to check equality on lists of maps
-				mergedObj, _ = compareSpecs(val1.(map[string]interface{}), val2, ctype)
+				mergedObj, _ = compareSpecs(val1, val2, ctype)
 			default:
 				mergedObj = val1
 			}
 			// if a match is found, this field is already in the template, so we can skip it in future checks
-			if equalObjWithSort(mergedObj, val2) {
+			if sameNamedObjects || equalObjWithSort(mergedObj, val2) {
 				count++
 
 				newArr[newArrIdx] = mergedObj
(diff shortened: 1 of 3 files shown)


test/e2e/case22_pod_update_image_test.go
```diff
@@ -0,0 +1,74 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"open-cluster-management.io/config-policy-controller/test/utils"
+)
+
+var _ = FDescribe("Test an alternative kubeconfig for policy evaluation", Ordered, func() {
```
dhaiducek, line 17: Looks like a leftover from testing?

```suggestion
var _ = Describe("Test an alternative kubeconfig for policy evaluation", Ordered, func() {
```

dhaiducek (changes requested):
Thanks! I think `name` is a great place to start for our array behaviors!

One minor change needed from testing.

test/e2e/case22_pod_update_image_test.go
mprahl: Thanks for catching that! That's what happens when you try to get a PR up right before the long weekend.  😆

test/e2e/case22_pod_update_image_test.go
```diff
@@ -0,0 +1,74 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"open-cluster-management.io/config-policy-controller/test/utils"
+)
+
+var _ = Describe("Test an alternative kubeconfig for policy evaluation", Ordered, func() {
```
JustinKuli, line 17: This is the same Describe text as case 21 - I'm guessing it was just copied over?

JustinKuli (dismissed):
Basically a nit, this looks good!

I must have been misremembering https://kubernetes.io/docs/concepts/workloads/pods/#pod-update-and-replacement, I thought changing the image in a running pod wouldn't work, but I was wrong.

test/e2e/case22_pod_update_image_test.go
mprahl: Thanks for catching that!

controllers/configurationpolicy_controller.go
```diff
@@ -1914,16 +1917,34 @@ func mergeArrays(newArr []interface{}, old []interface{}, ctype string) (result
...
+				val1, ok := val1.(map[string]interface{})
+				if !ok {
+					mergedObj = val1
+
+					break
+				}
+
+				if name2, ok := val2["name"].(string); ok && name2 != "" {
+					if name1, ok := val1["name"].(string); ok && name1 == name2 {
+						sameNamedObjects = true
```
dhaiducek, line 1937: I was thinking about this over the weekend. I'm wondering if we should go ahead and make this a loop over a package variable array of strings rather than a single hardcoded value. Then it'll be ready to go if/when we support customizing matching keys.

mprahl: It's an interesting idea, but I think that falls under the "premature optimization" category.  :smile:

dhaiducek (approved):

willkutler (approved):

PR #62, opened 2022-08-22
fix logic in sortRelatedObjectsAndUpdate

ChunxiAlexLuo:
Signed-off-by: Chunxi Luo chuluo@redhat.com

https://github.com/stolostron/backlog/issues/24572

So I found more issues than the original bug, it is not only related to namespace change, but also happens under the same namespace.  

For example, test case A):

1) Create an enforcement policy `create-pod` with enabling pruneObjectBehavior option to create a pod `policy-pod-1` in the specific namespace ( ie `auto-test-1`)

2) The pod `policy-pod-1` is created in the namespaces `auto-test-1`, check `create-pod` YAML and  find `policy-pod-1` under `relatedObjects`

3) edit the policy `create-pod` to change the pod name to `policy-pod-2` but still target the same namespace `auto-test-1`

4) The pod `policy-pod-2` is also created in the namespace `auto-test-1`, then check the pods in the namespaces `auto-test-1`, both pods `policy-pod-1` and `policy-pod-2` exist. 

5) check `create-pod` YAML and only find `policy-pod-2` under `relatedObjects`, `policy-pod-1` disappeared in `relatedObjects`.

6) Delete the policy `create-pod`, the second pod `policy-pod-2` is deleted, but the first pod `policy-pod-1` still be existed in the same namespace `auto-test-1`.

Test case B is described in the original issue:
1) Create an enforcement policy with enabling pruneObjectBehavior option to create a pod in the specific namespace ( ie auto-test-1, auto-test-2)

2) The pod is created in these two namespaces, then edit the policy to change the object namespace as the default namespace

3) The pod is created in the default namespace, then check the pod in the above two namespaces, this pod still be existed.

4) Delete the policy, this pod is deleted from the default namespace, but two previous pod still be existed in `auto-test-1`, `auto-test-2` namespace.

Both Test Cases A) and B) are due to the same reason that in `sortRelatedObjectsAndUpdate`, when the oldEntry is different and not in the new related, we don't append it to the new related.

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -932,6 +932,23 @@ func sortRelatedObjectsAndUpdate(
 		}
 	}
 
+	// add RelatedObject into newRelated which is in oldRelated but not in newRelated
+	for _, oldEntry := range oldRelated {
+		ifAppend := true
+
+		if oldEntry.Properties != nil && (*oldEntry.Properties.CreatedByPolicy) {
+			for _, newEntry := range related {
+				if gocmp.Equal(newEntry.Object, oldEntry.Object) {
+					ifAppend = false
+				}
+			}
+
+			if ifAppend {
+				related = append(related, oldEntry)
+			}
+		}
+	}
+
 	if len(oldRelated) == len(related) {
 		for i, entry := range oldRelated {
 			if !gocmp.Equal(entry, related[i]) {


ChunxiAlexLuo:
My original test only tested case b) so I'm adding case a).  Let me know if I missed any original design logic as well as a better way to refactor the whole sortRelatedObjectsAndUpdate function.

mprahl:
@willkutler and @JustinKuli, for case A outlined in the PR comment, should a policy that is modified to enforce a different object clean up the old object when `pruneObjectBehavior` is set? I'm personally on the fence about this because the behavior of `pruneObjectBehavior` is on policy deletion and not on policy modification.

I think ArgoCD would clean up the object in this case and it may be worthwhile making this consistent with ArgoCD since I imagine this is the behavior that users may expect:
https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#resources-prune-deletion-propagation-policy

controllers/configurationpolicy_controller.go
```diff
@@ -932,6 +932,23 @@ func sortRelatedObjectsAndUpdate(
 		}
 	}
 
+	// add RelatedObject into newRelated which is in oldRelated but not in newRelated
+	for _, oldEntry := range oldRelated {
+		ifAppend := true
```
JustinKuli, line 937: Nit: rename to `shouldAppend`. Or possibly reverse it and call it `found`, so that the end is "if not found, append".

controllers/configurationpolicy_controller.go
```diff
@@ -932,6 +932,23 @@ func sortRelatedObjectsAndUpdate(
...
+
+		if oldEntry.Properties != nil && (*oldEntry.Properties.CreatedByPolicy) {
```
JustinKuli, line 939: If the delete option is ever `DeleteAll`, then it wouldn't matter if the object was created by the policy or not, right? So I'm wondering if this `if` statement should be included at all.

ChunxiAlexLuo:
> @willkutler and @JustinKuli, for case A outlined in the PR comment, should a policy that is modified to enforce a different object clean up the old object when `pruneObjectBehavior` is set? I'm personally on the fence about this because the behavior of `pruneObjectBehavior` is on policy deletion and not on policy modification.
> 
> I think ArgoCD would clean up the object in this case and it may be worthwhile making this consistent with ArgoCD since I imagine this is the behavior that users may expect: https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#resources-prune-deletion-propagation-policy

In my opinion, case A and B don't have too much difference,  case A is modifying the policy to create the second pod in the same namespace, and case B is changing the policy to create the second pod in a different namespace. Do we want to: 1) just fix case B but not case B, or 2) fix both A) and B), or 3) don't fix any of them?

JustinKuli (commented):
I think this is the right approach for now. Correct me if I'm wrong, but basically this ensures that related objects are never removed from the list, so they can be considered for deletion when the policy is (eventually) deleted?

Matt brings up a good point - users might expect that these 'old related' objects might be deleted based on the `prune` behavior, but the original design is specific that these would only be cleaned up on **policy deletion**.

I think there could be a few additional things we want to do here in the future:
- add another property to note this situation (so users can identify these old objects)
- add an option to "eagerly" clean up these objects (like the ArgoCD behavior, probably)

ChunxiAlexLuo:
> I think this is the right approach for now. Correct me if I'm wrong, but basically this ensures that related objects are never removed from the list, so they can be considered for deletion when the policy is (eventually) deleted?
> 
> Matt brings up a good point - users might expect that these 'old related' objects might be deleted based on the `prune` behavior, but the original design is specific that these would only be cleaned up on **policy deletion**.
> 
> I think there could be a few additional things we want to do here in the future:
> 
> * add another property to note this situation (so users can identify these old objects)
> * add an option to "eagerly" clean up these objects (like the ArgoCD behavior, probably)

I agree with @JustinKuli, so two options: 1) don't fix the original issue and just update the note/documentation to mention this clean-up isn't within the pruneObjectBehavior scope
or 
2) add a flag here to optionally clean up these objects

Any other option?

controllers/configurationpolicy_controller.go
ChunxiAlexLuo: If we still decide to fix the original issue, yes, need to remove this and keep all objects in the `relatedObjects`, then determined by `DeleteAll` or `DeleteIfCreated` in `cleanUpChildObjects`

mprahl:
> > I think this is the right approach for now. Correct me if I'm wrong, but basically this ensures that related objects are never removed from the list, so they can be considered for deletion when the policy is (eventually) deleted?
> > Matt brings up a good point - users might expect that these 'old related' objects might be deleted based on the `prune` behavior, but the original design is specific that these would only be cleaned up on **policy deletion**.
> > I think there could be a few additional things we want to do here in the future:
> > 
> > * add another property to note this situation (so users can identify these old objects)
> > * add an option to "eagerly" clean up these objects (like the ArgoCD behavior, probably)
> 
> I agree with @JustinKuli, so two options: 1) don't fix the original issue and just update the note/documentation to mention this clean-up isn't within the pruneObjectBehavior scope or 2) add a flag here to optionally clean up these objects
> 
> Any other option?

I don't have a strong opinion, but my slight preference is to leave the current behavior and in 2.7 to add an option to clean up on updates to the policy. My thought is that it's safer to err on the side of not deleting and reinforcing the mindset that deletions only occur to the objects the policy currently applies to when it's deleted.

What do you all think?

PR #63, opened 2022-08-22
Perform validation of object definitions

mprahl:
Each separate change is separated by commits.

This also adds locking when the discovery information is refreshed to avoid race conditions when concurrency is enabled.

This causes a policy to be non-compliant when the objectDefinition fails
OpenAPI validation to prevent an infinite enforce loop where fields are
ignored by the Kubernetes API when the object is created/updated and
in the next loop, the ConfigurationPolicy sees that the object from the
API doesn't match what's in the policy and enforces again.

Relates:
https://github.com/stolostron/backlog/issues/24973
https://bugzilla.redhat.com/show_bug.cgi?id=2116528


mprahl:
@JustinKuli could you please double check where I placed the locking? Note that I chose not to use `defer` to unlock the lock to reduce the time that the lock is taken up and avoiding potential mistakes where a method that has a read lock calls a method with a read/write lock.

mprahl:
This is marked WIP since I need to add this validation on create as well and also fix the CI.

willkutler (approved):

PR #69, opened 2022-10-07
Allow enforcing policies that define a status with missing required fields

mprahl:
The status in the objectDefinition is ignored when in enforced mode. This allows a user to switch a policy between inform and enforce without having to remove the status check.

Note that the additional commit is to fix the case 20 test which was blocking the CI from passing.

Relates:
https://github.com/stolostron/backlog/issues/26155 https://bugzilla.redhat.com/show_bug.cgi?id=2132396

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -25,12 +25,14 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	apimachineryerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
 	"k8s.io/client-go/tools/record"
+	kubeopenapivalidation "k8s.io/kube-openapi/pkg/util/proto/validation"
 	"k8s.io/kubectl/pkg/util/openapi"
 	openapivalidation "k8s.io/kubectl/pkg/util/openapi/validation"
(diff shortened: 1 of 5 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -67,6 +70,11 @@ var (
 	reasonCleanupError       = "Error cleaning up child objects"
 )
 
+// validationErrIgnorePath is a regex to ignore paths from MissingRequiredFieldError errors when enforcing a policy.
+// These paths are in the format of "Pod.status.conditions[0]". This ignores any missing fields under status since those
+// fields are never actually updated on enforce.
+var validationErrIgnorePath = regexp.MustCompile(`.+\.status\..*`)
```
dhaiducek, line 76: This feels too broad. Wouldn't this capture any `status` field, even if it's buried under `spec` or `data`?

dhaiducek: Thanks! That does seem much better!

What I'm not sure on is whether the error path might update to include the group/version along with the kind to be clearer in cases where it's not a core resource or there are multiple kinds of the same name, something like `io.open-cluster-management.v1.Policies` rather than just `Policies`. Does the path always only have the Kind name and nothing else? I couldn't figure that out for sure.

mprahl: So I tried this `ConfigurationPolicy`:
```yaml
apiVersion: policy.open-cluster-management.io/v1
kind: ConfigurationPolicy
metadata:
  name: case23-policy-missing-status-fields
spec:
  remediationAction: enforce
  object-templates:
    - complianceType: musthave
      objectDefinition:
        apiVersion: policy.open-cluster-management.io/v1
        kind: Policy
        metadata:
          name: nginx-case23-missing-status
          namespace: default
```

It's missing the spec in the `Policy` and the `Path` on the `ValidationError` is `Policy`. `Path` for the `MissingRequiredFieldError` is `io.open-cluster-management.policy.v1.Policy`. Since this code uses `Path` on `ValidationError`, this should be okay.

willkutler (approved):

PR #70, opened 2022-10-12
potential role merge issue fix

willkutler:
https://github.com/stolostron/backlog/issues/26055

The first fix (for developer-a in the BZ) is changing the sort method we use when comparing lists. Previously we sorted lists at the top level first and then sorted any lower-level lists second, but that could cause problems if the sorting of the lower-level lists affected the top-level sort. This PR adds a recursive sort function that will sort the lowest level lists in an object first, which should eliminate this issue. In the future, it might make sense to use this sort function elsewhere in the code if it can make some parts of the code cleaner.

The second fix (for developer-original in the BZ) is adding a new `mergeSubLIsts` bool to the functions in the merge logic. The idea with this variable is that the merge functions are set to attempt to merge any lists into one when the compliance type is set to `musthave`, which can cause issues with a field like `role.rules`, for which we want to treat each sub-list as a complete entity and do not want to merge the lists of apiGroups, resources, or verbs between rules. So we set `mergeSubLists` to false when the key being processed is `rules`, and the merge function will only merge the rules list in the template with the existing rules list, rather than trying to merge the fields inside each rule as well.

The second fix needs some more testing to ensure it doesn't break anything, and it is likely that changing the merge in the controller to use strategic merge would produce cleaner code, but this solution should result in less code being changed. If we decide to use this fix, I'd still like to open an issue to investigate replacing the majority of the merge logic in the controller with strategic merge in 2.7 or a later release - it would make the code easier to maintain and debug than the current implementation.

### DIFF
--- a/api/v1/groupversion_info.go
+++ b/api/v1/groupversion_info.go
@@ -2,8 +2,8 @@
 // Copyright Contributors to the Open Cluster Management project
 
 // Package v1 contains API Schema definitions for the policy v1 API group
-//+kubebuilder:object:generate=true
-//+groupName=policy.open-cluster-management.io
+// +kubebuilder:object:generate=true
+// +groupName=policy.open-cluster-management.io
 package v1
 
 import (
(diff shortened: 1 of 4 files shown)


mprahl:
This got us most of the way there but we ended up going with #71.

PR #71, opened 2022-10-13
Favor the existing array ordering and content when comparing arrays 

mprahl:
`mergeArrays` had two issues. One example issue is when all array entries (N) in the policy were a subset of a single existing array entry, it would lead to the existing array entry being duplicated N times. Then all the other values in the existing array would be appended.

The solution to that is to stop looking for policy array entries that match the existing array entry after the match has been found. In the case there are duplicate array entries in the existing array, then it stops after the number of matches equals the number of duplicates.

The other issue is that the existing array and the policy array were not compared in the same order, so if there were policy array entries that were subsets of an existing array entry, then the merged array could be merged out of order.

Additionally in mergeSpecsHelper, if the values of the map were arrays, then the policy value was sorted but not the existing value when the arrays were merged. This could lead to the merged array being in a different order than the existing array, even though they were the same. The equality check run later would sort the array of maps of arrays in the policy value and existing value differently and cause an erroneous mismatch.

Relates:
https://bugzilla.redhat.com/show_bug.cgi?id=2130985 https://github.com/stolostron/backlog/issues/26055

Co-authored-by: Will Kutler <wkutler@redhat.com>


willkutler:
/lgtm

I think we should use this PR as it is cleaner

willkutler (approved):

PR #77, opened 2022-11-04
Add metric to monitor common related objects

dhaiducek:
This adds a `common_related_objects` metric, which is a Gauge Vector sliced by the related object and the policy. Each policy and related object pair has a gauge set to the total number of instances of that related object. For example, if two policies point to a related object, there would be two gauges for each policy/related object pair, each set to `2`. Gauges set to `1` are ignored/cleaned up.

ref:
- https://github.com/stolostron/backlog/issues/25357

### DIFF
--- a/controllers/metric.go
+++ b/controllers/metric.go
@@ -1,0 +1,60 @@
+package controllers
+
+import (
+	"fmt"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	evalLoopHistogram = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "config_policies_evaluation_duration_seconds",
+			Help:    "The seconds that it takes to evaluate all configuration policies on the cluster",
+			Buckets: []float64{1, 3, 9, 10.5, 15, 30, 60, 90, 120, 180, 300, 450, 600},
+		},
+	)
+	policyEvalSecondsCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "config_policy_evaluation_seconds_total",
+			Help: "The total seconds taken while evaluating the configuration policy. Use this alongside " +
+				"config_policy_evaluation_total.",
+		},
+		[]string{"name"},
+	)
+	policyEvalCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "config_policy_evaluation_total",
+			Help: "The total number of evaluations of the configuration policy. Use this alongside " +
+				"config_policy_evaluation_seconds_total.",
+		},
+		[]string{"name"},
+	)
+	compareObjSecondsCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "compare_objects_seconds_total",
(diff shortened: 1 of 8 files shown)


dhaiducek:
I'm curious about the performance hit for this implementation and wondering whether there should be a flag to shut off this metric (or maybe all metrics...).

dhaiducek:
Sorry for the churn here. CI is passing and this PR is ready for review.

dhaiducek:
I also wound up adding in a flag, `enable-metrics` to disable this metric if desired.

gparvin (commented):
Only had a minor comment about a comment.  Looks good.

test/utils/utils.go
```diff
@@ -242,3 +243,52 @@ func GetLastEvaluated(configPolicy *unstructured.Unstructured) (string, int64) {
 
 	return lastEvaluated, lastEvaluatedGeneration
 }
+
+// GetMetrics execs into the propagator pod and curls the metrics endpoint, filters
```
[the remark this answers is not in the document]
dhaiducek: Good catch on my copy/paste mistake!

controllers/metric.go
```diff
@@ -0,0 +1,110 @@
...
+			Name: "compare_objects_evaluation_total",
+			Help: "The total number of times the comparison algorithm is run on an object. " +
+				"Use this alongside compare_objects_seconds_total.",
+		},
+		[]string{"config_policy_name", "namespace", "object"},
+	)
+	// The policyRelatedObjectMap collects a map of related objects to policies
+	// in order to populate the gauge:
+	//   <kind.version/namespace/name>: []<policy-namespace/policy-name>
+	policyRelatedObjectMap   map[string][]string
```
mprahl, line 55: Could you please use a concurrency safe map implementation? Something like https://pkg.go.dev/sync#Map or using your own mutex.

dhaiducek: Good point--I'll update it.

gparvin (approved):

JustinKuli:
I'm slightly worried about the performance impact of *all* our metrics, not just this one. So I'm glad this is configurable.

From https://prometheus.io/docs/practices/naming/#labels : 
> CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

Each of our metrics has a `policy_name` label of some kind, plus this new one and `compare_objects*` have basically an `object_name` label. Those are the kind of "unbounded sets of values" that are warned against... I don't have a better solution, especially for this exact feature, I'm just commenting that we need to be careful.

dhaiducek:
I agree--this definitely felt like a "square peg in a round hole" kind of thing since Prometheus isn't really designed for the goals of the metric. (We almost need a separate controller that can handle this sort of object duplication, but that might be a one-off at this point.) Hopefully this metric wouldn't occur frequently, but I definitely wanted to be able to turn it off if it became unruly.

dhaiducek:
@JustinKuli @gparvin Are we prepared to unhold this PR, particularly given that it can be disabled? 

I think we might revisit this metric and ones like it in the future to see if we can generate them, potentially outside of Prometheus, in a manner more fitting to their intentions.

PR #78, opened 2022-11-08
Limit policy status event messages to 1024 characters

mprahl:
Relates:
https://issues.redhat.com/browse/ACM-2001

### DIFF
--- a/controllers/configurationpolicy_controller_test.go
+++ b/controllers/configurationpolicy_controller_test.go
@@ -236,6 +236,8 @@ func TestConvertPolicyStatusToString(t *testing.T) {
 		compliantDetails = append(compliantDetails, compliantDetail)
 	}
 
+	samplePolicy := getSamplePolicy()
+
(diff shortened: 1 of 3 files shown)


controllers/configurationpolicy_controller_test.go
```diff
@@ -246,6 +248,28 @@ func TestConvertPolicyStatusToString(t *testing.T) {
...
+		CompliancyDetails: []policyv1.TemplateStatus{
+			{
+				ComplianceState: policyv1.NonCompliant,
+				Conditions:      []policyv1.Condition{{Message: msg}},
+			},
+		},
+	}
+	statusMsg := convertPolicyStatusToString(&samplePolicy)
+
+	assert.Contains(t, statusMsg, "...")
```
JustinKuli, line 270: Is it possible to also assert its length?

mprahl: Good point!

PR #79, opened 2022-11-10
Add description for pruneObjectBehavior

gparvin:
Adding a description to the pruning parameter in the CRD

Refs:
 - https://github.com/stolostron/backlog/issues/26528

Signed-off-by: Gus Parvin <gparvin@redhat.com>

### DIFF
--- a/api/v1/configurationpolicy_types.go
+++ b/api/v1/configurationpolicy_types.go
@@ -27,7 +27,8 @@ type RemediationAction string
 // +kubebuilder:validation:Enum=low;Low;medium;Medium;high;High;critical;Critical
 type Severity string
 
-// PruneObjectBehavior : DeleteAll, DeleteIfCreated, or None
+// PruneObjectBehavior is used to remove objects that are managed by the
+// policy upon policy deletion.
 // +kubebuilder:validation:Enum=DeleteAll;DeleteIfCreated;None;
 type PruneObjectBehavior string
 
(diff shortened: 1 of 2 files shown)


api/v1/configurationpolicy_types.go
```diff
@@ -27,7 +27,8 @@ type RemediationAction string
 // +kubebuilder:validation:Enum=low;Low;medium;Medium;high;High;critical;Critical
 type Severity string
 
-// PruneObjectBehavior : DeleteAll, DeleteIfCreated, or None
+// PruneObjectBehavior is used to clean up resources that are created by the
+// policy. Values: DeleteAll, DeleteIfCreated, or None (default)
```
mprahl, line 31: I think the values come up nicely with `kubectl explain` since the enum is set.
```suggestion
// policy upon policy deletion.
```

api/v1/configurationpolicy_types.go
```diff
@@ -27,7 +27,8 @@ type RemediationAction string
 // +kubebuilder:validation:Enum=low;Low;medium;Medium;high;High;critical;Critical
 type Severity string
 
-// PruneObjectBehavior : DeleteAll, DeleteIfCreated, or None
+// PruneObjectBehavior is used to remove objects that are created by the
```
mprahl, line 30:
```suggestion
// PruneObjectBehavior is used to remove objects that are managed by the
```

mprahl (approved):

PR #81, opened 2022-11-23
Add a metric to measure how long resolving policy templates take

ChunxiAlexLuo:
Signed-off-by: Chunxi Luo <chuluo@redhat.com>

https://github.com/stolostron/backlog/issues/25358

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -127,6 +127,8 @@ func (r *ConfigurationPolicyReconciler) Reconcile(ctx context.Context, request c
 		// If the metric was not deleted, that means the policy was never evaluated so it can be ignored.
 		_ = policyEvalSecondsCounter.DeleteLabelValues(request.Name)
 		_ = policyEvalCounter.DeleteLabelValues(request.Name)
+		_ = plcTempsProcessSecondsCounter.DeleteLabelValues(request.Name)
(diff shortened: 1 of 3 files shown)


gparvin (commented):
Can we get a test added for this?

controllers/configurationpolicy_controller.go
```diff
@@ -853,6 +877,11 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 				objectT.ObjectDefinition.Raw = resolvedTemplate.ResolvedJSON
 			}
 		}
+
+		endTime := time.Now().UTC()
+		durationSeconds := float64(endTime.Sub(startTime)) / float64(time.Second)
```
JustinKuli, line 882: Nit: I think `time.Since(startTime).Seconds()` would be more clear, and could be more performant, although I don't think there would be a big difference.

JustinKuli (commented):
We might want this behind a flag like https://github.com/open-cluster-management-io/config-policy-controller/pull/77 has, so the metric could be turned off.

ChunxiAlexLuo:
STEP: Checking metric endpoint for policy template counter for policy policy-update-referenced-object
STEP: Policy policy-update-referenced-object total template process counter : 5 
STEP: Policy policy-update-referenced-object total template process seconds : 0.010091988

gparvin (approved):
Thanks Alex, this looks good.

PR #85, opened 2022-12-05
Remove confusing Undetermined compliance state

JustinKuli:
The situation where this compliance state was being used was appearing if the ConfigurationPolicy had no templates. Marking it as Unknown is better, since at least that is a value listed in the api folder...

This change also makes this situation create a compliance event, since the framework *needs* a new event in all situations when a policy, else when it transitions from Pending to something else, it will not update.

Refs:
 - https://issues.redhat.com/browse/ACM-2263


dhaiducek (approved):
LGTM!

PR #87, opened 2023-01-04
Use a single dynamic client in the reconciler

mprahl:
This makes the code simpler and removes the overhead of creating separate copies of API configs and dynamic clients on each object being evaluated.

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -107,8 +107,9 @@ type ConfigurationPolicyReconciler struct {
 	InstanceName          string
 	// The Kubernetes client to use when evaluating/enforcing policies. Most times, this will be the same cluster
 	// where the controller is running.
-	TargetK8sClient kubernetes.Interface
-	TargetK8sConfig *rest.Config
+	TargetK8sClient        kubernetes.Interface
+	TargetK8sDynamicClient dynamic.Interface
+	TargetK8sConfig        *rest.Config
 	// Whether custom metrics collection is enabled
 	EnableMetrics bool
 	discoveryInfo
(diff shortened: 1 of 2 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -561,19 +559,23 @@ func (r *ConfigurationPolicyReconciler) cleanUpChildObjects(plc policyv1.Configu
...
 func (r *ConfigurationPolicyReconciler) definitionIsDeleting() (bool, error) {
-	key := types.NamespacedName{Name: "configurationpolicies.policy.open-cluster-management.io"}
-	v1def := extensionsv1.CustomResourceDefinition{}
+	crdGVR := schema.GroupVersionResource{
+		Group: "apiextensions.k8s.io", Version: "v1", Resource: "customresourcedefinitions",
+	}
+	crdName := "configurationpolicies.policy.open-cluster-management.io"
 
-	v1err := r.Get(context.TODO(), key, &v1def)
+	v1def, v1err := r.TargetK8sDynamicClient.Resource(crdGVR).Get(context.TODO(), crdName, metav1.GetOptions{})
```
JustinKuli, line 567: I think the CRD will be on the hosting cluster, not the target cluster.

Also, I'm wondering if going through the dynamic client removes the need for https://github.com/open-cluster-management-io/config-policy-controller/pull/88 ? Since I can't tell how the cache settings there would get to the dynamic client (but maybe I'm missing something)

JustinKuli (changes requested):

controllers/configurationpolicy_controller.go
mprahl: Sorry about that. I didn't mean to commit this part.

mprahl: I was originally going towards the route of using the dynamic client but then realized it did actually need to be a client pointing to the cluster the config-policy-controller is running on. I decided to not have yet another client on the reconciler and figured the caching provided by #88 might be a good thing rather than query for the CRD on every ConfigurationPolicy deletion.

The latest push removes the code that was accidentally committed.

PR #89, opened 2023-01-05
Event on object update

willkutler:
Fires an event when an object is updated by the config policy controller, rather than just when one is created or deleted.

refs:
- https://issues.redhat.com/browse/ACM-2021?filter=-1


willkutler:
closing to start another branch since I think that will be easier

PR #90, opened 2023-01-06
Fire event on object update

willkutler:
Fires an event when an object is updated by the config policy controller, rather than just when one is created or deleted. Also changes events sent by successful enforce actions to compliant and adds a noncompliant "object not found" event before the enforce event.

refs:
- https://issues.redhat.com/browse/ACM-2021?filter=-1

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -1459,6 +1459,17 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 	if !exists && obj.shouldExist {
 		// it is a musthave and it does not exist, so it must be created
 		if strings.EqualFold(string(remediation), string(policyv1.Enforce)) {
+			// object is missing, so send noncompliant event
+			_ = createStatus("", obj.gvr.Resource, compliantObject, obj.namespaced, obj.policy,
+				obj.index, false, true)
+			obj.policy.Status.ComplianceState = policyv1.NonCompliant
+			statusStr := convertPolicyStatusToString(obj.policy)
+			objLog.Info("Sending an update policy status event", "policy", obj.policy.Name, "status", statusStr)
+			r.Recorder.Event(obj.policy, eventWarning, fmt.Sprintf(eventFmtStr, obj.policy.GetName(), obj.name),
+				statusStr)
+			// update parent policy status
+			r.addForUpdate(obj.policy, true)
+
 			var uid string
 			statusUpdateNeeded, uid, err = r.enforceByCreatingOrDeleting(obj)
 
@@ -1471,6 +1482,7 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 					CreatedByPolicy: &created,
 					UID:             uid,
 				}
+				compliant = true
 			}
 		} else { // inform
 			compliant = false
@@ -1514,7 +1526,21 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 
 		before := time.Now().UTC()
 
-		throwSpecViolation, msg, pErr := r.checkAndUpdateResource(obj, compType, mdCompType, remediation)
+		throwSpecViolation, msg, pErr, triedUpdate, updatedObj := r.checkAndUpdateResource(obj, compType, mdCompType,
+			remediation)
+
+		if triedUpdate {
+			// object has a mismatch and needs an update to be enforced, throw violation for mismatch
+			_ = createStatus("", obj.gvr.Resource, compliantObject, obj.namespaced, obj.policy, obj.index,
+				false, true)
+			obj.policy.Status.ComplianceState = policyv1.NonCompliant
+			statusStr := convertPolicyStatusToString(obj.policy)
+			objLog.Info("Sending an update policy status event", "policy", obj.policy.Name, "status", statusStr)
+			r.Recorder.Event(obj.policy, eventWarning, fmt.Sprintf(eventFmtStr, obj.policy.GetName(), obj.name),
+				statusStr)
+			// update parent policy status
+			r.addForUpdate(obj.policy, true)
+		}
 
 		duration := time.Now().UTC().Sub(before)
 		seconds := float64(duration) / float64(time.Second)
@@ -1538,8 +1564,17 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 			// it is a must have and it does exist, so it is compliant
 			compliant = true
 			if strings.EqualFold(string(remediation), string(policyv1.Enforce)) {
-				statusUpdateNeeded = createStatus("", obj.gvr.Resource, compliantObject, obj.namespaced, obj.policy,
-					obj.index, compliant, true)
+				if updatedObj {
+					// object updated in checkAndUpdateResource, send event
+					reason := "K8s update success"
+					idStr := identifierStr([]string{obj.name}, obj.namespace, obj.namespaced)
+					msg := fmt.Sprintf("%v %v was updated successfully", obj.gvr.Resource, idStr)
+
+					statusUpdateNeeded = addConditionToStatus(obj.policy, obj.index, true, reason, msg)
+				} else {
+					statusUpdateNeeded = createStatus("", obj.gvr.Resource, compliantObject, obj.namespaced, obj.policy,
+						obj.index, compliant, true)
+				}
 				created := false
 				creationInfo = &policyv1.ObjectProperties{
 					CreatedByPolicy: &created,
@@ -1559,7 +1594,14 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 			compliant = false
 		}
 
+		if compliant {
+			obj.policy.Status.ComplianceState = policyv1.Compliant
+		} else {
+			obj.policy.Status.ComplianceState = policyv1.NonCompliant
+		}
+
 		statusStr := convertPolicyStatusToString(obj.policy)
+
 		log.V(1).Info("Sending an update policy status event", "policy", obj.policy.Name, "status", statusStr)
 		r.Recorder.Event(obj.policy, eventType, fmt.Sprintf(eventFmtStr, obj.policy.GetName(), obj.name), statusStr)
 
@@ -1845,6 +1887,8 @@ func (r *ConfigurationPolicyReconciler) enforceByCreatingOrDeleting(obj singleOb
 			if !uidIsString || err != nil {
 				log.Error(err, "Tried to set UID in status but the field is not a string")
 			}
+
+			completed = true
 		}
 	} else {
 		log.Info("Enforcing the policy by deleting the object")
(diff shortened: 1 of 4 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -1538,8 +1538,19 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
...
-					obj.index, compliant, true)
+				if updatedObj {
+					// object updated in checkAndUpdateResource, send event
+
+					log.Info("Updated must have object", "resource", obj.gvr.Resource, "name", obj.name)
+					reason := "K8s update success"
+					idStr := identifierStr([]string{obj.name}, obj.namespace, obj.namespaced)
+					msg := fmt.Sprintf("%v %v was updated successfully", obj.gvr.Resource, idStr)
+
+					statusUpdateNeeded = addConditionToStatus(obj.policy, obj.index, true, reason, msg)
```
mprahl, line 1549: I'm wondering if this should remain compliant here since that is different behavior than when the object is created by a musthave ConfigurationPolicy.

The switching from compliant to noncompliant can be a signal for a user to investigate why it had changed.

controllers/configurationpolicy_controller.go
```diff
@@ -1538,8 +1538,19 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 			// it is a must have and it does exist, so it is compliant
 			compliant = true
 			if strings.EqualFold(string(remediation), string(policyv1.Enforce)) {
-				statusUpdateNeeded = createStatus("", obj.gvr.Resource, compliantObject, obj.namespaced, obj.policy,
-					obj.index, compliant, true)
+				if updatedObj {
+					// object updated in checkAndUpdateResource, send event
+
+					log.Info("Updated must have object", "resource", obj.gvr.Resource, "name", obj.name)
```
mprahl, line 1544: This log message is already present in `checkAndUpdateResource`.

controllers/configurationpolicy_controller.go
```diff
@@ -1538,8 +1538,19 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
...
+					reason := "K8s update success"
```
mprahl, line 1545: How about this?
```suggestion
					reason := "Object was updated"
```

test/e2e/case27_showupdateinstatus_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"open-cluster-management.io/config-policy-controller/test/utils"
+)
+
+const (
+	case27ConfigPolicyName string = "policy-pod-create"
+	case27CreateYaml       string = "../resources/case27_showupdateinstatus/case27-create-pod-policy.yaml"
+	case27UpdateYaml       string = "../resources/case27_showupdateinstatus/case27-update-pod-policy.yaml"
+)
+
+var _ = Describe("Test cluster version obj template handling", func() {
```
mprahl, line 19: Could you please update these test descriptions to match the test case?

test/e2e/case27_showupdateinstatus_test.go
```diff
@@ -0,0 +1,52 @@
...
+			utils.Kubectl("apply", "-f", case27UpdateYaml, "-n", testNamespace)
+			Eventually(func() interface{} {
+				managedPlc := utils.GetWithTimeout(clientManagedDynamic, gvrConfigPolicy,
+					case27ConfigPolicyName, testNamespace, true, defaultTimeoutSeconds)
+
+				return utils.GetStatusMessage(managedPlc)
+			}, 30, 0.5).Should(Equal(
+				"pods [nginx-pod-e2e] in namespace default was updated successfully"))
+		})
+		It("Cleans up", func() {
```
mprahl, line 47: Could you put this in an `AfterAll` instead?

test/e2e/case27_showupdateinstatus_test.go
```diff
@@ -0,0 +1,52 @@
...
+				managedPlc := utils.GetWithTimeout(clientManagedDynamic, gvrConfigPolicy,
+					case27ConfigPolicyName, testNamespace, true, defaultTimeoutSeconds)
+
+				return utils.GetStatusMessage(managedPlc)
+			}, 120, 1).Should(Equal(
+				"pods [nginx-pod-e2e] in namespace default found as specified, " +
+					"therefore this Object template is compliant"))
+		})
+		It("should be updated properly on the managed cluster", func() {
+			By("Creating " + case27ConfigPolicyName + " on managed")
```
mprahl, line 37:
```suggestion
			By("Updating " + case27ConfigPolicyName + " on managed")
```

test/resources/case27_showupdateinstatus/case27-create-pod-policy.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: policy.open-cluster-management.io/v1
```
mprahl, line 1: Optional: It'd be nice if this didn't use a pod since the pod requires pulling down a container image from the internet to run the test. Perhaps a ConfigMap would be a better option?

controllers/configurationpolicy_controller.go
willkutler: @mprahl we have "K8s creation success" and "K8s deletion success" for the creation and deletion conditions- I'd prefer to keep it this way to keep it consistent if that's ok with you

controllers/configurationpolicy_controller.go
willkutler: I'm confused - this is compliant, right? I have compliant set to true and when I tested it the status was compliant

test/resources/case27_showupdateinstatus/case27-create-pod-policy.yaml
willkutler: I changed this but was wondering why we don't want to use pods since we have a lot of tests that use them - is it just to try speed up the test a bit?

controllers/configurationpolicy_controller.go
mprahl: @willkutler right, I was questioning if it actually should be compliant like you have it set to. I'm leaning towards making it noncompliant. Let's have a team Slack discussion on it.

test/resources/case27_showupdateinstatus/case27-create-pod-policy.yaml
mprahl: It's speed but it's also unnecessarily having our tests reaching out over the internet and downloading a decent sized artifact (the container image). It also running an additional container, so it's unnecessary resource utilization.

In general, tests shouldn't reach out to the internet at all unless it's required for the test. This makes them more resilient as all the test's dependencies are local. This particular case, we don't check if the pod is running or not, so even if the container registry was down, the test would still pass, but it's still not best practice based on what I stated earlier.

controllers/configurationpolicy_controller.go
mprahl: That's a good point. Let's keep it the way you have it in the PR.

controllers/configurationpolicy_controller.go
willkutler: Oh, that makes sense. Personally I would rather have the creation event be compliant and keep the update compliant, since those events aren't really violations. I'll make a post in the team slack channel about it.

controllers/configurationpolicy_controller.go
```diff
@@ -1459,6 +1459,19 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
...
 		// it is a musthave and it does not exist, so it must be created
 		if strings.EqualFold(string(remediation), string(policyv1.Enforce)) {
+			// object is missing, so send noncompliant event
+			updateParent := createStatus("", obj.gvr.Resource, compliantObject, obj.namespaced, obj.policy,
+				obj.index, false, true)
+			if updateParent {
+				eventType := eventWarning
+				obj.policy.Status.ComplianceState = policyv1.NonCompliant
+				statusStr := convertPolicyStatusToString(obj.policy)
+				log.Info("Sending an update policy status event", "policy", obj.policy.Name, "status", statusStr)
```
mprahl, line 1469: Perhaps you could use `objLog` here instead?

controllers/configurationpolicy_controller.go
```diff
@@ -2382,13 +2415,28 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
...
-				return true, "", false
+				return true, "", false, false
 			} else if isStatus {
 				statusUpdated = true
 				log.Info("Ignoring an update to the object status", "key", key)
 			} else {
 				updateNeeded = true
 				log.Info("Queuing an update for the object due to a value mismatch", "key", key)
+
+				updateParent := createStatus("", obj.gvr.Resource, map[string]map[string]interface{}{
```
mprahl, line 2426: Could you please move this code to the calling method (handleSingleObj)? That way this method is scoped to just updating the object if needed and the the calling method handles all the status updates.

This makes the responsibilities tighter and clearer for each method.

controllers/configurationpolicy_controller.go
```diff
@@ -1459,6 +1459,19 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
...
 		if strings.EqualFold(string(remediation), string(policyv1.Enforce)) {
+			// object is missing, so send noncompliant event
+			updateParent := createStatus("", obj.gvr.Resource, compliantObject, obj.namespaced, obj.policy,
+				obj.index, false, true)
+			if updateParent {
+				eventType := eventWarning
+				obj.policy.Status.ComplianceState = policyv1.NonCompliant
+				statusStr := convertPolicyStatusToString(obj.policy)
+				log.Info("Sending an update policy status event", "policy", obj.policy.Name, "status", statusStr)
+				r.Recorder.Event(obj.policy, eventType, fmt.Sprintf(eventFmtStr, obj.policy.GetName(), obj.name), statusStr)
```
mprahl, line 1470:
```suggestion
				r.Recorder.Event(obj.policy, eventWarning, fmt.Sprintf(eventFmtStr, obj.policy.GetName(), obj.name), statusStr)
```

controllers/configurationpolicy_controller.go
```diff
@@ -1459,6 +1459,19 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 	if !exists && obj.shouldExist {
 		// it is a musthave and it does not exist, so it must be created
 		if strings.EqualFold(string(remediation), string(policyv1.Enforce)) {
+			// object is missing, so send noncompliant event
+			updateParent := createStatus("", obj.gvr.Resource, compliantObject, obj.namespaced, obj.policy,
```
mprahl, line 1463: After thinking about this, I think createStatus needs to be able to append two statuses so that it just updates the `lastTransition` field on duplicates rather than always appending in the event an object is constantly updated each evaluation interval.

controllers/configurationpolicy_controller.go
```diff
@@ -2325,15 +2369,15 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
 	complianceType string,
 	mdComplianceType string,
 	remediation policyv1.RemediationAction,
-) (throwSpecViolation bool, message string, processingErr bool) {
+) (throwSpecViolation bool, message string, processingErr bool, updateNeeded bool, updateSucceeded bool) {
```
mprahl, line 2372: Could you please document these new return values in the function doc comment?

mprahl (approved):
Nice job!

PR #91, opened 2023-01-10
Fix the conditions history in compliancyDetails

mprahl:
The status.compliancyDetails[].conditions array was always limited to just a single condition due to a bug. This sets the limit to 10 instead.

Relates:
https://issues.redhat.com/browse/ACM-2708

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -48,8 +48,9 @@ import (
 )
 
 const (
-	ControllerName string = "configuration-policy-controller"
-	CRDName        string = "configurationpolicies.policy.open-cluster-management.io"
+	ControllerName                 string = "configuration-policy-controller"
+	CRDName                        string = "configurationpolicies.policy.open-cluster-management.io"
+	complianceStatusConditionLimit        = 10
 )
 
 var log = ctrl.Log.WithName(ControllerName)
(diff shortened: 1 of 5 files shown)


willkutler (approved):

dhaiducek (dismissed):

dhaiducek (dismissed):

dhaiducek:
I've been thinking about this, and I don't think it's a bug per se. This looks intentional since the status-sync is the thing that has the 10 history limit, so the Policy as the source of the history and the ConfigurationPolicy as only holding the current state. Though I could see the ConfigurationPolicy not having the same information as the Policy as a bug, but maybe we only need to address this in 2.7 as a result? Does the current behavior cause issues or is it just an inconsistency?

willkutler:
@dhaiducek it doesn't cause issues with the status of the policy at all (besides the history inconsistency with config policies) but I would still say it's a bug since we have a conditions array that isn't really being used. If we want to only hold the latest condition in the config policy and have policies be responsible for history, that would be fine but I think we'd want the `status.compliancyDetails[].conditions` field to be a map rather than an array of maps (this would be more of a 2.8 change). If we keep it as an array, I think it makes sense for it to hold a history, since that seems to be the intended behavior based on how we wrote the config policy CRD.

mprahl:
> I've been thinking about this, and I don't think it's a bug per se. This looks intentional since the status-sync is the thing that has the 10 history limit, so the Policy as the source of the history and the ConfigurationPolicy as only holding the current state. Though I could see the ConfigurationPolicy not having the same information as the Policy as a bug, but maybe we only need to address this in 2.7 as a result? Does the current behavior cause issues or is it just an inconsistency?

@dhaiducek the status-sync gets its history from status events sent by the ConfigurationPolicy, so the conditions array doesn't influence that. Having a history in the ConfigurationPolicy is still desirable as it contains more information than the status stored by the status-sync, in particular it shows information per object-template entry.

mprahl:
Sorry about the noise on this PR @dhaiducek and @willkutler, but it's rebased on #90 and the tests are now passing.

willkutler (approved):

dhaiducek (approved):
Thanks for the explanations/background! I think I was questioning whether a backport was necessary since the issue is targeted for 2.5 and 2.6, but it does seem like it'd be nice to have for debugging and consistency.

PR #95, opened 2023-01-19
Allow multiline templatization

willkutler:
Adds a new field, object-templates-raw, that allows users to paste in an object template. This makes multiline templates, like those using the range keyword, possible. It can be set instead of object-templates, but both cannot be set.

ref: https://issues.redhat.com/browse/ACM-2739?filter=-1

### DIFF
--- a/api/v1/configurationpolicy_types.go
+++ b/api/v1/configurationpolicy_types.go
@@ -157,8 +157,19 @@ type ConfigurationPolicySpec struct {
 	// 'matchLabels' and/or 'matchExpressions' are, 'include' will behave as if ['*'] were given. If
 	// 'matchExpressions' and 'matchLabels' are both not provided, 'include' must be provided to
 	// retrieve namespaces.
-	NamespaceSelector  Target             `json:"namespaceSelector,omitempty"`
-	ObjectTemplates    []*ObjectTemplate  `json:"object-templates,omitempty"`
+	NamespaceSelector Target `json:"namespaceSelector,omitempty"`
+	// 'object-templates' and 'object-templates-raw' are arrays of objects for the configuration
+	// policy to check, create, modify, or delete on the cluster. 'object-templates' is an array
+	// of objects, while 'object-templates-raw' is a string containing an array of objects in
+	// YAML format. Only one of the two object-templates variables can be set in a given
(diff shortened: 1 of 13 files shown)


test/e2e/case28_multiline_templatization_test.go
```diff
@@ -0,0 +1,115 @@
...
+			}
+		})
+
+		AfterAll(func() {
+			deleteConfigPolicies([]string{case28RangePolicyName})
+			utils.Kubectl("delete", "-f", case28ConfigMapsYaml)
+		})
+	})
+	Describe("Test invalid multiline templates", func() {
+		It("should generate noncompliant for invalid template strings", func() {
```
mprahl, line 86: Could you please check status messages here to ensure the test policy isn't non-compliant for another reason?

mprahl (changes requested):
Looks great! Just a few comments.

willkutler:
@mprahl thanks for the review! added a comment in the type file and an extra check for status message in the test case

api/v1/configurationpolicy_types.go
```diff
@@ -157,8 +157,14 @@ type ConfigurationPolicySpec struct {
...
 	// retrieve namespaces.
-	NamespaceSelector  Target             `json:"namespaceSelector,omitempty"`
+	NamespaceSelector Target `json:"namespaceSelector,omitempty"`
+	// 'object-templates' and 'object-templates-raw' are arrays of objects for the configuration
+	// policy to check, create, modify, or delete on the cluster. 'object-templates' is an array
+	// of objects, while 'object-templates-raw' is a string containing an array of objects in
+	// YAML format. Only one of the two object-templates variables can be set in a given
+	// configurationPolicy.
 	ObjectTemplates    []*ObjectTemplate  `json:"object-templates,omitempty"`
+	ObjectTemplatesRaw string             `json:"object-templates-raw,omitempty"`
```
mprahl, line 167: Could you add a description so that `kubectl explain ConfigurationPolicy.spec.object-templates-raw` will return something helpful? Even if it's just a copy of the above comment or just say to look at the docs for `object-templates`.

mprahl (approved):
Nice!

PR #96, opened 2023-01-19
Clean up when deployment is being deleted

JustinKuli:
Previously, when the controller is removed, it could leave a mess behind in the form of ConfigurationPolicy objects with finalizers on them. That would also result in the CRD being stuck.

Now, the controller will put a finalizer on its own deployment when at least one ConfigurationPolicy has a finalizer. Also, while the deployment is being deleted, the finalizers on ConfigurationPolicies will be removed immediately, in the same way they would be when the CRD was being deleted.

Note: in this implementation, related objects will *never* be pruned when the controller is removed.

Since the controller deployment is in a different namespace than the ConfigutaionPolicies it watches, the controller's cache cannot be limited to a single namespace the way it was before. Instead, this uses additional cache selectors to limit watches on ConfigurationPolicies and Deployments to the relevant namespaces.

Refs:
 - https://issues.redhat.com/browse/ACM-2923

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -18,6 +18,7 @@ import (
 	gocmp "github.com/google/go-cmp/cmp"
 	"github.com/prometheus/client_golang/prometheus"
 	templates "github.com/stolostron/go-template-utils/v3/pkg/templates"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	extensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	extensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
@@ -51,6 +52,7 @@ const (
 	ControllerName                 string = "configuration-policy-controller"
 	CRDName                        string = "configurationpolicies.policy.open-cluster-management.io"
 	complianceStatusConditionLimit        = 10
+	pruneObjectFinalizer           string = "policy.open-cluster-management.io/delete-related-objects"
 )
 
 var log = ctrl.Log.WithName(ControllerName)
@@ -186,6 +188,26 @@ func (r *ConfigurationPolicyReconciler) PeriodicallyExecConfigPolicies(freq uint
 			}
 		} else {
 			skipLoop = true
+		}
+
+		needDeploymentFinalizer := false
+
+		for _, plc := range policiesList.Items {
+			if objHasFinalizer(&plc, pruneObjectFinalizer) {
+				needDeploymentFinalizer = true
+
+				break
+			}
+		}
+
+		if err := r.manageDeploymentFinalizer(needDeploymentFinalizer); err != nil {
+			if errors.Is(err, common.ErrNoNamespace) || errors.Is(err, common.ErrRunLocal) {
+				log.Info("Not managing the controller's deployment finalizer because it is running locally")
+			} else {
+				log.Error(err, "Failed to manage the controller's deployment finalizer, skipping loop")
+
+				skipLoop = true
+			}
 		}
 
 		// This is done every loop cycle since the channel needs to be variable in size to account for the number of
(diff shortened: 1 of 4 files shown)


JustinKuli:
I'll be writing a test for this in our integration suite, since it primarily deals with the situation when the deployment is being deleted, and the tests here do not have a deployment to check (in the tests here, it runs locally)

JustinKuli:
Related integration test (draft): https://github.com/stolostron/governance-policy-framework/pull/541

JustinKuli:
/hold

Need to mess with cache settings so that the Deployment can actually be gotten.

controllers/configurationpolicy_utils.go
```diff
@@ -540,18 +541,19 @@ func configPlcHasFinalizer(plc policyv1.ConfigurationPolicy, finalizer string) b
...
+	if objHasFinalizer(obj, finalizer) {
+		return obj.GetFinalizers()
 	}
 
-	return append(plc.GetFinalizers(), finalizer)
+	return append(obj.GetFinalizers(), finalizer)
 }
 
-func removeConfigPlcFinalizer(plc policyv1.ConfigurationPolicy, finalizer string) []string {
+// nolint: unparam
```
mprahl, line 552: Is this nolint actually needed?

controllers/configurationpolicy_controller.go
```diff
@@ -648,31 +688,31 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 		// kick off object deletion if configurationPolicy has been deleted
 		if plc.ObjectMeta.DeletionTimestamp != nil {
 			// If the CRD is deleting, don't prune objects
-			crdDeleting, err := r.definitionIsDeleting()
+			cleanupNow, err := r.cleanupImmediately()
```
mprahl, line 691: Will this get triggered in hosted mode where only a single hosted cluster is deleted but one other remains? I think `plc.ObjectMeta.DeletionTimestamp != nil` would only be true if the CRD was being deleted.

controllers/configurationpolicy_utils.go
JustinKuli: The linter complained about it... We only have one finalizer, so it makes sense. The real mystery is why the linter doesn't complain about the other (similar) functions with this issue.

controllers/configurationpolicy_controller.go
JustinKuli: I was assuming the namespace would be deleted when the controller was being uninstalled, but I think you're right and there are situations that wouldn't be true.

This also made me realize a problem with the `cleanupImmediately` implementation where if there was an error getting the deployment, but the CRD is successfully found *not* to be deleting, it decides immediate cleanup is *not* necessary. But like you say, in hosted mode, that's the wrong decision. If it can't get the deployment, it can't decide that immediate cleanup is not necessary.

controllers/configurationpolicy_utils.go
mprahl: Oh, you think it's detecting that a constant is always passed as its parameter?

controllers/configurationpolicy_controller.go
```diff
@@ -565,6 +587,52 @@ func (r *ConfigurationPolicyReconciler) cleanUpChildObjects(plc policyv1.Configu
...
+	if defErr == nil && defDeleting {
+		return true, nil
+	}
+
+	if deployErr == nil && defErr == nil {
+		// if either was deleting, we would've already returned.
+		return false, nil
+	}
+
+	// At least one had an uxpected error, so the decision can't be made right now
```
mprahl, line 609:
```suggestion
	// At least one had an unexpected error, so the decision can't be made right now
```

main.go
```diff
@@ -163,29 +156,63 @@ func main() {
...
+			}),
+		}
+	}
+
+	watchNamespace, err := common.GetWatchNamespace()
+	if strings.Contains(watchNamespace, ",") {
+		err = fmt.Errorf("multiple watched namespaces are not allowed for this controller")
+	}
+
+	if err != nil {
```
mprahl, line 190: Optional: I prefer error handling upfront since it's a more common pattern.

mprahl (approved):
Looks great! I left a couple of optional comments.

/hold

main.go
JustinKuli: Let me know what you think of the update. This is kind of a weird one, basically I wanted to add an error case, but I didn't want to change the behavior of `GetWatchNamespace` to return the error, because it was copied from somewhere else.

mprahl (approved):

PR #97, opened 2023-01-20
Fix the gosec:G601 error

mprahl:


JustinKuli (approved):
Yep, that's what I was about to do :D

PR #99, opened 2023-01-23
Revert "Fix the conditions history in compliancyDetails"

mprahl:
This reverts commit 0234403dd4b34b2fd410ed25ef19fb95cadb6732 since conditions aren't meant to contain a history. They are supposed to contain just the current state. See:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

The ConfigurationPolicy status events combined all conditions and led to the wrong status message being stored in the Policy.


JustinKuli (approved):
/ltgm

PR #103, opened 2023-02-07
add label to designate config policy as template

willkutler:


JustinKuli (approved):
Can you add a link to related work for why this is necessary? Just to keep things traceable. And it looks like Prow wants you to rebase it, then I will approve again.

PR #104, opened 2023-02-07
Trigger uninstalls through a new annotation

mprahl:
Previous to this, a finalizer on the Deployment was added so that if the Deployment was deleted, it would handle immediate clean up. This doesn't handle the common case where the config-policy-controller ManagedClusterAddOn is deleted, which causes the ManifestWork to be deleted, which triggers all Configuration Policy controller deployment artifacts, including the service account.

A new approach is taken so that a new annotation of policy.open-cluster-management.io/uninstalling=true is set on the Deployment to indicate that the Configuration Policy controller should remove all finalizers because it's getting deleted.

The Policy Addon controller will be updated so that when the config-policy-controller ManagedClusterAddOn object is deleted, a finalizer will prevent it and a Pod will run on the managed cluster with the new `trigger-uninstall` subcommand. This sets the uninstalling annotation on the Deployment and then waits until all ConfigurationPolicy finalizers have been removed. Once the command ends, the Pod exits, and the ManagedClusterAddOn object's finalizer is removed and the uninstall proceeds.

Relates:
https://issues.redhat.com/browse/ACM-3233
https://issues.redhat.com/browse/ACM-2923

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -334,14 +334,18 @@ e2e-test: e2e-dependencies
 	$(GINKGO) -v --fail-fast --slow-spec-threshold=10s $(E2E_TEST_ARGS) test/e2e
 
 .PHONY: e2e-test-coverage
-e2e-test-coverage: E2E_TEST_ARGS = --json-report=report_e2e.json --label-filter='!hosted-mode' --output-dir=.
+e2e-test-coverage: E2E_TEST_ARGS = --json-report=report_e2e.json --label-filter='!hosted-mode && !running-in-cluster' --output-dir=.
 e2e-test-coverage: e2e-run-instrumented e2e-test e2e-stop-instrumented
 
 .PHONY: e2e-test-hosted-mode-coverage
-e2e-test-hosted-mode-coverage: E2E_TEST_ARGS = --json-report=report_e2e_hosted_mode.json --label-filter="hosted-mode" --output-dir=.
+e2e-test-hosted-mode-coverage: E2E_TEST_ARGS = --json-report=report_e2e_hosted_mode.json --label-filter="hosted-mode && !running-in-cluster" --output-dir=.
 e2e-test-hosted-mode-coverage: COVERAGE_E2E_OUT = coverage_e2e_hosted_mode.out
 e2e-test-hosted-mode-coverage: export TARGET_KUBECONFIG_PATH = $(PWD)/kubeconfig_managed2
 e2e-test-hosted-mode-coverage: e2e-run-instrumented e2e-test e2e-stop-instrumented
+
+.PHONY: e2e-test-running-in-cluster
+e2e-test-running-in-cluster: E2E_TEST_ARGS = --label-filter="running-in-cluster" --covermode=atomic --coverprofile=coverage_e2e_uninstall.out --coverpkg=open-cluster-management.io/config-policy-controller/pkg/triggeruninstall
+e2e-test-running-in-cluster: e2e-test
 
 .PHONY: e2e-build-instrumented
 e2e-build-instrumented:
--- a/main.go
+++ b/main.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/go-logr/zapr"
 	"github.com/spf13/pflag"
@@ -43,6 +44,7 @@ import (
 	policyv1 "open-cluster-management.io/config-policy-controller/api/v1"
 	"open-cluster-management.io/config-policy-controller/controllers"
 	"open-cluster-management.io/config-policy-controller/pkg/common"
+	"open-cluster-management.io/config-policy-controller/pkg/triggeruninstall"
 	"open-cluster-management.io/config-policy-controller/version"
 )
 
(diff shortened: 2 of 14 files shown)


pkg/triggeruninstall/triggeruninstall.go
```diff
@@ -0,0 +1,112 @@
...
+
+		_, err = deploymentRsrc.Update(ctx, deployment, metav1.UpdateOptions{})
+		if err != nil {
+			if k8serrors.IsServerTimeout(err) || k8serrors.IsTimeout(err) || k8serrors.IsConflict(err) {
+				klog.Infof("Retrying setting the Deployment uninstall annotation due to error: %s", err)
+
+				continue
+			}
+
+			return err
```
JustinKuli, line 59: Why only retry with certain errors?

main.go
```diff
@@ -372,3 +392,45 @@ func main() {
...
+	triggerUninstallFlagSet.AddGoFlagSet(flag.CommandLine)
+
+	_ = triggerUninstallFlagSet.Parse(os.Args[2:])
+
+	if deploymentName == "" || deploymentNamespace == "" || policyNamespace == "" {
+		fmt.Fprintln(os.Stderr, "--deployment-name, --deployment-namespace, --policy-namespace must all have values")
+		os.Exit(1)
+	}
+
+	terminatingCtx := ctrl.SetupSignalHandler()
```
JustinKuli, line 422: I'm wondering if this should have a maximum time before we cancel it.

test/e2e/case28_trigger_uninstall_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+			g.Expect(policy2.GetFinalizers()).To(ContainElement(pruneObjectFinalizer))
+		}, defaultTimeoutSeconds, 1).Should(Succeed())
+
+		By("Triggering an uninstall")
+		kubeconfigBytes, err := os.ReadFile(kubeconfigManaged)
+		Expect(err).To(BeNil())
+
+		config, err := clientcmd.RESTConfigFromKubeConfig(kubeconfigBytes)
+		Expect(err).To(BeNil())
```
JustinKuli, line 62: Does `LoadConfig` not work for this? (In the suite file)

test/e2e/case28_trigger_uninstall_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+		ctx, ctxCancel := context.WithDeadline(
+			context.Background(),
+			// Cancel the context after the default timeout seconds to cause TriggerUninstall to stop prematurely
+			time.Now().Add(time.Duration(defaultTimeoutSeconds)*time.Second),
```
JustinKuli, line 67: By cancelling the context automatically, this does not verify that TriggerUninstall would eventually complete on its own. Was there a problem with waiting for it?

JustinKuli (commented):
Nothing wrong, just questions/comments. I think this will work.

pkg/triggeruninstall/triggeruninstall.go
mprahl: I don't want the uninstall to stall out for an unrecoverable error and continuously query the Kubernetes API. The addon controller PR uses the Pod restart policy so that Kubernetes can handle restarts of such unhandled errors:
https://github.com/open-cluster-management-io/governance-policy-addon-controller/pull/60

This allows Kubernete's backoff mechanism to work and sometimes a restart could help such as when the instantiated Kubernetes client becomes stale, though unlikely. It also may alert the cluster administrator that there is a failing Pod that needs investigation.

main.go
mprahl: Sure, I'll set it to default to 5 minutes.

test/e2e/case28_trigger_uninstall_test.go
mprahl: Thanks for the pointer. I didn't see it there.

test/e2e/case28_trigger_uninstall_test.go
mprahl: This was to avoid the test running forever if `TriggerUninstall` misbehaves. I realize the comment is not clear. I'll update that.

pkg/triggeruninstall/triggeruninstall.go
JustinKuli: Got it, other retries are just handled by the container restarting.

JustinKuli (dismissed):
🤞

Makefile
```diff
@@ -334,15 +334,19 @@ e2e-test: e2e-dependencies
...
 
 .PHONY: e2e-test-hosted-mode-coverage
-e2e-test-hosted-mode-coverage: E2E_TEST_ARGS = --json-report=report_e2e_hosted_mode.json --label-filter="hosted-mode" --output-dir=.
+e2e-test-hosted-mode-coverage: E2E_TEST_ARGS = --json-report=report_e2e_hosted_mode.json --label-filter="hosted-mode && !running-in-cluster" --output-dir=.
 e2e-test-hosted-mode-coverage: COVERAGE_E2E_OUT = coverage_e2e_hosted_mode.out
 e2e-test-hosted-mode-coverage: export TARGET_KUBECONFIG_PATH = $(PWD)/kubeconfig_managed2
 e2e-test-hosted-mode-coverage: e2e-run-instrumented e2e-test e2e-stop-instrumented
 
+.PHONY: e2e-test-running-in-cluster
+e2e-test-running-in-cluster: E2E_TEST_ARGS = --label-filter="running-in-cluster" -covermode=atomic --coverprofile=coverage_e2e_uninstall.out --coverpkg=open-cluster-management.io/config-policy-controller/pkg/triggeruninstall
```
JustinKuli, line 347: I think it's `--covermode`, but we'll find out
```suggestion
e2e-test-running-in-cluster: E2E_TEST_ARGS = --label-filter="running-in-cluster" --covermode=atomic --coverprofile=coverage_e2e_uninstall.out --coverpkg=open-cluster-management.io/config-policy-controller/pkg/triggeruninstall
```

JustinKuli (dismissed):

Makefile
mprahl: The other way worked locally, but I missed that it wasn't consistent. I just fixed it.

JustinKuli (approved):

PR #105, opened 2023-02-08
Process object-templates-raw properly with no templatization

willkutler:
If object-templates-raw was set without including a template, it was not properly being converted into a normal object-templates array and thus was not generating a status properly. This change marshals the object-templates-raw string into an object for further processing.

It also renames the test case for object-templates-raw since we had two test 28s.


controllers/configurationpolicy_controller.go
```diff
@@ -1011,6 +1011,22 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
+				err := json.Unmarshal(resolvedTemplate.ResolvedJSON, &objTemps)
+				if err != nil {
+					addTemplateErrorViolation("Error unmarshalling raw template", err.Error())
+
+					return
+				}
+
+				plc.Spec.ObjectTemplates = objTemps
+
+				break
```
dhaiducek, line 1029: This looks like the same flow as the template logic above. Should the `if` above add an `|| isRawObjTemplate` instead or is there a reason to have this separate?

controllers/configurationpolicy_controller.go
```diff
@@ -1011,6 +1011,22 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
+					addTemplateErrorViolation("Error parsing raw template", err.Error())
+				}
+				err := json.Unmarshal(resolvedTemplate.ResolvedJSON, &objTemps)
+				if err != nil {
+					addTemplateErrorViolation("Error unmarshalling raw template", err.Error())
+
+					return
+				}
+
+				plc.Spec.ObjectTemplates = objTemps
```
mprahl, line 1027: @willkutler could you just directly use `gopkg.in/yaml.v3` and call unmarshall into `objTemps`?

controllers/configurationpolicy_controller.go
```diff
@@ -1011,6 +1011,22 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 
 				// Otherwise, set the resolved data for use in further processing
 				plc.Spec.ObjectTemplates[i].ObjectDefinition.Raw = resolvedTemplate.ResolvedJSON
+			} else if isRawObjTemplate {
+				// no templates to resolve, but the template resolver will handle YAML conversion
+				resolvedTemplate, tplErr := tmplResolver.ResolveTemplate(rawData, nil)
+				if tplErr != nil {
+					addTemplateErrorViolation("Error parsing raw template", err.Error())
+				}
+				err := json.Unmarshal(resolvedTemplate.ResolvedJSON, &objTemps)
```
mprahl, line 1020: It also seems that `objTemps` shouldn't be defined outside of the for loop since it's not used outside of it.

controllers/configurationpolicy_controller.go
willkutler: the flow is a bit different now after moving the yaml marshal into the controller

dhaiducek:
@willkutler Would you please squash the commits?

willkutler:
@dhaiducek done!

controllers/configurationpolicy_controller.go
```diff
@@ -1011,6 +1012,33 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
+				var yamlObj interface{}
+
+				err = yaml.Unmarshal(rawData, &yamlObj)
+				if err != nil {
+					addTemplateErrorViolation("Error unmarshaling YAML for raw template", err.Error())
+
+					return
+				}
+
+				resolvedJSON, err := json.Marshal(yamlObj)
```
mprahl, line 1026: Optional: I think you can use `sigs.k8s.io/yaml` instead to avoid the JSON Marshal and Unmarshal code. I assume you are doing this because we don't have `yaml` tags on the struct fields, and `sigs.k8s.io/yaml` handles that for you I believe.

controllers/configurationpolicy_controller.go
```diff
@@ -1011,6 +1012,33 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
+				if err != nil {
+					addTemplateErrorViolation("Error parsing raw template", err.Error())
+
+					return
+				}
+				err = json.Unmarshal(resolvedJSON, &objTemps)
+				if err != nil {
+					addTemplateErrorViolation("Error unmarshalling raw template", err.Error())
```
mprahl, line 1034:
```suggestion
					addTemplateErrorViolation("Error converting object-templates-raw to object-templates. Please verify the syntax.", err.Error())
```

test/e2e/case29_multiline_templatization_test.go
```diff
@@ -15,42 +15,44 @@ import (
 )
 
 const (
-	case28RangePolicyName string = "case28-configpolicy"
-	case28RangePolicyYaml string = "../resources/case28_multiline_templatization/case28_policy.yaml"
-	case28ConfigMapsYaml  string = "../resources/case28_multiline_templatization/case28_configmaps.yaml"
+	case29RangePolicyName      string = "case29-configpolicy"
```
mprahl, line 18: Good catch on the rename, but case29 is now taken in #104 .

willkutler:
@mprahl added those review fixes, thanks

mprahl (approved):

PR #107, opened 2023-03-02
Fix status for invalid object

willkutler:
Previously, a config policy with an invalid field in the object definition would cause the policy status to flip between "object not found" and "object could not be created/updated" indefinitely, which became an issue at scale. This change makes it so that an invalid template will only cause one violation event, which lines up with the single event sent out when a policy template has an invalid field.


JustinKuli (changes requested):
The new tests are passing locally for me even without the controller code change. I think the code change is correct, it just needs to be verified another way.

willkutler:
good catch @JustinKuli - the event time field I was checking wasn't actually set, which caused the .before to always pass and that check could never fail. I verified that the new test fails on older builds.

PR #115, opened 2023-03-30
change unit test

yiraeChristineKim:
Objects managed by ConfigurationPolicies are left behind if they are renamed

The pruneObjectBehavior of a ConfigurationPolicy does not help clean up objects when the policy is edited such that the managed object's name is changed. Ideally, editing the policy would result in deleting the old object, if the pruneObjectBehavior field was set.

added Dale's suggestion

ref: https://issues.redhat.com/browse/ACM-3050


willkutler (approved):

willkutler:
nice work Christine! /lgtm

PR #116, opened 2023-03-31
Bug: `ConfigurationPolicy` message for `enforce` omits objects when multiple namespaces are specified

yiraeChristineKim:
ConfigurationPolicy message for enforce omits objects when multiple namespaces are specified

When set to inform, the full list of objects is shown. But when set to enforce (the first two messages displayed), only the last object message is shown. In multiple namespaces, the message should be merged.

Should be fixed like this example:
pod [pod1] in namespace test1 found; [pod1] in namespace test2 found as specified, therefore this Object template is compliant

Ref: https://issues.redhat.com/browse/ACM-2604

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -1068,9 +1068,9 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 		nonCompliantObjects := map[string]map[string]interface{}{}
 		compliantObjects := map[string]map[string]interface{}{}
 		enforce := strings.EqualFold(string(plc.Spec.RemediationAction), string(policyv1.Enforce))
-		kind := ""
+		kind := templateObjs[indx].kind
 		objShouldExist := !strings.EqualFold(string(objectT.ComplianceType), string(policyv1.MustNotHave))
-
+		mergeMessageEnforce := false
 		// If the object does not have a namespace specified, use the previously retrieved namespaces
 		// from the NamespaceSelector. If no namespaces are found/specified, use the value from the
 		// object so that the objectTemplate is processed:
@@ -1084,10 +1084,13 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 			relevantNamespaces = []string{templateObjs[indx].namespace}
 		}
 
+		if enforce && len(relevantNamespaces) > 1 {
+			mergeMessageEnforce = true
+		}
+
 		numCompliant := 0
 		numNonCompliant := 0
 		handled := false
-
 		// iterate through all namespaces the configurationpolicy is set on
 		for _, ns := range relevantNamespaces {
 			log.Info(
@@ -1106,20 +1109,33 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 			}
 
 			if objKind != "" {
+				// object template enforced, objKind is empty
 				kind = objKind
 			}
 
+			if mergeMessageEnforce {
+				names = []string{templateObjs[indx].name}
+			}
+
+			// object template enforced, already handled in handleObjects
 			if names == nil {
-				// object template enforced, already handled in handleObjects
 				handled = true
 			} else {
+				// when multiple messages and enforce mode, it doesn't mean remediactionAction is changed
 				enforce = false
+			}
+
+			// violations for enforce configurationpolicies are already handled in handleObjects,
+			// so we only need to generate a violation if the remediationAction is set to inform
+			// Or multiple namespaces and enforce use this
+			if !handled && !enforce {
 				if !compliant {
 					if len(names) == 0 {
 						numNonCompliant++
 					} else {
 						numNonCompliant += len(names)
 					}
+
 					nonCompliantObjects[ns] = map[string]interface{}{
 						"names":  names,
 						"reason": reason,
@@ -1137,8 +1153,7 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 				relatedObjects = updateRelatedObjectsStatus(relatedObjects, object)
 			}
 		}
-		// violations for enforce configurationpolicies are already handled in handleObjects,
-		// so we only need to generate a violation if the remediationAction is set to inform
+
 		if !handled && !enforce {
 			objData := templateIdentifier{
 				index:       indx,
@@ -1147,9 +1162,10 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 				namespaced:  templateObjs[indx].isNamespaced,
 			}
 
-			statusUpdateNeeded := createInformStatus(
+			statusUpdateNeeded := createMergedStatus(
 				objShouldExist, numCompliant, numNonCompliant, compliantObjects, nonCompliantObjects, &plc, objData,
 			)
+
 			if statusUpdateNeeded {
 				parentStatusUpdateNeeded = true
 			}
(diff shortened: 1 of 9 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -1041,9 +1041,9 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 		nonCompliantObjects := map[string]map[string]interface{}{}
 		compliantObjects := map[string]map[string]interface{}{}
 		enforce := strings.EqualFold(string(plc.Spec.RemediationAction), string(policyv1.Enforce))
-		kind := ""
+		kind := templateObjs[indx].kind
```
willkutler, line 1044: we might need to have a check here so the controller doesn't hit a panic if kind is not specified

yiraeChristineKim: okie! good idea

test/e2e/case5_multi_test.go
```diff
@@ -87,4 +87,134 @@ var _ = Describe("Test multiple obj template handling", func() {
 			deleteConfigPolicies(policies)
 		})
 	})
+
+	FDescribe("Check messages when it is multiple namesapces and multiple obj-templates", Ordered, func() {
```
mprahl, line 91:
```suggestion
	Describe("Check messages when it is multiple namesapces and multiple obj-templates", Ordered, func() {
```

test/e2e/case8_status_check_test.go
```diff
@@ -140,11 +140,13 @@ var _ = Describe("Test pod obj template handling", func() {
 				return utils.GetComplianceState(managedPlc)
 			}, defaultTimeoutSeconds, 1).Should(Equal("Compliant"))
 		})
-		It("Cleans up", func() {
+		AfterEach(func() {
```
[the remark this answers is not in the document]
mprahl: Did you mean `AfterAll` here?

controllers/configurationpolicy_controller.go
```diff
@@ -1079,20 +1082,32 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 			}
 
 			if objKind != "" {
+				// object template enforced, objKind is empty
 				kind = objKind
 			}
 
+			if mergeMessageEnforce {
+				names = append([]string{}, templateObjs[indx].name)
```
mprahl, line 1090: Could you please explain this line?

`handleObjects` seems to return a slice with the object name in it already unless the object is unnamed.

controllers/configurationpolicy_controller.go
```diff
@@ -1110,28 +1125,75 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
 		}
 	}
 
 	r.checkRelatedAndUpdate(plc, relatedObjects, oldRelated, parentStatusUpdateNeeded)
 }
 
+// createMergedStatus updates the status field for a configurationpolicy with remediationAction=inform
+// based on how many compliant/noncompliant objects are found when processing the templates in the configurationpolicy
+// Or multiple namespaces and enforce use this
+func createMergedStatus(
```
mprahl, line 1157: Could you please keep this function where it was and just rename it and add the comment? That way the Git diff doesn't make it seem like you wrote a new function.

controllers/configurationpolicy_controller.go
```diff
@@ -1110,28 +1125,75 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
-		if !handled && !enforce {
-			objData := map[string]interface{}{
-				"indx":        indx,
-				"kind":        kind,
-				"desiredName": templateObjs[indx].name,
-				"namespaced":  templateObjs[indx].isNamespaced,
-			}
 
-			statusUpdateNeeded := createInformStatus(
+		objData := map[string]interface{}{
```
mprahl, line 1129: Keeping this under the if statement would be more efficient since you wouldn't have to allocate memory for this map if the if condition is false.

controllers/configurationpolicy_utils.go
```diff
@@ -346,7 +346,16 @@ func isAutogenerated(key string) (result bool) {
 func formatTemplate(unstruct unstructured.Unstructured, key string) (obj interface{}) {
 	if key == "metadata" {
 		//nolint:forcetypeassert
-		metadata := unstruct.Object[key].(map[string]interface{})
+		var metadata map[string]interface{}
+		if unstruct.Object[key] != nil {
+			metadata = unstruct.Object[key].(map[string]interface{})
+		} else {
+			log.Info("metadata was nil, add minimum to avoid panic")
```
mprahl, line 353: Could you please explain the scenario where this introduces a panic?

controllers/configurationpolicy_controller.go
yiraeChristineKim: when it is enforce it sends nil but I need this tempObj.name, I can change handlesingle, handleobj but it is more work and risky to cause another bugs, rather I chose this way clean simple

controllers/configurationpolicy_controller.go
yiraeChristineKim: good!

controllers/configurationpolicy_utils.go
yiraeChristineKim: In case, without `name` key, metadata is nil, 
`complianceType: musthave
      objectDefinition:
        apiVersion: v1
        kind: Pod
        metadata:`
metadata convert to interface{}, so miss matching type, causes err

test/e2e/case5_multi_test.go
yiraeChristineKim: Thanks!

controllers/configurationpolicy_controller.go
```diff
@@ -1110,22 +1125,27 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
+			statusUpdateNeeded := createMergedStatus(
 				objShouldExist, numCompliant, numNonCompliant, compliantObjects, nonCompliantObjects, &plc, objData,
 			)
+
 			if statusUpdateNeeded {
 				parentStatusUpdateNeeded = true
 			}
+
+			if mergeMessageEnforce {
+				r.Recorder.Event(&plc, eventNormal, fmt.Sprintf(plcFmtStr, plc.GetName()),
```
yiraeChristineKim, line 1146: Is this correct sending event? is this message will attach to policy history? curious

controllers/configurationpolicy_controller.go
dhaiducek: If `names == nil` is equivalent to `mergeMessageEnforce == true`, then should setting `names` just be moved down to the `if names == nil {` conditional?
Also, this might look cleaner as `names = []string{templateObjs[indx].name}`. Is using `append()` necessary here?

controllers/configurationpolicy_controller.go
```diff
@@ -1344,7 +1364,8 @@ type templateIdentifier struct {
 
 // createInformStatus updates the status field for a configurationpolicy with remediationAction=inform
 // based on how many compliant/noncompliant objects are found when processing the templates in the configurationpolicy
-func createInformStatus(
+// Or multiple namespaces and enforce use this
+func createMergedStatus(
```
dhaiducek, line 1368: The function name in the description should also match the new function name.

dhaiducek (changes requested):
Thanks! This will be a great update! I've left some comments/questions to consider.

controllers/configurationpolicy_controller.go
yiraeChristineKim: name should be string array

yiraeChristineKim: good good

controllers/configurationpolicy_controller.go
yiraeChristineKim: good good good Thanks

dhaiducek (approved):
/hold if others would like another review

LGTM! The existing code that used `name == nil` as a signal that it had been "handled" is making my head spin, but you solution appears to work alongside it well. Thanks for the update!

PR #117, opened 2023-04-06
Tweak error handling

dhaiducek:
- Adds a target for foreground logging during E2E
- Removes a pointless return of an input pointer variable
- Adds error handling when the Get for the deletion object fails

### DIFF
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -21,6 +21,7 @@ jobs:
     env:
       REGISTRY: localhost:5000
     strategy:
+      fail-fast: false
       matrix:
         # Run tests on minimum and newest supported OCP Kubernetes
         # The "minimum" tag is set in the Makefile
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -517,13 +517,21 @@ func (r *ConfigurationPolicyReconciler) cleanUpChildObjects(plc policyv1.Configu
 		// determine whether object should be deleted
 		needsDelete := false
 
-		existing, _ := getObject(
+		existing, err := getObject(
 			namespaced,
 			object.Object.Metadata.Namespace,
 			object.Object.Metadata.Name,
 			mapping.Resource,
 			r.TargetK8sDynamicClient,
 		)
+		if err != nil {
+			log.Error(err, "Failed to get child object")
+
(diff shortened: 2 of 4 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -2714,24 +2722,42 @@ func (r *ConfigurationPolicyReconciler) addForUpdate(policy *policyv1.Configurat
...
 		"Updating configurationPolicy status", "status", policy.Status.ComplianceState, "policy", policy.GetName(),
 	)
 
-	err := r.Status().Update(context.TODO(), policy)
-	if err != nil {
-		return policy, err
+	updatedStatus := policy.Status
+
+	maxRetries := 3
+	for i := 1; i <= maxRetries; i++ {
```
mprahl, line 2743: This comes from the cache, so we don't need any retry logic. If it fails to retrieve from the cache, we have bigger problems,

.github/workflows/kind.yml
```diff
@@ -21,6 +21,7 @@ jobs:
     env:
       REGISTRY: localhost:5000
     strategy:
+      fail-fast: false
```
JustinKuli, line 24: 👍

controllers/configurationpolicy_controller.go
```diff
@@ -517,13 +517,21 @@ func (r *ConfigurationPolicyReconciler) cleanUpChildObjects(plc policyv1.Configu
...
-		existing, _ := getObject(
+		existing, err := getObject(
 			namespaced,
 			object.Object.Metadata.Namespace,
 			object.Object.Metadata.Name,
 			mapping.Resource,
 			r.TargetK8sDynamicClient,
 		)
+		if err != nil {
+			log.Error(fmt.Errorf("failed to get object"), "Error")
```
JustinKuli, line 528: Why not
```suggestion
			log.Error(err, "Failed to get object")
```

dhaiducek: In my first iteration, I wasn't using the error object because it was empty. I'll update--thanks!

controllers/configurationpolicy_controller.go
dhaiducek: Looks like your PR worked, so I'll take the refetch out. I think the other changes are still relevant.

gparvin (approved):
This all looks good to me

PR #118, opened 2023-04-06
Use patches for adding and removing finalizers

mprahl:
Using update can have unintended consequences such as setting empty values of fields that were nil on the server. This happens for the evaluationInterval field where if it wasn't set, the update adds an empty map, which triggers the spec-sync to remove the nil map. This unnecessary spec-sync update causes the status update to fail later on in the ConfigurationPolicy processing.

Additionally, when a finalizer can't be added or removed, the evaluation stops and would be retried on the next run. This is a safer approach.

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -704,11 +704,13 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 
 		if cleanupNow {
 			if objHasFinalizer(&plc, pruneObjectFinalizer) {
-				plc.SetFinalizers(removeObjFinalizer(&plc, pruneObjectFinalizer))
+				patch := removeObjFinalizerPatch(&plc, pruneObjectFinalizer)
 
-				err := r.Update(context.TODO(), &plc)
+				err = r.Patch(context.TODO(), &plc, client.RawPatch(types.JSONPatchType, patch))
 				if err != nil {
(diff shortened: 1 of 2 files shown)


yiraeChristineKim (approved):

controllers/configurationpolicy_controller.go
```diff
@@ -717,11 +719,13 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 
 		// set finalizer if it hasn't been set
 		if !objHasFinalizer(&plc, pruneObjectFinalizer) {
-			plc.SetFinalizers(addObjFinalizer(&plc, pruneObjectFinalizer))
+			mergePatch := []byte(`{"metadata": {"finalizers": ["` + pruneObjectFinalizer + `"]}}`)
 
-			err := r.Update(context.TODO(), &plc)
+			err := r.Patch(context.TODO(), &plc, client.RawPatch(types.MergePatchType, mergePatch))
```
JustinKuli, line 724: @mprahl could this accidentally remove any other finalizer on the resource? I was expecting a JSON patch like `[{"op":"add","path":"/metadata/finalizers/-","value":"..."}]`

mprahl: Good catch. I thought the merge patch would act like strategic merge but I got the two confused. Here is a PR with the fix:
https://github.com/open-cluster-management-io/config-policy-controller/pull/119

PR #119, opened 2023-04-06
Account for existing finalizers when adding the prune object finalizer

mprahl:
A merge patch will overwrite the existing value.


dhaiducek (approved):

mprahl:
@dhaiducek I tweaked this a little bit so that I think it's safer approach.

dhaiducek (approved):

PR #127, opened 2023-04-21
QPS config

JustinKuli:
The recommended settings are like 2:30:45 for concurrency:qps:burst. When QPS or Burst are not explicitly set, they will scale based on concurrency.

Refs:
 - https://issues.redhat.com/browse/ACM-4297

### DIFF
--- a/main.go
+++ b/main.go
@@ -67,61 +67,54 @@ func init() {
 	utilruntime.Must(extensionsv1beta1.AddToScheme(scheme))
 }
 
+type ctrlOpts struct {
+	clusterName           string
+	hubConfigPath         string
+	targetKubeConfig      string
+	metricsAddr           string
+	probeAddr             string
+	frequency             uint
+	decryptionConcurrency uint8
+	evaluationConcurrency uint8
+	enableLease           bool
+	enableLeaderElection  bool
+	legacyLeaderElection  bool
+	enableMetrics         bool
+}
+
 func main() {
 	klog.InitFlags(nil)
+
+	subcommand := ""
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -447,3 +428,109 @@ func handleTriggerUninstall() {
...
+		&opts.enableMetrics,
+		"enable-metrics",
+		true,
+		"Disable custom metrics collection",
+	)
+
+	flags.Float32Var(
+		&opts.clientQPS,
+		"client-max-qps",
+		30, // 15 * concurrency is recommended
```
mprahl, line 524: I was thinking that if the QPS is not explicitly set, then we can do 15 * evaluationConcurrency rather than rely on the user to know they should also adjust QPS.

main.go
```diff
@@ -176,6 +153,9 @@ func main() {
 		os.Exit(1)
 	}
 
+	cfg.Burst = int(opts.clientBurst)
+	cfg.QPS = opts.clientQPS
```
yiraeChristineKim, line 157: Where cfg.QPS  used?

mprahl: `cfg` gets assigned to `targetK8sConfig` which then gets used to create Kubernetes API client instances for `ConfigurationPolicyReconciler`.

main.go
JustinKuli: I was planning on doing that kind of configuration in the addon-controller, but I can do it here.

yiraeChristineKim (dismissed):

mprahl (approved):

PR #132, opened 2023-05-16
Update API discovery cache even when incomplete

JustinKuli:
Previously, if there were any APIs having issues on the cluster, no new API information could be used by the controller. Now, partial results can still be updated. Additionally, API discovery errors during startup will not prevent the controller from functioning at all.

Refs:
 - https://issues.redhat.com/browse/ACM-5491

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -178,13 +178,16 @@ func (r *ConfigurationPolicyReconciler) PeriodicallyExecConfigPolicies(
 		}
 
 		start := time.Now()
-		policiesList := policyv1.ConfigurationPolicyList{}
 
 		var skipLoop bool
-		var discoveryErr error
 
 		if len(r.apiResourceList) == 0 || len(r.apiGroups) == 0 {
-			discoveryErr = r.refreshDiscoveryInfo()
+			discoveryErr := r.refreshDiscoveryInfo()
+
+			// If there was an error and no API information was received, then skip the loop.
+			if discoveryErr != nil && (len(r.apiResourceList) == 0 || len(r.apiGroups) == 0) {
+				skipLoop = true
+			}
 		}
 
 		// If it's been more than 10 minutes since the last refresh, then refresh the discovery info, but ignore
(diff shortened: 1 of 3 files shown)


JustinKuli:
I was considering reporting the controller as unhealthy if the API discovery was having an error, but I realized that restarting the container (which is what kubernetes will do when it's unhealthy) won't be likely to fix anything. It would be easy to add that in though, see https://github.com/JustinKuli/config-policy-controller-1/tree/5491-incomplete-server-apis.

Additionally, @mprahl suggested that we take a more similar approach to the API discovery as in the dependency-watches library: https://github.com/stolostron/kubernetes-dependency-watches/pull/8. I think that is a good longer-term solution, but this controller shares the API information with the templates library, and I would rather have a smaller fix in sooner, rather than a bigger one in multiple places that could be tricky to coordinate.

controllers/configurationpolicy_controller.go
```diff
@@ -96,6 +96,7 @@ type discoveryInfo struct {
 	apiResourceList        []*metav1.APIResourceList
 	apiGroups              []*restmapper.APIGroupResources
 	discoveryLastRefreshed time.Time
+	refreshError           error
```
mprahl, line 99: What's the purpose for this being a struct member as opposed to a local variable? I don't see it used outside of the `refreshDiscoveryInfo` method.

test/e2e/case18_discovery_refresh_test.go
```diff
@@ -54,12 +56,33 @@ var _ = Describe("Test discovery info refresh", func() {
...
+		utils.Kubectl("apply", "-f", badAPIServiceYaml)
+
+		By("Checking that the API causes an error during API discovery")
+		Eventually(
+			func() interface{} {
+				cmd := exec.Command("kubectl", "api-resources")
+
+				err := cmd.Start()
+				if err != nil {
+					return err
```
mprahl, line 69: This error is not indicative of the error you are looking for so `Eventually` shouldn't stop from this error.

JustinKuli: Nice catch.

controllers/configurationpolicy_controller.go
```diff
@@ -288,35 +289,47 @@ func (r *ConfigurationPolicyReconciler) handlePolicyWorker(
...
 	}
 
-	r.apiGroups = apiGroups
 	// Reset the OpenAPI cache in case the CRDs were updated since the last fetch.
 	r.openAPIParser = openapi.NewOpenAPIParser(dd)
-	r.discoveryLastRefreshed = time.Now().UTC()
+	r.discoveryInfo.discoveryLastRefreshed = time.Now().UTC()
 
-	return nil
+	return r.refreshError // can be nil
```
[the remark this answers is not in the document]
JustinKuli: The adjustment for this allowed for some other refactorings around "the loop" - if it's not clear how it works now, I can split it into another commit to make it more clear.

controllers/configurationpolicy_controller.go
```diff
@@ -202,49 +205,45 @@ func (r *ConfigurationPolicyReconciler) PeriodicallyExecConfigPolicies(
 			skipLoop = true
 		}
 
-		if !skipLoop && (discoveryErr == nil || cleanupImmediately) {
+		if !skipLoop {
```
mprahl, line 208: If the API server is down of the hosted cluster in hosted mode, we still want to clean up the finalizers on the management cluster.
```suggestion
		if !skipLoop || cleanupImmediately {
```

JustinKuli: Right

mprahl (approved):

PR #135, opened 2023-05-25
is it required to validate object

clyang82:
This is not a final PR need to be reviewed.

I just post this PR to discuss:
1. is it really need to validate the object since the unstructured object is just retrieved from the runtime environment. refer to https://github.com/open-cluster-management-io/config-policy-controller/blob/049994499251e22ff843e1d14654daf281ddc900/controllers/configurationpolicy_controller.go#L1504-L1506
2. https://github.com/open-cluster-management-io/config-policy-controller/blob/049994499251e22ff843e1d14654daf281ddc900/controllers/configurationpolicy_controller.go#L2544 need to consume around 10M memory based on my tests. And it cannot be released. Is it possible to reduce the cost?
Here is my tests code
```
	var m1, m2 goruntime.MemStats

	goruntime.ReadMemStats(&m1)
	fmt.Println("Alloc in validateObject: m1", m1.Alloc/(1<<10))

	// Parse() handles caching of the OpenAPI data.
	r.lock.RLock()
	openAPIResources, err := r.openAPIParser.Parse()
	r.lock.RUnlock()
	goruntime.ReadMemStats(&m2)
	fmt.Println("Alloc in validateObject: m2", m2.Alloc/(1<<10))
	fmt.Println("memory usage Alloc:", (m2.Alloc-m1.Alloc)/(1<<10))

```

FYI @gparvin @mprahl

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -18,6 +18,7 @@ import (
 	gocmp "github.com/google/go-cmp/cmp"
 	"github.com/prometheus/client_golang/prometheus"
 	templates "github.com/stolostron/go-template-utils/v3/pkg/templates"
+	"golang.org/x/mod/semver"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	extensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -95,6 +96,7 @@ type cachedEncryptionKey struct {
 type discoveryInfo struct {
 	apiResourceList        []*metav1.APIResourceList
 	apiGroups              []*restmapper.APIGroupResources
+	serverVersion          string
 	discoveryLastRefreshed time.Time
 }
 
@@ -296,6 +298,13 @@ func (r *ConfigurationPolicyReconciler) refreshDiscoveryInfo() error {
 	defer func() { r.lock.Unlock() }()
 
 	dd := r.TargetK8sClient.Discovery()
+
+	serverVersion, err := dd.ServerVersion()
+	if err != nil {
+		log.Error(err, "Could not get the server version")
+	}
+
+	r.serverVersion = serverVersion.String()
 
 	_, apiResourceList, resourceErr := dd.ServerGroupsAndResources()
 	if resourceErr != nil {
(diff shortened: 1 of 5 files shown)


mprahl:
@clyang82 interesting find! So the reason it's needed is that during object creation, it validates to ensure all required fields and no unknown fields are set in the policy. On updates, it takes the object returned from the Kubernetes API and merges in any updates from the policy. The validation ensures that no unknown fields were set.

The reason it is important that no unknown fields are set is that Kubernetes silently ignores unknown fields and then when ConfigurationPolicy is evaluated again, it'll think that it's non-compliant because the unknown field is missing and tries to update it again.

I think we could consider setting `FieldValidation` to `Strict` on the create and update requests so that the requests fail when unknown fields are set. This would only handle the case for `enforce` though since in `inform`, it wouldn't know that there is an unknown field without client side validation.

What do you think @JustinKuli?

clyang82:
Thanks @mprahl for your comments. I gave a try to set `FieldValidation` to `Strict` on the create and update. Please take a look at if you are OK with the changes.

JustinKuli:
The changes sound right to me. I'm happy to see the kind tests cover both the newer k8s with the validation, and the older version where it wasn't possible server-side.

@clyang82 what did you mean in the original description "And it cannot be released." ?

controllers/configurationpolicy_controller.go
```diff
@@ -297,6 +299,13 @@ func (r *ConfigurationPolicyReconciler) refreshDiscoveryInfo() error {
 
 	dd := r.TargetK8sClient.Discovery()
 
+	serverVersion, err := dd.ServerVersion()
+	if err != nil {
+		log.Error(err, "Could not get the server version")
+	}
+
+	r.serverVersion = serverVersion.String()
```
mprahl, line 307: Will this work if `err != nil`?

JustinKuli: Assuming that in that case it is just set to an empty string, then the semver library would treat it as invalid and do this:
> An invalid semantic version string is considered less than a valid one. All invalid semantic version strings compare equal to each other.

So I think it would fallback to the "old" validation gracefully.

clyang82:
> The changes sound right to me. I'm happy to see the kind tests cover both the newer k8s with the validation, and the older version where it wasn't possible server-side.
> 
> @clyang82 what did you mean in the original description "And it cannot be released." ?

I mean the memory is allocated by `r.openAPIParser.Parse()` won't be released.

controllers/configurationpolicy_controller.go
clyang82: Yes. @JustinKuli is correct. in that case it would fallback to the original validation.

PR #136, opened 2023-05-25
Log policy NonCompliance

gparvin:
Consistently log the policy noncompliance and compliance as updates are made.  The goal is to allow tools that scrape logs to be able to obtain the violation message and the details on when compliance changes happen.

Refs:
 - https://issues.redhat.com/browse/ACM-5568

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -2826,11 +2826,14 @@ func (r *ConfigurationPolicyReconciler) updatePolicyStatus(
 			eventType = eventWarning
 		}
 
+		eventMessage := fmt.Sprintf("%s%s", policy.Status.ComplianceState, msg)
+		log.Info("Policy status message", "policy", policy.GetName(), "status", eventMessage)
+
 		r.Recorder.Event(
 			policy,
 			eventType,
 			"Policy updated",
-			fmt.Sprintf("Policy status is %s%s", policy.Status.ComplianceState, msg),
+			fmt.Sprintf("Policy status is %s", eventMessage),
 		)
 	}
 


gparvin:
Sample lop entry:
```
2023-05-24T13:37:53.496-0400	info	configuration-policy-controller	controllers/configurationpolicy_controller.go:2915	Policy status message	{"name": "case23-invalid-field", "namespace": "managed", "status": "NonCompliant; violation - Error validating the object case23, the error is `ValidationError(ConfigMap): unknown field \"invalid\" in io.k8s.api.core.v1.ConfigMap`"}
```

controllers/configurationpolicy_controller.go
```diff
@@ -2911,7 +2908,7 @@ func convertPolicyStatusToString(plc *policyv1.ConfigurationPolicy) (results str
 			}
 		}
 	}
-
+	policyLog.Info("Policy status message", "status", result)
```
JustinKuli, line 2911: Having the log inside this function might result in it being duplicated sometimes, but I think that is a better problem than missing it sometimes. I wonder if it should be behind `V(1)` or a special flag, to keep things from getting too noisy?

gparvin: The goal is to have this logged by default so servicing systems that only have the GRC governance metrics and logs can determine the NonCompliance messages.  Since the NonCompliance message isn't in a metric we need it to always be in log.  I'll have to see if the V(1) logs show up by default.  I'll do a duplication detection too to see if that happens -- it looks like I've cleaned up the logs I captured last week so will recollect.

dhaiducek: I agree with @JustinKuli. I think these could be very noisy and make the logs grow quickly--I'd prefer we ask that servicing systems bump the log level if they want the information present in the logs. It's not too difficult since we have a custom annotation that can be applied to the ManagedClusterAddon. Does SD have a requirement that the status message be present at default logging levels?

gparvin:
/hold
I changed to the V(1) logging -- but I'm not convinced there's a valid concern around log growth due to this change yet.

gparvin:
I rebuilt the changes on the latest code base.  To get the message to display I had to use: `--log-level debug`.  I couldn't tell that `--v 1` does anything.  Is that what we expect?

controllers/configurationpolicy_controller.go
```diff
@@ -2826,11 +2826,14 @@ func (r *ConfigurationPolicyReconciler) updatePolicyStatus(
 			eventType = eventWarning
 		}
 
+		eventMessage := fmt.Sprintf("%s%s", policy.Status.ComplianceState, msg)
+		log.Info("Policy status message", "status", eventMessage)
```
mprahl, line 2830: The logger here doesn't automatically include the policy name. Could you please include that as a key/value pair?

mprahl (approved):

controllers/configurationpolicy_controller.go
gparvin: Good point.  Now it has the policy name.  Thanks

PR #139, opened 2023-05-31
Bug: Objects are pruned on templating errors

yiraeChristineKim:
Description:
Deploying a enforced ConfigurationPolicy with a valid template and then updating the template to be invalid caused the created object to be pruned. (I did it with hub templates, but I'd suspect managed cluster templates would have a similar behavior.)

 

expected:  Not perform any deletion while in this error state.

Ref: https://issues.redhat.com/browse/ACM-5301
Signed-off-by: Yi Rae Kim <yikim@redhat.com>

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -698,7 +698,7 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 				fmt.Sprintf(plcFmtStr, plc.GetName()), convertPolicyStatusToString(&plc))
 		}
 
-		r.checkRelatedAndUpdate(plc, relatedObjects, oldRelated, statusChanged)
+		r.checkRelatedAndUpdate(plc, relatedObjects, oldRelated, statusChanged, true)
 
 		parent := ""
 		if len(plc.OwnerReferences) > 0 {
@@ -793,7 +793,7 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 				}
 
 				// don't change related objects while deletion is in progress
-				r.checkRelatedAndUpdate(plc, oldRelated, oldRelated, parentStatusUpdateNeeded)
+				r.checkRelatedAndUpdate(plc, oldRelated, oldRelated, parentStatusUpdateNeeded, true)
 			}
 
 			return
@@ -810,6 +810,8 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 		}
 	}
 
+	// When it is hub or managed template parse error, deleteDetachedObjs should be false
+	// Then it doesn't remove resources
 	addTemplateErrorViolation := func(reason, msg string) {
 		log.Info("Setting the policy to noncompliant due to a templating error", "error", msg)
 
@@ -829,7 +831,8 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 			)
 		}
 
-		r.checkRelatedAndUpdate(plc, relatedObjects, oldRelated, parentStatusUpdateNeeded)
+		// deleteDetachedObjs should be false
+		r.checkRelatedAndUpdate(plc, relatedObjects, oldRelated, parentStatusUpdateNeeded, false)
 	}
 
 	// initialize apiresources for template processing before starting objectTemplate processing
@@ -1054,7 +1057,7 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 
 	if err != nil {
 		if parentStatusUpdateNeeded {
-			r.checkRelatedAndUpdate(plc, relatedObjects, oldRelated, parentStatusUpdateNeeded)
+			r.checkRelatedAndUpdate(plc, relatedObjects, oldRelated, parentStatusUpdateNeeded, false)
 		}
 
 		return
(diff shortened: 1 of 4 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -967,7 +969,7 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 					encryptionConfig, usedKeyCache, err = r.getEncryptionConfig(plc, true)
 
 					if err != nil {
-						addTemplateErrorViolation("", err.Error())
+						addTemplateErrorViolation("", err.Error(), true)
```
yiraeChristineKim, line 972: There are more `addTemplateErrorViolation` I am actually not sure which one should I add "skip removing resources"... Would anyone give me advice?

controllers/configurationpolicy_controller.go
```diff
@@ -1193,8 +1195,14 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
 	plc policyv1.ConfigurationPolicy, related, oldRelated []policyv1.RelatedObject, sendEvent bool,
+	isTemplateParseErr ...bool,
 ) {
-	r.sortRelatedObjectsAndUpdate(&plc, related, oldRelated, r.EnableMetrics)
+	tempParseErr := false
+	if len(isTemplateParseErr) > 0 {
+		tempParseErr = isTemplateParseErr[0]
+	}
+
+	r.sortRelatedObjectsAndUpdate(&plc, related, oldRelated, r.EnableMetrics, tempParseErr)
```
yiraeChristineKim, line 1205: why I didn't use "variadic parameter", sortRelatedObjectsAndUpdate is only used here

controllers/configurationpolicy_controller.go
```diff
@@ -1193,8 +1195,14 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 // checkRelatedAndUpdate checks the related objects field and triggers an update on the ConfigurationPolicy
 func (r *ConfigurationPolicyReconciler) checkRelatedAndUpdate(
 	plc policyv1.ConfigurationPolicy, related, oldRelated []policyv1.RelatedObject, sendEvent bool,
+	isTemplateParseErr ...bool,
```
JustinKuli, line 1198: I think I would prefer this to not be variadic, because it makes it easier to review that we've considered each time that this function is called. I would also prefer it to be called something like `deleteDetachedObjs` to describe what it does, rather than where it's coming from, since there could be other cases where we don't want to do this cleanup, but the reason has nothing to do with template errors.

JustinKuli: Looking at other times `checkRelatedAndUpdate` is called, I think there are more error cases where we don't want to delete "detached" objects.

JustinKuli: Final note (sorry), I think the function signature is long enough now that each parameter should be on a separate line, like https://github.com/open-cluster-management-io/config-policy-controller/blob/main/controllers/configurationpolicy_controller.go#L1394. But that's just a style thing.

controllers/configurationpolicy_controller.go
yiraeChristineKim: @JustinKuli  Did I miss anything?!

controllers/configurationpolicy_controller.go
```diff
@@ -1074,7 +1077,7 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 				convertPolicyStatusToString(&plc))
 		}
 
-		r.checkRelatedAndUpdate(plc, relatedObjects, oldRelated, statusUpdateNeeded)
+		r.checkRelatedAndUpdate(plc, relatedObjects, oldRelated, statusUpdateNeeded, false)
```
JustinKuli, line 1080: I think this one should be `true`, this case could come up if they have a template with a range, but it changes to be empty. In this case, I think we could assume that it's working correctly, and they might want to remove the objects that used to be there.

controllers/configurationpolicy_controller.go
```diff
@@ -810,6 +810,8 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 		}
 	}
 
+	// When it is hub or managed template parse error, deleteDetachedObjs should be true
```
JustinKuli, line 813:
```suggestion
	// When it is hub or managed template parse error, deleteDetachedObjs should be false
```

controllers/configurationpolicy_controller.go
```diff
@@ -1187,14 +1190,17 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 		}
 	}
 
-	r.checkRelatedAndUpdate(plc, relatedObjects, oldRelated, parentStatusUpdateNeeded)
+	r.checkRelatedAndUpdate(plc, relatedObjects, oldRelated, parentStatusUpdateNeeded, false)
```
JustinKuli, line 1193: I think this one should be `true`, I think this is the "happy path" and that `true` is be needed here to ensure that when the related objects changed, then old ones are correctly removed... this might be the cause of the currently failing test in the PR.

JustinKuli (changes requested):
Thanks for the update, I think it's looking right overall, just a few last details to iron out! Once we get the tests passing I think this will be good.

controllers/configurationpolicy_controller.go
yiraeChristineKim: good catch Thanks!

JustinKuli (approved):

PR #140, opened 2023-06-07
Add error output for `api-resources` command

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-5849


dhaiducek:
Not sure what happened here. GitHub didn't pick up my new pushes and then couldn't reconcile reopening it when it was closed. New PR:
- https://github.com/open-cluster-management-io/config-policy-controller/pull/141

PR #142, opened 2023-06-14
[Critical Bug]unexpectedly deleted when pruneObjectBehavior is None

yiraeChristineKim:
Description of problem:

When a ConfigurationPolicy stops managing an object, it is always deleted regardless of pruneObjectBehavior being set to `None`.

Below are a few examples of when a ConfigurationPolicy would stop managing an object:

The ConfigurationPolicy is updated to manage a different object.
The ConfigurationPolicy's namespaceSelector is changed.
The ConfigurationPolicy's templating resolves differently based on changing conditions to no longer manage an object.
Version-Release number of selected component (if applicable):
How reproducible:

Very reproducable

Steps to Reproduce:
Create a ConfigurationPolicy with pruneObjectBehavior set to "None" (default) that creates/updates a ConfigMap called "test"
Update the ConfigurationPolicy to create/update a ConfigMap called "not test"
Actual results:

After following the reproducing steps, the ConfigMap called "test" is deleted.

Expected results:

After following the reproducing steps, the ConfigMap called "test" should not be deleted.

Additional info:

This will cause unexpected data loss to customers.

Ref: https://issues.redhat.com/browse/ACM-5939


controllers/configurationpolicy_controller.go
```diff
@@ -706,7 +728,7 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 				fmt.Sprintf(plcFmtStr, plc.GetName()), convertPolicyStatusToString(&plc))
 		}
 
-		r.checkRelatedAndUpdate(plc, relatedObjects, oldRelated, statusChanged, true)
```
yiraeChristineKim, line 709: Since we check only namespace, name, kind, apiVersion.. so no needed anymore

test/e2e/__debug_bin
JustinKuli, line 1: We probably don't want this

controllers/configurationpolicy_utils.go
```diff
@@ -557,9 +557,9 @@ func removeObjFinalizerPatch(obj metav1.Object, finalizer string) []byte {
 	return nil
 }
 
-func containRelated(arr []policyv1.RelatedObject, input policyv1.RelatedObject) bool {
-	// should compare only object
-	for _, r := range arr {
+func containRelated(related []policyv1.RelatedObject, input policyv1.RelatedObject) bool {
+	// should compare name, APIVersion, Kind  and namepcace
```
JustinKuli, line 561:
```suggestion
	// should compare name, APIVersion, Kind  and namespace
```

controllers/configurationpolicy_controller.go
```diff
@@ -506,9 +506,29 @@ func (r *ConfigurationPolicyReconciler) getObjectTemplateDetails(
...
+			if !exist {
+				objShouldRemoved = append(objShouldRemoved, oldR)
+			}
+		}
+
+		plc.Status.RelatedObjects = objShouldRemoved
+	}
+
 	for _, object := range plc.Status.RelatedObjects {
 		// set up client for object deletion
```
JustinKuli, line 533: I'm thinking out loud here in case this is cleaner (it is to me at least)... what do you think of this?

```
func (r *ConfigurationPolicyReconciler) cleanUpChildObjects(plc policyv1.ConfigurationPolicy,
	newRelated []policyv1.RelatedObject,
) []string {
	deletionFailures := []string{}
	objsToDelete := []policyv1.RelatedObject{}
	
    log.Info("preparing to delete detached objects...")

	if len(newRelated) == 0 { // delete all
	    objsToDelete = plc.Status.RelatedObjects
	} else { // When spec is updated and new related objects are created
		oldRelated := plc.Status.RelatedObjects

		for _, oldR := range oldRelated {
			if !containRelated(newRelated, oldR) {
				objsToDelete = append(objsToDelete, oldR)
			}
		}
	}

	for _, object := range objsToDelete {
		// set up client for object deletion
```

JustinKuli: Actually, the two cases can just work the same, because if the `newRelated` is empty, then `containRelated` will just always be false.

```
func (r *ConfigurationPolicyReconciler) cleanUpChildObjects(plc policyv1.ConfigurationPolicy,
	newRelated []policyv1.RelatedObject,
) []string {
	deletionFailures := []string{}
	objsToDelete := []policyv1.RelatedObject{}
	
    log.Info("preparing to delete detached objects...")
    
	oldRelated := plc.Status.RelatedObjects
	for _, oldR := range oldRelated {
		if !containRelated(newRelated, oldR) {
			objsToDelete = append(objsToDelete, oldR)
		}
	}
		
	for _, object := range objsToDelete {
		// set up client for object deletion
```

JustinKuli (commented):
Other than the extra file, I think this looks right! Feel free to ignore my suggestion, I might have gotten a bit carried away...

controllers/configurationpolicy_utils.go
yiraeChristineKim: good

controllers/configurationpolicy_controller.go
yiraeChristineKim: what is deletionFailures

yiraeChristineKim: also logic is same but change is variable name right?

yiraeChristineKim: we shouldn't change `for _, object := range objsToDelete {`

yiraeChristineKim: when newRelated  = nil we have to fall in loop, unnecessary isn't it?

controllers/configurationpolicy_controller.go
```diff
@@ -506,10 +506,30 @@ func (r *ConfigurationPolicyReconciler) getObjectTemplateDetails(
...
+
+		for _, oldR := range objsToDelete {
+			exist := containRelated(newRelated, oldR)
+
+			if !exist {
+				objShouldRemoved = append(objShouldRemoved, oldR)
+			}
+		}
+
+		objsToDelete = objShouldRemoved
```
mprahl, line 529: Optional simplification
```suggestion
		objsToDelete := []policyv1.RelatedObject{}

		for _, oldR := range objsToDelete {
			if !containRelated(newRelated, oldR) {
				objsToDelete = append(objsToDelete, oldR)
			}
		}
```

controllers/configurationpolicy_controller.go
```diff
@@ -767,7 +789,8 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
 		if plc.ObjectMeta.DeletionTimestamp != nil {
 			log.V(1).Info("Config policy has been deleted, handling child objects")
 
-			failures := r.cleanUpChildObjects(plc)
+			// At here, relatedObjects is always []
+			failures := r.cleanUpChildObjects(plc, relatedObjects)
```
mprahl, line 793: Maybe this is clearer?
```suggestion
			failures := r.cleanUpChildObjects(plc, nil)
```

test/resources/case13_templatization/case13_fromsecret.yaml
```diff
@@ -3,6 +3,7 @@ kind: ConfigurationPolicy
 metadata:
   name: tmplt-policy-secret-duplicate
 spec:
+  pruneObjectBehavior: DeleteAll
```
mprahl, line 6: Is this change needed?

controllers/configurationpolicy_controller.go
```diff
@@ -571,6 +591,8 @@ func (r *ConfigurationPolicyReconciler) cleanUpChildObjects(plc policyv1.Configu
 					object.Properties.UID == uid {
 					needsDelete = true
 				}
+			} else { // None or not
+				needsDelete = false
```
mprahl, line 595: This is happening too late in the method since the code above is performing API queries only for the return values to never get used.

My suggestion is to add something to this effect in the start of the method:
```go
if !strings.EqualFold(string(plc.Spec.RemediationAction), "enforce") {
	return deletionFailures
}

if !(string(plc.Spec.PruneObjectBehavior) == "DeleteAll" || string(plc.Spec.PruneObjectBehavior) == "DeleteIfCreated") {
	return deletionFailures
}
```

Afterwards, you can also remove the if statement of `if strings.EqualFold(string(plc.Spec.RemediationAction), "enforce") {` on line 581 of the diff.

mprahl (changes requested):
This looks good but just a performance concern and a few optional improvements.

test/resources/case13_templatization/case13_fromsecret.yaml
yiraeChristineKim: yes

controllers/configurationpolicy_controller.go
yiraeChristineKim: good

controllers/configurationpolicy_controller.go
```diff
@@ -506,10 +506,39 @@ func (r *ConfigurationPolicyReconciler) getObjectTemplateDetails(
...
+	if !(string(plc.Spec.PruneObjectBehavior) == "DeleteAll" ||
+		string(plc.Spec.PruneObjectBehavior) == "DeleteIfCreated") {
+		return deletionFailures
+	}
+
+	objsToDelete := plc.Status.RelatedObjects
+
+	// When spec is updated and new related objects are created
+	if len(newRelated) != 0 {
+		log.Info("preparing to delete detached objects...")
```
mprahl, line 528: Optional:
```suggestion
		log.Info("Preparing to delete detached objects...")
```

mprahl: Shouldn't this log message only be shown if `objsToDelete` has entries?

test/resources/case13_templatization/case13_fromsecret.yaml
JustinKuli: I had requested this :)

If this test doesn't have `pruneObjectBehavior` set, then the template error definitely shouldn't cause anything to be removed (because now, things are only removed when that is set). The test is meant to answer the question "Will objects be removed when there is a templating error, and I generally want objects to be pruned?" with "No, objects will not be removed".

mprahl (commented):
A few minor comments

JustinKuli (approved):

PR #144, opened 2023-06-20
Bug: ACM-5052, Policy gets shortly into non-compliant state

JeffeyL:
**Description of problem:** When `stringData` is set on a Secret within a policy, Kubernetes automatically converts it to a base64 encoded `data` field. The next time the policy is evaluated, the controller expects to see a `stringData` field - however, `stringData` is no longer set (due to the conversion) so the controller sees a mismatch.

**Solution:** When the controller handles a single key, a check has been added for the exact case where `stringData` is set within a Secret. In this case, the `data` field will be base64 _decoded_ and then used for comparison.

Ref: https://issues.redhat.com/browse/ACM-5052


controllers/configurationpolicy_controller.go
```diff
@@ -2556,6 +2558,19 @@ func handleSingleKey(
 			mergedValue.(map[string]interface{}), existingValue.(map[string]interface{}))
 	}
 
+	if key == "stringData" && existingObj.UnstructuredContent()["kind"] == "Secret" {
+		// override automatic conversion from stringData to data prior to evaluation
+		existingValue = existingObj.UnstructuredContent()["data"]
+		for k, val := range existingValue.(map[string]interface{}) {
+			decodedVal, err := base64.StdEncoding.DecodeString(val.(string))
+			if err != nil {
+				message := fmt.Sprintf("Error decoding secret %s: %s", val.(string), err)
```
JustinKuli, line 2567: Will this message be written into the policy status, or into the log? If so, I think we should be careful not to leak any of the data from the secret. In this case, it's data that couldn't be decoded (somehow), but it could still be sensitive. I think I'd rather this message just give the secret's name and the key where the error occurred (the error itself might also leak sensitive information, although I'm not sure what the error string would really look like).

Again, just me being paranoid 😆

JustinKuli (changes requested):
Overall this looks really good! I think it would fix the problem (in a surprisingly clean way!), but I think we need a better test to be sure.

test/e2e/case32_secret_stringdata_test.go
```diff
@@ -0,0 +1,63 @@
...
+	It("Verifies the config policy stays compliant "+case32ConfigPolicyName+" in "+testNamespace, func() {
+		By("Making sure " + case32ConfigPolicyName + " is consistently Compliant for " + strconv.Itoa(defaultTimeoutSeconds) + " seconds")
+		Consistently(func() interface{} {
+			cfgplc := utils.GetWithTimeout(
+				clientManagedDynamic, gvrConfigPolicy, case32ConfigPolicyName, testNamespace, true, defaultTimeoutSeconds,
+			)
+
+			return utils.GetComplianceState(cfgplc)
+		}, defaultTimeoutSeconds, 1).Should(Equal("Compliant"))
+	})
```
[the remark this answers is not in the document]
JeffeyL: That makes sense - after looking at case31 I can see why this is a more reliable approach. I'll have it fixed along with the other changes you mentioned. Thanks for the review!

controllers/configurationpolicy_controller.go
```diff
@@ -2556,6 +2558,29 @@ func handleSingleKey(
...
+		if !ok {
+			message := "Error casting encoded data"
+
+			return message, false, mergedValue, false
+		}
+
+		for k, val := range value {
+			decodedVal, err := base64.StdEncoding.DecodeString(val.(string))
+			if err != nil {
+				message := fmt.Sprintf("Error decoding secret: %s", k)
```
JustinKuli, line 2575: This message sounds like it will include the secret's name, but actually that's the key in the secret that couldn't be decoded... is that right?

test/resources/case32_secret_stringdata/case32_create_secret.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: policy.open-cluster-management.io/v1
+kind: ConfigurationPolicy
+metadata:
+  name: case32config
+spec:
+  evaluationInterval:
+    compliant: 15s
+    noncompliant: 15s
```
JustinKuli, line 8: Nit: you could just leave the evalauationInterval unset, the default is effectively 10 seconds anyway.

JustinKuli (commented):
Looks great! Just small comments that aren't really necessary.

I think https://github.com/open-cluster-management-io/config-policy-controller/pull/145 will fix the timeout in the most recent CI run, so we could wait for that to merge, then re-base this. Or maybe a re-run will randomly work.

controllers/configurationpolicy_controller.go
```diff
@@ -2556,6 +2558,30 @@ func handleSingleKey(
 			mergedValue.(map[string]interface{}), existingValue.(map[string]interface{}))
 	}
 
+	if key == "stringData" && existingObj.GetKind() == "Secret" {
+		// override automatic conversion from stringData to data prior to evaluation
+		existingValue = existingObj.UnstructuredContent()["data"]
```
dhaiducek, line 2563: Optional/FYI: There's an `unstructured` function, `unstructured.NestedStringMap()` that could replace the type assertions and grab the `map[string]string` directly, though it's not a huge benefit here.
```
encodedValue, _, err := unstructured.NestedStringMap(existingObj.Object, "data")
```

dhaiducek (commented):
Nice and clean solution! I dropped a couple comments to consider.

controllers/configurationpolicy_controller.go
```diff
@@ -2556,6 +2558,30 @@ func handleSingleKey(
 			mergedValue.(map[string]interface{}), existingValue.(map[string]interface{}))
 	}
 
+	if key == "stringData" && existingObj.GetKind() == "Secret" {
```
yiraeChristineKim, line 2561: I am curious about what happens and what is our expected result when a user adds the string `stingDate`, in other words, not encoded data. for example,
```
- complianceType: musthave
          objectDefinition:
            apiVersion: v1
            kind: Secret
            metadata:
              name: htpasswd-secret
              namespace: openshift-config
            stringData:
              htpasswd: "password"
```

yiraeChristineKim: somewhere throw an error?

test/e2e/case32_secret_stringdata_test.go
```diff
@@ -0,0 +1,66 @@
...
+	It("Verifies the config policy stays compliant "+case32ConfigPolicyName+" in "+testNamespace, func() {
+		By("Checking the events on the configuration policy" + case32ConfigPolicyName)
+		Consistently(func() int {
+			eventlen := len(utils.GetMatchingEvents(
+				clientManaged, testNamespace,
+				case32ConfigPolicyName, case32ConfigPolicyName,
+				"NonCompliant;", defaultTimeoutSeconds))
+
+			return eventlen
+		}, defaultTimeoutSeconds, 5).Should(BeNumerically("<", 3))
```
yiraeChristineKim, line 57: We need to increase consistently checking the time, cuz I tested with an unchanged(just main branch build) controller accidentally.. it passed

yiraeChristineKim: Also the screenshot in the ticket, flapping happened every **one min**

yiraeChristineKim: I suggest ```Should(Equal(1))``` to more accurate test

JeffeyL: This ended up being a bigger issue than I initially thought - turns out that the old code and the new code had the same number of events containing `NonCompliant;` as well as `Compliant;`, but the messages were slightly different. The old one would say in the message that the Secret is being "updated", while the new (correct) code has an event with a message that states that there are no actions to be taken. Thus, I modified the `Consistently` block to check for the `updated` keyword in the message instead of the original `NonCompliant` keyword.

Also, I ended up adding the evaluationInterval back in the yaml. Manual testing showed that by default, the policy was not getting evaluated often enough (around every minute or so) which was increasing the testing period.

yiraeChristineKim: If only adding evaluationInterval works.... this is another issue.. can you raise this in our standup?

yiraeChristineKim: Also the messages should present`NonCompliant` or `Compliant`

controllers/configurationpolicy_controller.go
```diff
@@ -2411,6 +2412,7 @@ func handleSingleKey(
 
 	desiredValue := formatTemplate(desiredObj, key)
 	existingValue := existingObj.UnstructuredContent()[key]
+
```
yiraeChristineKim, line 2415: Can you remove this blank

yiraeChristineKim (dismissed):

yiraeChristineKim (dismissed):

test/resources/case32_secret_stringdata/case32_create_secret.yaml
```diff
@@ -0,0 +1,22 @@
...
+        namespace: managed
+      stringData:
+        htpasswd: |
+          alice:$QGJM$2qku.rrSEWRFD/tjOf.dZQ1q.RaR5tTmfNyZn0$1uI6eDx/Ont37ws
+          bob:S1KPovd0N/iDb9$dlTBH7weXG5rQV$5EPf9KrbJjw2HXy35ujhPw2y0$SIto
+          cadence:W$MeWCMWuKi0ApUyZ8byG5$DNL.3FD$c72kFhyZIX8Sf0yJH2AEOGFsRgEHJ
+  pruneObjectBehavior: None
+  remediationAction: enforce
+  severity: medium
+
```
dhaiducek, line 22: There's some non-newline whitespace on this line
```suggestion

```

dhaiducek (commented):
Minor nit about whitespace, but otherwise I think we can merge this!

test/resources/case32_secret_stringdata/case32_create_secret.yaml
yiraeChristineKim: goood catch!!

JeffeyL: Changed

dhaiducek (approved):

dhaiducek:
@JustinKuli requested changes here, so he'll need to re-review it for it to merge.

JustinKuli (approved):

JeffeyL:
/cherry-pick-release-2.7

PR #150, opened 2023-07-06
Empty label fields in policies are ignored

zyjjay:
Fix parsing issue when policy label fields are set to empty strings

ref: https://issues.redhat.com/browse/ACM-5398

### DIFF
--- a/controllers/configurationpolicy_utils.go
+++ b/controllers/configurationpolicy_utils.go
@@ -197,10 +197,11 @@ func checkFieldsWithSort(mergedObj map[string]interface{}, oldObj map[string]int
 			// extra check to see if value is a byte value
 			mQty, err := apiRes.ParseQuantity(mVal)
 			if err != nil {
-				oVal := oldObj[i]
-				if oVal == nil {
-					oVal = ""
+				oVal, ok := oldObj[i]
+				if !ok {
+					return false
 				}
+
 				// An error indicates the value is a regular string, so check equality normally
 				if fmt.Sprint(oVal) != fmt.Sprint(mVal) {
 					return false
--- a/test/e2e/case33_empty_label_test.go
+++ b/test/e2e/case33_empty_label_test.go
@@ -1,0 +1,68 @@
+// Copyright (c) 2021 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package e2e
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"open-cluster-management.io/config-policy-controller/test/utils"
+)
+
+var _ = Describe("Test resource creation when there are empty labels in configurationPolicy", Ordered, func() {
+	const (
+		case33ConfigPolicy     = "../resources/case33_empty_label_handling/empty-labels-config.yaml"
+		case33ConfigMapYaml    = "../resources/case33_empty_label_handling/configmap.yaml"
+		case33ConfigPolicyName = "case33-empty-labels"
+		case33ConfigMapName    = "case33-configmap"
+		case33Label            = "new-label"
+	)
+
+	It("verifies the configmap"+case33ConfigMapName+" exists in "+testNamespace, func() {
+		By("creating the configmap " + case33ConfigMapName)
+		utils.Kubectl("apply", "-f", case33ConfigMapYaml, "-n", testNamespace)
+
+		By("checking configmap exists in " + testNamespace)
+		cm := utils.GetWithTimeout(clientManagedDynamic, gvrConfigMap, case33ConfigMapName,
+			testNamespace, true, defaultTimeoutSeconds)
+		Expect(cm).NotTo(BeNil())
+	})
+
+	It("verifies the policy "+case33ConfigPolicyName+" exists in "+testNamespace, func() {
+		By("creating the ConfigurationPolicy " + case33ConfigPolicyName)
+		utils.Kubectl("apply", "-f", case33ConfigPolicy, "-n", testNamespace)
+
+		By("checking ConfigurationPolicy exists in " + testNamespace)
+		cfPlc := utils.GetWithTimeout(clientManagedDynamic, gvrConfigPolicy,
+			case33ConfigPolicyName, testNamespace, true, defaultTimeoutSeconds)
+		Expect(cfPlc).NotTo(BeNil())
+	})
+
+	It("verifies the configmap "+case33ConfigMapName+" has the proper labels", func() {
+		By("waiting for policy compliance")
+		Eventually(func() interface{} {
+			plc := utils.GetWithTimeout(clientManagedDynamic, gvrConfigPolicy,
+				case33ConfigPolicyName, testNamespace, true, defaultTimeoutSeconds)
+
+			return utils.GetComplianceState(plc)
+		}, defaultTimeoutSeconds, 1).Should(Equal("Compliant"))
+
+		By("verifying label exists and correct")
+		Eventually(func() interface{} {
+			configmap, _ := clientManaged.CoreV1().ConfigMaps(testNamespace).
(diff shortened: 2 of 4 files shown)


controllers/configurationpolicy_utils.go
```diff
@@ -198,11 +198,8 @@ func checkFieldsWithSort(mergedObj map[string]interface{}, oldObj map[string]int
 			mQty, err := apiRes.ParseQuantity(mVal)
 			if err != nil {
 				oVal := oldObj[i]
-				if oVal == nil {
```
yiraeChristineKim, line 201: Why did you remove this?

test/e2e/case33_empty_label_test.go
```diff
@@ -0,0 +1,77 @@
...
+					node, _ := clientManaged.CoreV1().Nodes().Get(context.TODO(), case33NodeName, metav1.GetOptions{})
+					labelMap := node.GetLabels()
+					found := false
+
+					_, ok := labelMap[case33StorageLabel]
+					if ok {
+						found = true
+					}
+
+					return found
```
yiraeChristineKim, line 58:
```suggestion

					_, ok := labelMap[case33StorageLabel]
					if ok {
						return ture
					}

					return false
```

yiraeChristineKim (changes requested):

test/e2e/case33_empty_label_test.go
```diff
@@ -0,0 +1,77 @@
...
+				}, defaultTimeoutSeconds, 1).Should(BeTrue())
+
+				By("verifying label field is correct")
+				Eventually(func() interface{} {
+					node, _ := clientManaged.CoreV1().Nodes().Get(context.TODO(), case33NodeName, metav1.GetOptions{})
+					labelMap := node.GetLabels()
+					merged := labelMap[case33StorageLabel]
```
yiraeChristineKim, line 65:
```suggestion
					ok, _ := labelMap[case33StorageLabel]
					return ok
```
We wanted to verify whether the label exist or not

kubeconfig_test-managed-control-plane
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
```
yiraeChristineKim, line 1: Where this file used? cannot find

controllers/configurationpolicy_utils.go
zyjjay: When the label that we want to add doesn't exist in the current object, `oVal` is set to `nil`. If we set `oVal` to an empty string here, the comparison `oVal != mVal` will always evaluate to `false` when `mVal` is set to an empty string as well. So the controller never emits an update event on the object. I think it's important to retain the  `nil` value as a way to detect fields that don't exist in the current object.

test/e2e/case33_empty_label_test.go
zyjjay: I verified it in the `Eventually` block above this. This check is to make sure the value of the field is an empty string.

kubeconfig_test-managed-control-plane
zyjjay: Oh I have no idea what this is either, I will fix up my PR to remove this.

test/e2e/case33_empty_label_test.go
dhaiducek: Alternatively, this would probably be better handled by the `HaveKey()` Gomega assertion rather than returning a vague boolean.

You could even use `HaveKeyWithValue()` to assert both the key and the value to combine both of this and the next test.

https://onsi.github.io/gomega/#havekeykey-interface

controllers/configurationpolicy_utils.go
yiraeChristineKim: @JustinKuli Would you help review this?

test/resources/case33_empty_label_handling/empty-labels-config.yaml
```diff
@@ -0,0 +1,16 @@
...
+  severity: high
+  object-templates:
+  - complianceType: musthave
+    objectDefinition:
+      apiVersion: v1
+      kind: Node
+      metadata:
+        name: test-managed-control-plane
+        labels:
+          cluster.ocs.openshift.io/openshift-storage: ""
```
dhaiducek, line 16: As a sanity check, could we also check using a policy that sets `cluster.ocs.openshift.io/openshift-storage: null` to make sure the policy doesn't remove the field? (Though I'm also uncertain what the behavior would be there in terms of compliance...)

controllers/configurationpolicy_utils.go
JustinKuli: I think I'd prefer
```
				oVal, keyPresent := oldObj[i]
				if !keyPresent {
					return false
				}
```

But I think either would work.

test/e2e/case33_empty_label_test.go
```diff
@@ -0,0 +1,77 @@
...
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"open-cluster-management.io/config-policy-controller/test/utils"
+)
+
+const (
+	case33ConfigPolicy     = "../resources/case33_empty_label_handling/empty-labels-config.yaml"
+	case33ConfigPolicyName = "case33-empty-labels"
+	case33NodeName         = "test-managed-control-plane"
```
JustinKuli, line 19: I'd prefer to use a different resource for the test. This relies on using a kind cluster of a specific name. If the test creates a configmap or something, the policy could just label that.

test/resources/case33_empty_label_handling/empty-labels-config.yaml
JustinKuli: It would be nice to have a test just to check what this currently does, and understand if that behavior has changed (or if we change it accidentally in the future)

zyjjay: Currently, having an unset label/setting it to `nil` results in the deletion of **all** the labels attached to the object. I'll probably make a separate ticket to address this issue as it's a bit more involved and possibly affect annotations as well (need to confirm this but they both use the same helper function to populate the fields).

JustinKuli (dismissed):
/hold

Looks good to me! Thanks for the changes to the tests, it looks great now. Just holding for any other reviewer comments.

test/e2e/case33_empty_label_test.go
```diff
@@ -0,0 +1,68 @@
...
+var _ = Describe("Test resource creation when there are empty labels in configurationPolicy", Ordered, func() {
+	const (
+		case33ConfigPolicy     = "../resources/case33_empty_label_handling/empty-labels-config.yaml"
+		case33ConfigMapYaml    = "../resources/case33_empty_label_handling/configmap.yaml"
+		case33ConfigPolicyName = "case33-empty-labels"
+		case33ConfigMapName    = "case33-configmap"
+		case33Label            = "new-label"
+	)
+
+	It("verifies the configmap"+case33ConfigMapName+" exists in "+testNamespace, func() {
```
yiraeChristineKim, line 25: I don't understand why we need this step

zyjjay: The issue that the customer was facing was that when applying the label to a preexisting `Node` object, the empty labels are ignored and not applied. I'm just replicating a similar scenario except using `ConfigMaps` instead of a `Node` object.

mprahl (commented):
This looks great! I just left a minor comment in the test.

test/e2e/case33_empty_label_test.go
yiraeChristineKim: Oh now understand, The precondition is the existing object. Thanks @zyjjay

yiraeChristineKim (approved):

JustinKuli (approved):

mprahl:
@zyjjay the commit title should start with a verb in the imperative mood (i.e. Fix ....). Please do this next time.

mprahl (approved):
/unhold

mprahl:
> /cherry-pick release-2.7

You have to do that on the Stolostron PR:
https://github.com/stolostron/config-policy-controller/pull/568

PR #152, opened 2023-07-07
Remove `fail-fast` for Ginkgo

dhaiducek:
Parallel tests cause parallel processes to get interrupted on a failure, creating too much noise (and it's unnecessary now that the tests complete in a reasonable time).

The `-p` flag is also redundant since `--procs` is already specifying parallelism since its default is 1.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,7 +201,7 @@ install-resources:
 
 .PHONY: e2e-test
 e2e-test: e2e-dependencies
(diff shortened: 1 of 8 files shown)


yiraeChristineKim (dismissed):

dhaiducek:
There was one flake that showed up which I've added some additional logging for if it shows up again, but the fail fast flag removal was also working as expected, so this is ready for review.

JustinKuli (approved):

PR #153, opened 2023-07-07
Restrict permissions

dhaiducek:
Listing verbs explicitly excludes such verbs as "escalate", "bind", and "impersonate".

ref: https://issues.redhat.com/browse/ACM-5408


dhaiducek:
It makes sense now that I think about it, but it looks like the permissions will need to stay in place since unless the controller has these permissions, it can't grant them to others, so removing the permissions would be a regression on what the controller can do:
```
Could not create object
user [...] is attempting to grant RBAC permissions not currently held [...] Verbs:["deletecollection"]
```

PR #154, opened 2023-07-17
Add event log when deatached objs deleted

yiraeChristineKim:
Description: As a user, I would like to see the warning log when an object is deleted by the config controller.

Ref: https://issues.redhat.com/browse/ACM-5917


yiraeChristineKim:
already added log at line 650 in configurationplicy_controller.go log.Info("Object successfully deleted as part of child object pruning or detached objects")

PR #156, opened 2023-08-07
Reduce the copying when evaluating a policy

mprahl:
This reduces the recursive copying to just the value in the object
template since that is what is used for merging the existing value in.
For safety, the existing object that is used to merge into the template
object is copied just once up front.


mprahl:
@JustinKuli what are your thoughts on this one? I'm aiming to reduce memory consumption and this seemed like a decent candidate.

yiraeChristineKim (approved):

PR #157, opened 2023-08-08
NamespaceSelector updates triggering evaluations

JustinKuli:
This approach uses the new feature in github.com/stolostron/kubernetes-dependency-watches, but it comes with some downsides (compare to https://github.com/open-cluster-management-io/config-policy-controller/pull/158):
- that library doesn't help if the NamespaceSelector only uses Include and Exclude
- extra evaluations can be triggered when there aren't actually updates

This implementation also uses a channel for communication as opposed to a shared map that would likely require a mutex, but I don't know if this approach is actually better.


mprahl:
As of now, I prefer the other approach.

JustinKuli:
Closed in favor of #158

PR #158, opened 2023-08-08
NamespaceSelector 'reconciler' to help trigger evaluations

JustinKuli:
Instead of a watch for each policy's namespace selector, this implementation uses a separate controller, watching all the namespaces on the target cluster. (Note: since the existing controller manager was based on the cluster the controller is running on, but this is not always the same as the target cluster, some additional setup was needed.)

Being able to list all of the namespaces on the target cluster is needed for policies that only use Include and Exclude in their NamespaceSelector. Since that list can be cached, it seems nicer to the API server to do the LabelSelector calculations locally, and just have the one watch.

This implementation addresses the limitations noted in https://github.com/open-cluster-management-io/config-policy-controller/pull/157, and has adjusted the tests accordingly.

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -120,6 +120,7 @@ type ConfigurationPolicyReconciler struct {
 	TargetK8sClient        kubernetes.Interface
 	TargetK8sDynamicClient dynamic.Interface
 	TargetK8sConfig        *rest.Config
+	SelectorReconciler     common.SelectorReconciler
 	// Whether custom metrics collection is enabled
 	EnableMetrics bool
 	discoveryInfo
@@ -149,6 +150,8 @@ func (r *ConfigurationPolicyReconciler) Reconcile(ctx context.Context, request c
 			prometheus.Labels{"policy": fmt.Sprintf("%s/%s", request.Namespace, request.Name)})
 		_ = policyUserErrorsCounter.DeletePartialMatch(prometheus.Labels{"template": request.Name})
 		_ = policySystemErrorsCounter.DeletePartialMatch(prometheus.Labels{"template": request.Name})
+
+		r.SelectorReconciler.Stop(request.Name)
 	}
 
 	return reconcile.Result{}, nil
@@ -236,7 +239,7 @@ func (r *ConfigurationPolicyReconciler) PeriodicallyExecConfigPolicies(
 
 				for i := range policiesList.Items {
 					policy := policiesList.Items[i]
-					if !shouldEvaluatePolicy(&policy, cleanupImmediately) {
+					if !r.shouldEvaluatePolicy(&policy, cleanupImmediately) {
 						continue
 					}
 
@@ -346,7 +349,9 @@ func (r *ConfigurationPolicyReconciler) refreshDiscoveryInfo() error {
 // met, then that will also trigger an evaluation. If cleanupImmediately is true, then only policies
 // with finalizers will be ready for evaluation regardless of the last evaluation.
 // cleanupImmediately should be set true when the controller is getting uninstalled.
-func shouldEvaluatePolicy(policy *policyv1.ConfigurationPolicy, cleanupImmediately bool) bool {
+func (r *ConfigurationPolicyReconciler) shouldEvaluatePolicy(
+	policy *policyv1.ConfigurationPolicy, cleanupImmediately bool,
+) bool {
 	log := log.WithValues("policy", policy.GetName())
 
 	// If it's time to clean up such as when the config-policy-controller is being uninstalled, only evaluate policies
@@ -356,19 +361,19 @@ func shouldEvaluatePolicy(policy *policyv1.ConfigurationPolicy, cleanupImmediate
 	}
 
 	if policy.ObjectMeta.DeletionTimestamp != nil {
-		log.V(2).Info("The policy has been deleted and is waiting for object cleanup. Will evaluate it now.")
+		log.V(1).Info("The policy has been deleted and is waiting for object cleanup. Will evaluate it now.")
 
 		return true
 	}
 
 	if policy.Status.LastEvaluatedGeneration != policy.Generation {
-		log.V(2).Info("The policy has been updated. Will evaluate it now.")
+		log.V(1).Info("The policy has been updated. Will evaluate it now.")
 
 		return true
 	}
 
 	if policy.Status.LastEvaluated == "" {
-		log.V(2).Info("The policy's status.lastEvaluated field is not set. Will evaluate it now.")
+		log.V(1).Info("The policy's status.lastEvaluated field is not set. Will evaluate it now.")
 
 		return true
 	}
@@ -387,13 +392,23 @@ func shouldEvaluatePolicy(policy *policyv1.ConfigurationPolicy, cleanupImmediate
 	} else if policy.Status.ComplianceState == policyv1.NonCompliant && policy.Spec != nil {
 		interval, err = policy.Spec.EvaluationInterval.GetNonCompliantInterval()
 	} else {
-		log.V(2).Info("The policy has an unknown compliance. Will evaluate it now.")
+		log.V(1).Info("The policy has an unknown compliance. Will evaluate it now.")
+
+		return true
+	}
+
+	usesSelector := policy.Spec.NamespaceSelector.MatchLabels != nil ||
+		policy.Spec.NamespaceSelector.MatchExpressions != nil ||
+		len(policy.Spec.NamespaceSelector.Include) != 0
+
+	if usesSelector && r.SelectorReconciler.HasUpdate(policy.Name) {
+		log.V(1).Info("There was an update for this policy's namespaces. Will evaluate it now.")
 
 		return true
 	}
 
 	if errors.Is(err, policyv1.ErrIsNever) {
-		log.Info("Skipping the policy evaluation due to the spec.evaluationInterval value being set to never")
+		log.V(1).Info("Skipping the policy evaluation due to the spec.evaluationInterval value being set to never")
 
 		return false
 	} else if err != nil {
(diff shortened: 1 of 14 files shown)


main.go
```diff
@@ -266,9 +269,18 @@ func main() {
 			os.Exit(1)
 		}
 
+		targetK8sConfig.Burst = int(opts.clientBurst)
+		targetK8sConfig.QPS = opts.clientQPS
```
JustinKuli, line 273: This is probably needed in general.

mprahl: Good catch

main.go
```diff
@@ -277,6 +289,16 @@ func main() {
 
 	instanceName, _ := os.Hostname() // on an error, instanceName will be empty, which is ok
 
+	// A separate controller is needed because of hosted mode concerns.
+
+	nsSelReconciler := common.NamespaceSelectorReconciler{
+		Client: nsSelMgr.GetClient(),
+	}
+	if err = nsSelReconciler.SetupWithManager(nsSelMgr); err != nil {
```
mprahl, line 297: Do we need another manager? I'd prefer to consolidate with the existing manager. This will prevent your `neverEnqueue` workaround from creating an additional watch on ConfigurationPolicy objects.

mprahl: Oh, I see. Hosted mode would require a separate manager.

pkg/common/namespace_selection.go
```diff
@@ -103,3 +116,177 @@ func GetAllNamespaces(client kubernetes.Interface, labelSelector metav1.LabelSel
...
+		return []reconcile.Request{{NamespacedName: types.NamespacedName{
+			Name: "NamespaceSelector",
+		}}}
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("NamespaceSelector").
+		For(
+			&policyv1.ConfigurationPolicy{},
+			builder.WithPredicates(neverEnqueue)).
```
mprahl, line 141: Will this be an issue in hosted mode since the configuration policy CRD would not be installed on the hosted cluster?

main.go
```diff
@@ -277,6 +289,16 @@ func main() {
 
 	instanceName, _ := os.Hostname() // on an error, instanceName will be empty, which is ok
 
+	// A separate controller is needed because of hosted mode concerns.
+
+	nsSelReconciler := common.NamespaceSelectorReconciler{
+		Client: nsSelMgr.GetClient(),
```
mprahl, line 295: Could you please limit the fields in the namespaces that are cached by using the `Transform` function?

pkg/common/namespace_selection.go
```diff
@@ -103,3 +116,177 @@ func GetAllNamespaces(client kubernetes.Interface, labelSelector metav1.LabelSel
...
+	}
+
+	// unlock for now, the list filtering could take a non-trivial amount of time
+	r.lock.Unlock()
+
+	// New, or the target has changed.
+	namespaces := corev1.NamespaceList{}
+
+	// This List will be from the cache
+	if err := r.Client.List(context.TODO(), &namespaces); err != nil {
```
mprahl, line 227: You can use a label selector in the query when a label selector is used and not the include/exclude approach.

pkg/common/namespace_selection.go
```diff
@@ -103,3 +116,177 @@ func GetAllNamespaces(client kubernetes.Interface, labelSelector metav1.LabelSel
...
+
+func (r *NamespaceSelectorReconciler) HasUpdate(name string) bool {
+	r.lock.RLock()
+	defer r.lock.RUnlock()
+
+	if sel, found := r.selections[name]; found {
+		return sel.hasUpdate
+	}
+
+	return true // no selections populated, call Get to populate it
```
mprahl, line 262: I think this should return false since wouldn't this make `shouldEvaluatePolicy` always return true for those without namespace selectors?

pkg/common/namespace_selection.go
```diff
@@ -103,3 +116,177 @@ func GetAllNamespaces(client kubernetes.Interface, labelSelector metav1.LabelSel
...
+	selections map[string]NamespaceSelection
+	lock       sync.RWMutex
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *NamespaceSelectorReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	r.selections = make(map[string]NamespaceSelection)
+
+	neverEnqueue := predicate.NewPredicateFuncs(func(o client.Object) bool { return false })
+	mapToSingleton := func(_ client.Object) []reconcile.Request {
```
mprahl, line 131: This is a nice idea to reduce reconciles, however, a comment would be appreciated explaining this.

main.go
JustinKuli: And in non-hosted mode, a separate manager isn't created.

pkg/common/namespace_selection.go
JustinKuli: Let me test it. I think it might still get populated on the first `Get`.

JustinKuli: Nevermind, I think you're right

pkg/common/namespace_selection.go
JustinKuli: I don't know which happens, but I believe that setting a selector here would either:
- do a list against the API server, which is then watched and cached, or
- just use the cache, and "locally" handle the selector

I definitely don't want the extra API server list or watch, that seems unnecessary. And I don't think there would be much of a memory consumption difference if the cache somehow does the selector, or if this controller does it in the `filter` function.

Or am I missing something?

mprahl: Discussed about this outside of GitHub, but the summary is:
* The cache uses the existing watch and performs the label selector matching locally.
* This local match occurs before the cache deep copies the objects returned from the `List` method so memory consumption is decreased as opposed to filtering after the `List` method.

JustinKuli:
Still a draft, I want to add some more doc comments, but I think this is functionally very close to done.

pkg/common/namespace_selection.go
JustinKuli: I addressed this in `shouldEvaluatePolicy` - when the selector wouldn't be used, it doesn't even check if the SelectorReconciler thinks here is an update (avoiding a read lock).

pkg/common/namespace_selection.go
```diff
@@ -103,3 +116,199 @@ func GetAllNamespaces(client kubernetes.Interface, labelSelector metav1.LabelSel
...
+	for name, selection := range r.selections {
+		oldSelections[name] = selection
+	}
+
+	r.lock.RUnlock()
+
+	namespaces := corev1.NamespaceList{}
+
+	// This List will be from the cache
+	if err := r.Client.List(ctx, &namespaces); err != nil {
```
mprahl, line 180: If `r.selections` is empty, could you please return before calling `List`? That way if there are no policies with namespaceSelector then you can avoid the memory allocation.

pkg/common/namespace_selection.go
```diff
@@ -103,3 +116,199 @@ func GetAllNamespaces(client kubernetes.Interface, labelSelector metav1.LabelSel
...
+
+	for name, oldSelection := range oldSelections {
+		newNamespaces, err := filter(namespaces, oldSelection.target)
+		if err != nil {
+			log.Error(err, "Unable to filter namespaces for policy", "name", name)
+
+			r.update(name, NamespaceSelection{
+				target:     oldSelection.target,
+				namespaces: newNamespaces,
+				hasUpdate:  oldSelection.err != nil, // it has an update if the error state changed
```
mprahl, line 194: Did you mean this?
```suggestion
				hasUpdate:  oldSelection.err == nil, // it has an update if the error state changed
```

pkg/common/namespace_selection.go
```diff
@@ -103,3 +116,199 @@ func GetAllNamespaces(client kubernetes.Interface, labelSelector metav1.LabelSel
...
+				err:        err,
+			})
```
mprahl, line 196: Do you want to use a `continue` after the `Update` call?

pkg/common/namespace_selection.go
mprahl: I think the `evaluationInterval` CRD doc might need to change to indicate that an updated namespaceSelector result will cause the policy to be reevaluated.

mprahl: Also, `HasUpdate` is only called in `shouldEvaluatePolicy`. I think it'd be cleaner to just return `false`.

JustinKuli: Well, it's certainly cleaner, since if the map is missing the key, the empty value for the struct is returned, which then has `false` in the field this is looking for. So now it's a one-liner 😆 

I've updated the CRD field description.

controllers/configurationpolicy_controller.go
```diff
@@ -387,13 +392,23 @@ func shouldEvaluatePolicy(policy *policyv1.ConfigurationPolicy, cleanupImmediate
...
+		log.V(1).Info("The policy has an unknown compliance. Will evaluate it now.")
+
+		return true
+	}
+
+	usesSelector := policy.Spec.NamespaceSelector.MatchLabels != nil ||
+		policy.Spec.NamespaceSelector.MatchExpressions != nil ||
+		len(policy.Spec.NamespaceSelector.Include) != 0
+
+	if usesSelector && r.SelectorReconciler.HasUpdate(policy.Name) {
```
mprahl, line 404: I think this would be better since if the policy has no namespace selector it would just return false.
```suggestion
	if r.SelectorReconciler.HasUpdate(policy.Name) {
```

test/resources/case19_ns_selector/case19_cm_manifest.yaml
yiraeChristineKim, line 1: Is this file used?

controllers/configurationpolicy_controller.go
JustinKuli: My thought was to avoid `HasUpdate` when possible, because it requires getting a read lock.

test/resources/case19_ns_selector/case19_cm_manifest.yaml
JustinKuli: Good catch! This and case19_cm_long_interval_policy.yaml are unused

mprahl (approved):

PR #160, opened 2023-08-22
ACM-6596: Initialize controller for OperatorPolicy

JeffeyL:
ref: https://issues.redhat.com/browse/ACM-6596

### DIFF
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -5,19 +5,40 @@ package controllers
 
 import (
 	"context"
+	"fmt"
+	"time"
 
-	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	policyv1 "open-cluster-management.io/config-policy-controller/api/v1"
 	policyv1beta1 "open-cluster-management.io/config-policy-controller/api/v1beta1"
 )
+
+const (
+	OperatorControllerName string = "operator-policy-controller"
+)
+
+var OpLog = ctrl.Log.WithName(OperatorControllerName)
 
 // OperatorPolicyReconciler reconciles a OperatorPolicy object
 type OperatorPolicyReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
 }
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *OperatorPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named(OperatorControllerName).
+		For(&policyv1beta1.OperatorPolicy{}).
+		Complete(r)
+}
+
+// blank assignment to verify that OperatorPolicyReconciler implements reconcile.Reconciler
+var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
 
 //+kubebuilder:rbac:groups=policy.open-cluster-management.io,resources=operatorpolicies,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=policy.open-cluster-management.io,resources=operatorpolicies/status,verbs=get;update;patch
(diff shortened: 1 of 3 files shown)


controllers/operatorpolicy_controller.go
```diff
@@ -34,16 +66,137 @@ type OperatorPolicyReconciler struct {
...
+// updatePolicyStatus updates the status of the operatorPolicy.
+//
+// In the future, a condition should be added as well, and this should generate events.
+func (r *OperatorPolicyReconciler) updatePolicyStatus(
+	policy *policyv1beta1.OperatorPolicy,
+) error {
+	updatedStatus := policy.Status
+
+	maxRetries := 3
+	for i := 1; i <= maxRetries; i++ {
```
JustinKuli, line 164: I think we've tried this kind of retry thing before, but it eventually got removed (you've probably been digging through the code more recently than me though, so feel free to correct me). In my opinion, retries here aren't necessary - if it fails, it will just get retried next loop anyway.

JustinKuli: Especially as the status becomes more complex (when there are more things that could cause errors) I think it will just be easier to return the error immediately, and handle the retry at a higher level.

controllers/operatorpolicy_controller.go
```diff
@@ -34,16 +66,137 @@ type OperatorPolicyReconciler struct {
...
+// In the future, more reconciliation logic will be added. For reference:
+// https://github.com/JustinKuli/ocm-enhancements/blob/89-operator-policy/enhancements/sig-policy/89-operator-policy-kind/README.md
+func (r *OperatorPolicyReconciler) handleSinglePolicy(
+	policy *policyv1beta1.OperatorPolicy,
+) error {
+	// when testing, evaluate at every interval regardless of policy status
+	if testFlag {
+		policy.Status.ComplianceState = policyv1.Compliant
+
+		err := r.updatePolicyStatus(policy)
```
JustinKuli, line 142: I don't think this interacts with the `shouldEvaluate` logic the way you might want... It looks like it will only ever update the status if this `testFlag` is true, and there is no current way to turn that on. I think the flag can just be removed.

controllers/operatorpolicy_controller.go
```diff
@@ -34,16 +66,137 @@ type OperatorPolicyReconciler struct {
...
-func (r *OperatorPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
-		For(&policyv1beta1.OperatorPolicy{}).
-		Complete(r)
+// PeriodicallyExecOperatorPolicies loops through all operatorpolicies in the target namespace and triggers
+// template handling for each one. This function drives all the work the operator policy controller does.
+func (r *OperatorPolicyReconciler) PeriodicallyExecOperatorPolicies(
+	ctx context.Context, freq uint, elected <-chan struct{},
+) {
+	_ = ctx // currently unused
```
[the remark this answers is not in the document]
JustinKuli: But actually, you can use the given context in the `r.List(` call below.

controllers/operatorpolicy_controller.go
```diff
@@ -5,20 +5,52 @@ package controllers
...
 
+const (
+	OperatorControllerName string = "configuration-policy-controller"
+	OperatorCRDName        string = "configurationpolicies.policy.open-cluster-management.io"
+)
+
+var (
+	OpLog = ctrl.Log.WithName(OperatorControllerName)
+	// OperatorPlcChan a channel used to pass operator policies ready for update.
+	OperatorPlcChan chan *policyv1beta1.OperatorPolicy
```
JustinKuli, line 31: I know I've seen this in other controllers, but I don't think it's used. Until I see a possible use for it, I'd rather keep it out of here.

main.go
```diff
@@ -335,10 +337,22 @@ func main() {
...
+
 	if err = reconciler.SetupWithManager(mgr); err != nil {
 		log.Error(err, "Unable to create controller", "controller", "ConfigurationPolicy")
 		os.Exit(1)
 	}
+
+	if err = OpReconciler.SetupWithManager(mgr); err != nil {
+		log.Error(err, "Unable to create controller", "controller", "OperatorPolicy")
+		os.Exit(1)
+	}
```
JustinKuli, line 355: Can we put this behind a feature flag? Something that would default to false in `parseOpts`. Then when we want to test stuff with it, we can turn it on by editing the deployment or something.

main.go
```diff
@@ -362,6 +376,11 @@ func main() {
 		managerCancel()
 	}()
 
+	go func() {
+		OpReconciler.PeriodicallyExecOperatorPolicies(terminatingCtx, opts.frequency, mgr.Elected())
+		managerCancel()
+	}()
```
JustinKuli, line 382: This should also go behind the feature flag.

controllers/operatorpolicy_controller.go
JustinKuli: I found where this kind of thing *is* in the normal config-policy-controller... it looks like it was useful there because of some non-idempotent stuff: if the config-policy should create something, but then can't record that it created it in its status, then there will be buggy behavior on a later reconcile because it incorrectly thinks it *didn't* create the resource.

So... let's keep an eye out for that kind of situation in the operator policy controller, and see if this ends up being helpful. I think a better approach to the config-policy situation might have been to add an annotation or label to the object that the config-policy-controller created, that way it could determine later whether it had created it or not even if the status didn't get updated correctly.

JeffeyL: Thanks for pointing this out, I also believe that it would get retried the next loop. I'll keep an eye out for the situations you described in your last comment.

controllers/operatorpolicy_controller.go
JeffeyL: Oh yes definitely, having that if statement in there was an oversight from me, I had originally thought of using it in some way but it ended up not being of significant use.

main.go
JeffeyL: I'll look into this.

JeffeyL:
Changes:
- Removed ctx from `PeriodicallyExecOperatorPolicies`, I checked out the commit that added it ([here](https://github.com/zyjjay/config-policy-controller/commit/82a7f869022deba6afc3ec9f928a9a02dec02569)) and it seems to be related to having cleaner exit/uninstall behavior. It states in a comment that ctx should not be passed so I opted to not include it. May revisit this in the future if the operator policy controller requires similar behavior.
- Removed the retries
- Added a feature flag `--enable-operator-policy`, to enable put `--enable-operator-policy=true` in the launch script

controllers/operatorpolicy_controller.go
```diff
@@ -5,20 +5,46 @@ package controllers
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	policyv1 "open-cluster-management.io/config-policy-controller/api/v1"
 	policyv1beta1 "open-cluster-management.io/config-policy-controller/api/v1beta1"
 )
 
+const (
+	OperatorControllerName string = "configuration-policy-controller"
+	OperatorCRDName        string = "configurationpolicies.policy.open-cluster-management.io"
```
JustinKuli, line 25: Nit: `OperatorCRDName` isn't used, and `OperatorControllerName` should probably be "operator-policy-controller"

controllers/operatorpolicy_controller.go
```diff
@@ -5,20 +5,46 @@ package controllers
...
+)
+
+var OpLog = ctrl.Log.WithName(OperatorControllerName)
+
 // OperatorPolicyReconciler reconciles a OperatorPolicy object
 type OperatorPolicyReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme   *runtime.Scheme
+	Recorder record.EventRecorder
```
JustinKuli, line 34: Nit: I don't know if this reconciler will need these, I think the recorder in particular emits events asynchronously (and unreliably).

JustinKuli (approved):
/hold

Looks good enough to me! Some minor things you could address, but if you want to just move forward with the next steps, feel free to unhold this.

controllers/operatorpolicy_controller.go
JeffeyL: Whoops, that's slightly embarrassing

controllers/operatorpolicy_controller.go
JeffeyL: Did a quick scan through the other controller repos and they all seem to include a scheme in the controller, but proceed not to use it within the reconcile logic (some of the test code does use it, however). The kubebuilder docs have schemes in their controllers as well, but they actually use them for some function calls. Either way, the inclusion/exclusion does not seem to affect the functionality of the controller in any way that I can tell, so it seems to be safe to remove for now, and may be brought back for testing later once that is a concern.

JustinKuli:
It looks like it's failing test coverage, probably because with the feature flag, the new code never runs 😆 sorry about that.

Try adding the new flag to be true in https://github.com/open-cluster-management-io/config-policy-controller/blob/main/main_test.go#L19, that should get it running in the tests, which might just push it over the coverage threshold.

JustinKuli (approved):
/unhold

LGTM!

PR #161, opened 2023-08-30
Fix compliance when created resource has a status

JustinKuli:
Previously, the template would always be marked as compliant after the resource it defines was created. But if the object definition included a `status`, then the object might not actually match what was desired. Now, in this case, the template will be marked as non-compliant.

Refs:
 - https://issues.redhat.com/browse/ACM-7020

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -1621,27 +1621,6 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 	return relatedObjects, result
 }
 
-// generateSingleObjReason is a helper function to create a compliant/noncompliant reason for a named object.
-func generateSingleObjReason(objShouldExist bool, compliant bool, exists bool) string {
-	if objShouldExist {
-		if compliant {
-			return reasonWantFoundExists
-		}
-
-		if exists {
-			return reasonWantFoundNoMatch
-		}
-
-		return reasonWantFoundDNE
-	}
-
-	if compliant {
-		return reasonWantNotFoundDNE
-	}
-
-	return reasonWantNotFoundExists
-}
-
 type singleObject struct {
 	policy      *policyv1.ConfigurationPolicy
 	gvr         schema.GroupVersionResource
@@ -1684,15 +1663,16 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 ) {
 	objLog := log.WithValues("object", obj.name, "policy", obj.policy.Name, "index", obj.index)
 
+	result = objectTmplEvalResult{
+		objectNames: []string{obj.name},
+		namespace:   obj.namespace,
+		events:      []objectTmplEvalEvent{},
+	}
+
 	if !exists && obj.shouldExist {
 		// object is missing and will be created, so send noncompliant "does not exist" event regardless of the
 		// remediation action
-		noncompliantReason := generateSingleObjReason(obj.shouldExist, false, exists)
-		result = objectTmplEvalResult{
-			objectNames: []string{obj.name},
-			namespace:   obj.namespace,
-			events:      []objectTmplEvalEvent{{false, noncompliantReason, ""}},
-		}
+		result.events = append(result.events, objectTmplEvalEvent{false, reasonWantFoundDNE, ""})
 
 		// it is a musthave and it does not exist, so it must be created
 		if strings.EqualFold(string(remediation), string(policyv1.Enforce)) {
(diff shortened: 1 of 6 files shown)


JustinKuli:
I can squash the `refactor` commits together if that would be better.

controllers/configurationpolicy_controller.go
```diff
@@ -1703,7 +1703,18 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
...
+				_, hasStatus = tmplObj.Object["status"]
+			}
+
+			if completed && hasStatus {
+				msg += ", but has a status that might not be correct yet"
+				reason += ", but has a status that might not be correct yet"
+				result.events = append(result.events, objectTmplEvalEvent{false, reason, msg})
+			} else {
+				result.events = append(result.events, objectTmplEvalEvent{completed, reason, msg})
+			}
```
dhaiducek, line 1693: This is the creation flow, right? Does it still stay NonCompliant if the policy is updated and the status doesn't match or does some logic need to be there also?

dhaiducek (commented):
Overall looks good, and thanks for the refactor work! The renamed variables at least will make it clearer to read/follow!

controllers/configurationpolicy_controller.go
JustinKuli: Good catch, somehow I didn't think about the update flow having this same potential issue!

JustinKuli: It looks like it might work correctly, but I'll adjust the test to be sure.

JustinKuli: Updated test: https://github.com/open-cluster-management-io/config-policy-controller/compare/eceb957c80fbc032d898a463b02be38fbd6fa600..3421b2310929e5f8b3e6aafc17034c1352487cc3

JustinKuli: Initially, the CI here failed, but from the logs it looked like that was because the config-policy-controller was a little overwhelmed with the parallel tests. Because this test uses the status of *another* configurationpolicy, it can take extra loops to do what is expected... so bumping the timeouts to the default (1 minute) fixed it.

controllers/configurationpolicy_controller.go
```diff
@@ -1688,22 +1663,34 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
...
 			completed, reason, msg, uid, err := r.enforceByCreatingOrDeleting(obj)
 
-			result.events = append(result.events, objectTmplEvalEvent{completed, reason, msg})
+			hasStatus := false
+			if tmplObj, err := unmarshalFromJSON(objectT.ObjectDefinition.Raw); err == nil {
+				_, hasStatus = tmplObj.Object["status"]
+			}
+
+			if completed && hasStatus {
+				msg += ", but has a status that might not be correct yet"
```
mprahl, line 1688: How about this?
```suggestion
				msg += ". The status will be validated on the next evaluation."
```

mprahl (commented):
This looks good! The only thing is to consider changing the noncompliant message.

test/resources/case34_enforce_w_status/config-policy.yaml
```diff
@@ -0,0 +1,33 @@
...
+          remediationAction: inform
+          object-templates:
+            - complianceType: musthave
+              objectDefinition:
+                apiVersion: v1
+                kind: Namespace
+                metadata:
+                  name: default
+        status:
+          lastEvaluatedGeneration: 3
```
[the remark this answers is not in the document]
dhaiducek: Haha--I thought the same thing! I had to give it a double-take.

controllers/configurationpolicy_controller.go
mprahl: Instead, perhaps `msg += ". The status will be verified on the next evaluation."`

test/e2e/case34_enforce_w_status_test.go
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) 2023 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
```
yiraeChristineKim, line 2: the file name what does "w" mean?

JustinKuli: "with" - as in "enforce with status"

test/resources/case34_enforce_w_status/config-policy.yaml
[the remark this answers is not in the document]
JustinKuli: I like config policies, so I put a config policy in my config policy, so I can test config policies with config policies.

controllers/configurationpolicy_controller.go
JustinKuli: Good suggestion! I've reworded it slightly again, and I wanted to keep the comma: if a policy has multiple objects, the messages will be combined with semicolons, so using a period in the middle of a message seemed like it could be weird.

I also adjusted the `reason` to be shorter now.

mprahl (approved):

yiraeChristineKim (approved):

PR #162, opened 2023-09-05
Allow OperatorPolicy to create OLM subscriptions

zyjjay:
Modified the controller logic such that the OperatorPolicy controller can create OLM subscriptions based on the policy spec. Currently, an OperatorGroup is created for each Subscription. Future implementation will support creating OperatorGroups based on installModes supported by the generated CSVs.

ref: https://issues.redhat.com/browse/ACM-6597

### DIFF
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -6,9 +6,13 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
+	operatorv1 "github.com/operator-framework/api/pkg/operators/v1"
+	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


controllers/operatorpolicy_controller.go
```diff
@@ -121,16 +125,84 @@ func (r *OperatorPolicyReconciler) PeriodicallyExecOperatorPolicies(freq uint, e
...
+
+			if err != nil {
+				r.setCompliance(policy, policyv1.NonCompliant)
+				OpLog.Error(err, "Could not handle missing musthave object")
+
+				return err
+			}
+
+			operatorGroup := r.buildOperatorGroup(policy)
+			err = r.Create(context.TODO(), operatorGroup)
```
JustinKuli, line 151: This is fine for the initial prototype, but eventually it will need to consider what to do when there is already an operator group in the namespace. Just a comment here so we don't forget 😀

controllers/operatorpolicy_controller.go
```diff
@@ -176,3 +248,58 @@ func (r *OperatorPolicyReconciler) shouldEvaluatePolicy(
 
 	return true
 }
+
+// buildSubscription bootstraps the subscription spec defined in the operator policy
+// with the apiversion and kind in preparation for resource creation
+func (r *OperatorPolicyReconciler) buildSubscription(
```
JustinKuli, line 254: I don't think this will ever need anything from the Reconciler object, so this could be changed to just a function, as opposed to a method. The advantage to that is that it is then a bit easier to test in a unit test, because it's obvious that you don't need to instantiate a client or anything.

The CI is currently failing for the PR, because it is just barely missing the test coverage goal. I think adjusting this and `buildOperatorGroup` to be regular functions, and adding a unit test for each, would solve this nicely. Eventually of course we'll want e2e tests for OperatorPolicies.

test/resources/operatorplc.yaml
JustinKuli, line 1: Is this file or the `subscription.yaml` file being used for anything yet?

JustinKuli (changes requested):

test/resources/operatorplc.yaml
zyjjay: Ah that was for testing so not used for anything in this iteration, perhaps they could be kept for the unit tests.

JustinKuli (approved):
Thanks!

PR #163, opened 2023-09-19
Add an error when apiVersion is missing

JustinKuli:
The apiVersion field is generally required by kubernetes objects... it's surprising that this situation wasn't already caught by an error when decoding or getting a mapping.

Refs:
 - https://issues.redhat.com/browse/ACM-7127


JustinKuli:
Now that I added useful debugging for the flakey test, it's passed 4 times in a row 😿

gparvin (approved):

PR #168, opened 2023-10-24
Fix items in nested lists not always being matched

JustinKuli:
When a policy is looking for a field in a list, which is itself nested in an object in a list, and when there are multiple items in these lists on the cluster, the controller was not always matching things correctly. This could result in the policy's compliance depending on the order of the items in the nested list on the cluster. (The new test case may be more clear than this problem description.)

This situation is difficult to reproduce: because of the way some lists were compared before, fields that don't seem like they should affect the result can change the outcome.


mprahl (approved):
Thanks for the quick fix!

PR #173, opened 2023-11-03
Enable configuring the controller namespace

dhaiducek:
The `ClusterRoleBinding` had a hardcoded namespace, preventing the controller from running in a custom namespace.
Also, set `create-ns` as a prerequisite so that it doesn't have to be run manually.


JustinKuli (approved):
I couldn't quickly find a `kustomize` command for this (which is a little surprising, it seems like it could be a common thing), so I think the awkwardness of `sed` is probably worth it.

PR #174, opened 2023-11-06
Fix flaky event recording config-policy-controller E2E test

yiraeChristineKim:
Ref: https://issues.redhat.com/browse/ACM-7938

### DIFF
--- a/test/e2e/case13_templatization_test.go
+++ b/test/e2e/case13_templatization_test.go
@@ -482,7 +482,7 @@ var _ = Describe("Test templatization", Ordered, func() {
 					case13PruneTmpErr+"-configmap", "default", true, defaultTimeoutSeconds)
 
 				return configmap
-			}, 30, 1).ShouldNot(BeNil())
+			}, defaultConsistentlyDuration, 1).ShouldNot(BeNil())
 
 			By("Change to valid configmap")
 			utils.Kubectl("patch", "configurationpolicy", case13PruneTmpErr, "--type=json", "-p",
@@ -520,7 +520,7 @@ var _ = Describe("Test templatization", Ordered, func() {
 					case13PruneTmpErr+"-configmap", "default", true, defaultTimeoutSeconds)
 
 				return configmap
-			}, 30, 1).ShouldNot(BeNil())
+			}, defaultConsistentlyDuration, 1).ShouldNot(BeNil())
 		})
 		AfterAll(func() {
 			utils.Kubectl("delete", "configurationpolicy", case13PruneTmpErr,
--- a/test/e2e/case13_templatization_test.go
+++ b/test/e2e/case13_templatization_test.go
@@ -482,7 +482,7 @@ var _ = Describe("Test templatization", Ordered, func() {
 					case13PruneTmpErr+"-configmap", "default", true, defaultTimeoutSeconds)
 
 				return configmap
-			}, defaultTimeoutSeconds, 1).ShouldNot(BeNil())
+			}, 30, 1).ShouldNot(BeNil())
 
 			By("Change to valid configmap")
 			utils.Kubectl("patch", "configurationpolicy", case13PruneTmpErr, "--type=json", "-p",
@@ -520,7 +520,7 @@ var _ = Describe("Test templatization", Ordered, func() {
 					case13PruneTmpErr+"-configmap", "default", true, defaultTimeoutSeconds)
 
 				return configmap
-			}, defaultTimeoutSeconds, 1).ShouldNot(BeNil())
+			}, 30, 1).ShouldNot(BeNil())
 		})
 		AfterAll(func() {
 			utils.Kubectl("delete", "configurationpolicy", case13PruneTmpErr,
--- a/test/e2e/case20_delete_objects_test.go
+++ b/test/e2e/case20_delete_objects_test.go
@@ -219,7 +219,7 @@ var _ = Describe("Test Object deletion", Ordered, func() {
 					case20PodName, "default", true, defaultTimeoutSeconds)
 
 				return pod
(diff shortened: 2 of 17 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -106,7 +106,7 @@ var _ = BeforeSuite(func() {
 	clientManagedDynamic = NewKubeClientDynamic("", kubeconfigManaged, "")
 	defaultImageRegistry = "quay.io/open-cluster-management"
 	testNamespace = "managed"
-	defaultTimeoutSeconds = 60
+	defaultTimeoutSeconds = 100
```
yiraeChristineKim, line 109: Increase time because of minimum version (slow)

test/e2e/case13_templatization_test.go
```diff
@@ -482,7 +482,7 @@ var _ = Describe("Test templatization", Ordered, func() {
 					case13PruneTmpErr+"-configmap", "default", true, defaultTimeoutSeconds)
 
 				return configmap
-			}, defaultTimeoutSeconds, 1).ShouldNot(BeNil())
```
yiraeChristineKim, line 485: consistently timeout

test/e2e/case13_templatization_test.go
```diff
@@ -520,7 +520,7 @@ var _ = Describe("Test templatization", Ordered, func() {
...
```
yiraeChristineKim, line 523: consistently timeout

test/e2e/case20_delete_objects_test.go
```diff
@@ -219,7 +219,7 @@ var _ = Describe("Test Object deletion", Ordered, func() {
 					case20PodName, "default", true, defaultTimeoutSeconds)
 
 				return pod
-			}, defaultTimeoutSeconds, 1).Should(Not(BeNil()))
+			}, 30, 1).Should(Not(BeNil()))
```
yiraeChristineKim, line 222: consistently timeout

test/e2e/e2e_suite_test.go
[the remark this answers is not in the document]
mprahl: @yiraeChristineKim is it only specific tests that fail without this increased default timeout?

test/e2e/case13_templatization_test.go
JustinKuli: What do you mean?

I think this change is good - we don't need to check that the configmap is still there for 60 (or more) seconds, 30 should be fine. But I don't understand why it would be timing out right now?

test/e2e/case20_delete_objects_test.go
```diff
@@ -445,7 +445,7 @@ var _ = Describe("Test Object deletion", Ordered, func() {
 					case20ConfigPolicyNameChange, testNamespace, false, defaultTimeoutSeconds)
 
 				return managedPlc
-			}, defaultTimeoutSeconds, 1).Should(BeNil())
+			}, 30, 1).Should(BeNil())
```
JustinKuli, line 448: I wonder if we should have a separate setting like `defaultConsistentlyDuration`. It's not *necessary*, but this PR makes me feel like it could be nice.

test/e2e/case13_templatization_test.go
yiraeChristineKim: It doesn't any issue but I tried to reduce e2e test time. I think 60 secs is meaningless. Just suggestion. It took over 20mins before but now less than 15 mins

test/e2e/case20_delete_objects_test.go
yiraeChristineKim: `defaultConsistentlyDuration` good idea

test/e2e/e2e_suite_test.go
```diff
@@ -29,23 +29,24 @@ import (
...
-	gvrNS                 schema.GroupVersionResource
-	gvrSCC                schema.GroupVersionResource
-	gvrSecret             schema.GroupVersionResource
-	gvrClusterClaim       schema.GroupVersionResource
-	gvrConfigMap          schema.GroupVersionResource
-	gvrDeployment         schema.GroupVersionResource
-	gvrPolicy             schema.GroupVersionResource
+	testNamespace               string
+	defaultTimeoutSeconds       int
+	defaultConsistentlyDuration int
```
yiraeChristineKim, line 34: I added this only

test/e2e/case5_multi_test.go
```diff
@@ -21,7 +21,7 @@ const (
 	case5ComboYaml               string = "../resources/case5_multi/case5_multi_combo.yaml"
 )
 
-var _ = Describe("Test multiple obj template handling", func() {
+var _ = Describe("Test multiple obj template handling", Ordered, func() {
```
yiraeChristineKim, line 24: This caused minimum version issue

test/e2e/case20_delete_objects_test.go
yiraeChristineKim: 10-12 mins now

JustinKuli (approved):

test/e2e/case23_invalid_field_test.go
```diff
@@ -73,7 +73,7 @@ var _ = Describe("Test an objectDefinition with an invalid field", Ordered, func
 			}
 
 			return startTime.After(compPlcEvents[len(compPlcEvents)-1].LastTimestamp.Time)
-		}, defaultTimeoutSeconds, 1).Should(BeTrue())
+		}, 30, 1).Should(BeTrue())
```
mprahl, line 76: @yiraeChristineKim did you forget to update this value and the other value in this file to use the new global variable?

yiraeChristineKim: I think I missed these 2 Thanks Let me open pr

PR #179, opened 2023-11-22
ACM-8739: ACM Policy that applies stringdata in a secret regression with templates

JeffeyL:
Description: On second evaluation of a secret with stringData, the existingObject is modified while evaluating stringData for comparison. Since the order in which the keys are evaluated is varied, if the underlying object is modified first this may cause issues for future key evaluations.

Solution: when evaluating stringData, operate on a copy of the existing object's data so the actual object is not modified.

Ref: https://issues.redhat.com/browse/ACM-8739

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -2410,9 +2410,7 @@ func handleSingleKey(
 	}
 
 	if key == "stringData" && existingObj.GetKind() == "Secret" {
-		// override automatic conversion from stringData to data prior to evaluation
-		existingValue = existingObj.UnstructuredContent()["data"]
-
+		// override automatic conversion from stringData to data before evaluation
 		encodedValue, _, err := unstructured.NestedStringMap(existingObj.Object, "data")
 		if err != nil {
 			message := "Error accessing encoded data"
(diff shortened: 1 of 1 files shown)


JeffeyL:
The original 2.8 ticket is here: https://issues.redhat.com/browse/ACM-8415, I created a new ticket since this problem may also affect the most recent version.

controllers/configurationpolicy_controller.go
```diff
@@ -2410,16 +2410,16 @@ func handleSingleKey(
...
+		// override automatic conversion from stringData to data before evaluation
 		encodedValue, _, err := unstructured.NestedStringMap(existingObj.Object, "data")
 		if err != nil {
 			message := "Error accessing encoded data"
 
 			return message, false, mergedValue, false
 		}
 
+		existingValue = make(map[string]interface{}, len(encodedValue))
+
```
yiraeChristineKim, line 2422: Should we change this https://github.com/open-cluster-management-io/config-policy-controller/blob/34c5a931b272a414aed605cafac95687e0c6b859/controllers/configurationpolicy_controller.go#L2432? because of this

yiraeChristineKim (dismissed):

controllers/configurationpolicy_controller.go
yiraeChristineKim: https://github.com/open-cluster-management-io/config-policy-controller/blob/34c5a931b272a414aed605cafac95687e0c6b859/controllers/configurationpolicy_controller.go#L2382C2-L2391C4 maybe affect here if `existingValue`  is still interface{}

yiraeChristineKim (approved):

PR #180, opened 2023-11-30
Stop the NS controller manager during hosted mode uninstalls

mprahl:
In hosted mode, there is no guarantee that the managed cluster's API server is available when this controller is being uninstalled. In this case, we can't have any controller-runtime managers connecting to the managed cluster's API server during that time. If there are, the controller can restart without completing the required uninstall cleanup and then can't properly restart since the manager cannot be instantiated.

Relates:
https://issues.redhat.com/browse/ACM-8826


JustinKuli (approved):
Uninstalling is so complicated when you have finalizers 😅 This looks right to me though.

PR #182, opened 2023-12-04
Add relatedObjs to unnamed musthave object

yiraeChristineKim:
What the user wants: 
When the object, even if no having name,  turns compliant, appears in the list. We can see the object of Kind Node, compliant, and it does not match any specific name of the Kind object.

Ref: https://issues.redhat.com/browse/ACM-8782

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -1507,6 +1507,7 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 	}
 
 	var existingObj *unstructured.Unstructured
+	var allResourceNames []string
 
 	if objDetails.name != "" { // named object, so checking just for the existence of the specific object
 		// If the object couldn't be retrieved, this will be handled later on.
@@ -1521,7 +1522,7 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 		log.V(1).Info(
 			"The object template does not specify a name. Will search for matching objects in the namespace.",
 		)
-		objNames = getNamesOfKind(
+		objNames, allResourceNames = getNamesOfKind(
 			desiredObj,
 			mapping.Resource,
 			objDetails.isNamespaced,
@@ -1555,6 +1556,8 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 	}
 
 	objShouldExist := !strings.EqualFold(string(objectT.ComplianceType), string(policyv1.MustNotHave))
+
+	shouldAddCondensedRelatedObj := false
 
 	if len(objNames) == 1 {
 		name := objNames[0]
@@ -1598,6 +1601,15 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 			} else {
 				resultEvent.compliant = false
 				resultEvent.reason = reasonWantFoundDNE
+				// Length of objNames = 0, complianceType == musthave or mustonlyhave
+				// Find Noncompliant resources to add to the status.relatedObjects for debugging purpose
+				shouldAddCondensedRelatedObj = true
+				if objDetails.kind != "" && objDetails.name == "" {
+					// Change reason to Resource found but does not match
+					if len(allResourceNames) > 0 {
+						resultEvent.reason = reasonWantFoundNoMatch
+					}
+				}
 			}
 		} else {
 			if exists {
@@ -1606,21 +1618,36 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 			} else {
 				resultEvent.compliant = true
 				resultEvent.reason = reasonWantNotFoundDNE
+				// Compliant, complianceType == mustnothave
+				// Find resources in the same namespace to add to the status.relatedObjects for debugging purpose
+				shouldAddCondensedRelatedObj = true
 			}
 		}
 
 		result = objectTmplEvalResult{objectNames: objNames, events: []objectTmplEvalEvent{resultEvent}}
 
-		relatedObjects = addRelatedObjects(
-			resultEvent.compliant,
-			mapping.Resource,
-			objDetails.kind,
-			namespace,
-			objDetails.isNamespaced,
-			objNames,
-			resultEvent.reason,
-			nil,
-		)
+		if shouldAddCondensedRelatedObj {
+			// relatedObjs name is *
+			relatedObjects = addCondensedRelatedObjs(
+				mapping.Resource,
+				resultEvent.compliant,
+				objDetails.kind,
+				namespace,
+				objDetails.isNamespaced,
+				resultEvent.reason,
+			)
+		} else {
+			relatedObjects = addRelatedObjects(
+				resultEvent.compliant,
+				mapping.Resource,
+				objDetails.kind,
+				namespace,
+				objDetails.isNamespaced,
+				objNames,
+				resultEvent.reason,
+				nil,
+			)
+		}
 	}
 
 	return relatedObjects, result
@@ -1962,6 +1989,7 @@ func buildNameList(
 
 // getNamesOfKind returns an array with names of all of the resources found
 // matching the GVK specified.
+// allResourceList includes names that are under the same namespace and kind.
 func getNamesOfKind(
 	desiredObj unstructured.Unstructured,
 	rsrc schema.GroupVersionResource,
@@ -1970,30 +1998,31 @@ func getNamesOfKind(
 	dclient dynamic.Interface,
 	complianceType string,
 	zeroValueEqualsNil bool,
-) (kindNameList []string) {
+) (kindNameList []string, allResourceList []string) {
+	var resList *unstructured.UnstructuredList
+	var err error
+
 	if namespaced {
 		res := dclient.Resource(rsrc).Namespace(ns)
 
-		resList, err := res.List(context.TODO(), metav1.ListOptions{})
-		if err != nil {
-			log.Error(err, "Could not list resources", "rsrc", rsrc, "namespaced", namespaced)
+		resList, err = res.List(context.TODO(), metav1.ListOptions{})
+	} else {
+		res := dclient.Resource(rsrc)
 
-			return kindNameList
-		}
-
-		return buildNameList(desiredObj, complianceType, resList, zeroValueEqualsNil)
+		resList, err = res.List(context.TODO(), metav1.ListOptions{})
 	}
 
-	res := dclient.Resource(rsrc)
-
-	resList, err := res.List(context.TODO(), metav1.ListOptions{})
 	if err != nil {
 		log.Error(err, "Could not list resources", "rsrc", rsrc, "namespaced", namespaced)
 
-		return kindNameList
+		return kindNameList, allResourceList
 	}
 
-	return buildNameList(desiredObj, complianceType, resList, zeroValueEqualsNil)
+	for _, res := range resList.Items {
+		allResourceList = append(allResourceList, res.GetName())
+	}
+
+	return buildNameList(desiredObj, complianceType, resList, zeroValueEqualsNil), allResourceList
 }
 
 // enforceByCreatingOrDeleting can handle the situation where a musthave or mustonlyhave object is
--- a/controllers/configurationpolicy_utils.go
+++ b/controllers/configurationpolicy_utils.go
@@ -65,6 +65,45 @@ func addRelatedObjects(
 	return relatedObjects
 }
 
+// addCondensedRelatedObjs does not include all of relatedObjs.
+// The Name field is "*". The list of objects will be presented on the console.
+func addCondensedRelatedObjs(
+	rsrc schema.GroupVersionResource,
+	compliant bool,
+	kind string,
+	namespace string,
+	namespaced bool,
+	reason string,
+) (relatedObjects []policyv1.RelatedObject) {
+	metadata := policyv1.ObjectMetadata{Name: "*"}
+
+	if namespaced {
+		metadata.Namespace = namespace
+	} else {
+		metadata.Namespace = ""
+	}
+
+	// Initialize the related object from the object handling
+	relatedObject := policyv1.RelatedObject{
+		Reason: reason,
+		Object: policyv1.ObjectResource{
+			APIVersion: rsrc.GroupVersion().String(),
+			Kind:       kind,
+			Metadata:   metadata,
+		},
+	}
+
+	if compliant {
+		relatedObject.Compliant = string(policyv1.Compliant)
+	} else {
+		relatedObject.Compliant = string(policyv1.NonCompliant)
+	}
+
+	relatedObjects = append(relatedObjects, relatedObject)
+
+	return relatedObjects
+}
+
 // unmarshalFromJSON unmarshals raw JSON data into an object
 func unmarshalFromJSON(rawData []byte) (unstructured.Unstructured, error) {
 	var unstruct unstructured.Unstructured
--- a/test/e2e/case37_no_name_test.go
+++ b/test/e2e/case37_no_name_test.go
@@ -1,0 +1,396 @@
+package e2e
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"open-cluster-management.io/config-policy-controller/test/utils"
+)
+
+var _ = Describe("Test a namespace-scope policy that is missing name", Ordered, func() {
+	const (
+		case37GoodIngressPath       string = "../resources/case37_no_name/case37_good_ingress.yaml"
+		case37BadIngressPath        string = "../resources/case37_no_name/case37_bad_ingresses.yaml"
+		case37PolicyNSPath          string = "../resources/case37_no_name/case37_no_name_policy.yaml"
+		case37PolicyNSName          string = "case37-test-policy-1"
+		case37PolicyNotHavePath     string = "../resources/case37_no_name/case37_no_name_policy_nothave.yaml"
+		case37PolicyMustnothaveName string = "case37-test-policy-mustnothave"
+	)
+	Describe("Test a musthave", Ordered, func() {
+		BeforeAll(func() {
+			By("Creating a policy on managed")
+			utils.Kubectl("apply", "-f", case37PolicyNSPath, "-n", testNamespace)
+			plc := utils.GetWithTimeout(clientManagedDynamic, gvrConfigPolicy,
+				case37PolicyNSName, testNamespace, true, defaultTimeoutSeconds)
+			Expect(plc).ShouldNot(BeNil())
+		})
+		It("should have 1 NonCompliant relatedObject under the policy's status", func() {
+			By("Apply bad ingresses")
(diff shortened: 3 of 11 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -1970,30 +1991,50 @@ func getNamesOfKind(
 	dclient dynamic.Interface,
 	complianceType string,
 	zeroValueEqualsNil bool,
+	existDesiredObj bool,
```
JustinKuli, line 1994: 8 parameters on this function now 😨

It's worth considering this for a refactor, I think. Not necessarily in this PR, just mentioning it.

yiraeChristineKim: How about this?  		
`desiredObj *unstructured.Unstructured,`
`if desiredObj != nil {`

test/e2e/case37_no_name_test.go
```diff
@@ -0,0 +1,168 @@
...
+	case37GoodIngressPath     string = "../resources/case37_no_name/case37_good_ingress.yaml"
+	case37BadIngressPath      string = "../resources/case37_no_name/case37_bad_ingresses.yaml"
+	case37BadIngressClassPath string = "../resources/case37_no_name/case37_bad_ingressclass.yaml"
+	case37PolicyNSPath        string = "../resources/case37_no_name/case37_no_name_policy.yaml"
+	case37PolicyCSPath        string = "../resources/case37_no_name/case37_no_name_clusterscope_policy.yaml"
+	case37PolicyNSName        string = "case37-test-policy-1"
+	case37PolicyCSName        string = "case37-test-policy-clusterscope"
+)
+
+var _ = Describe("Test a namespace-scope policy that is missing name ", Ordered, func() {
```
JustinKuli, line 21: These tests look great! I see tests for objects when it is compliant *and* when it is non-compliant (I was worried just looking through the code that it only covered one of those cases).

Do these changes also address similar cases for `mustnothave` policies?

yiraeChristineKim:
```
if objShouldExist {
.
.
.
				if objDetails.kind != "" && objDetails.name == "" {
```

It doesn't affect `mustnothave`. Should I add a similar thing to `mustnothave`?

test/e2e/case37_no_name_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"open-cluster-management.io/config-policy-controller/test/utils"
+)
+
+const (
```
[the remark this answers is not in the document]
yiraeChristineKim: Good good

test/e2e/case37_no_name_test.go
JustinKuli: I think when a `mustnothave` policy is Compliant, it would have a similar situation: its relatedObjects list would be empty. It doesn't look like it's necessary to address that for this issue, but it's something to consider in the future. I think it could be nice if there was always something in that list.

yiraeChristineKim: Can you give me example relatedObjects when it is `mustnothave`

controllers/configurationpolicy_controller.go
```diff
@@ -1598,6 +1600,24 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
...
+						objDetails.isNamespaced,
+						namespace,
+						r.TargetK8sDynamicClient,
+						strings.ToLower(string(objectT.ComplianceType)),
+						true,
+					)
+
+					// Change reason to Resource found but does not match
+					if len(objNames) > 0 {
+						shouldAddCondensedRelatedObj = true
```
mprahl, line 1618: Why not just always set this to true instead of duplicating the list query performed in `getNamesOfKind`?

controllers/configurationpolicy_controller.go
```diff
@@ -1963,37 +1995,48 @@ func buildNameList(
...
 
 		return kindNameList
 	}
 
-	return buildNameList(desiredObj, complianceType, resList, zeroValueEqualsNil)
+	if len(resList.Items) == 0 {
+		return kindNameList
+	}
+
+	if desiredObj != nil {
```
mprahl, line 2029: This approach makes it seem a little "magical" in the sense that if I don't provide `desiredObj` I get a completely different behavior out of this function and things such as `complianceType` are ignored.

controllers/configurationpolicy_controller.go
yiraeChristineKim: len(objNames) == 0 there is nothing showing on relatedobj.  in other words there is nothing to debug

controllers/configurationpolicy_controller.go
yiraeChristineKim: so you are suggesting separate func?

controllers/configurationpolicy_controller.go
```diff
@@ -1598,6 +1600,24 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
...
+					)
+
+					// Change reason to Resource found but does not match
+					if len(objNames) > 0 {
+						shouldAddCondensedRelatedObj = true
+					}
+				}
 			}
 		} else {
 			if exists {
```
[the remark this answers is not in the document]
yiraeChristineKim: When `mustnothave` and name == "", users can check at relatedobj.  See this screenshot
Screenshot 2023-12-12 at 12 25 36 PM

``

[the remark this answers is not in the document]
yiraeChristineKim: Therefore, I have focused on `musthave` case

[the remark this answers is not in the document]
JustinKuli: I thought that currently, when a `mustnothave` policy with no name is Compliant, there is nothing in the Related Resources section? It *could* do the same kind of thing that this PR introduces for `musthave` policies with no name when they are NonCompliant.

I think the idea is "can we always put something in the related resources?" And the two cases where we currently don't are:
- `musthave`, no name, noncompliant (fixed in this PR)
- `mustnothave`, no name, compliant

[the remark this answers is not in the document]
yiraeChristineKim: I am working on it! Thanks @JustinKuli

controllers/configurationpolicy_controller.go
mprahl: Some refactoring is required. We shouldn't duplicate `res.List(` when `objNames == 0` (in the calling function) since that's a real API server call and not using a cache.

controllers/configurationpolicy_controller.go
mprahl: I'd prefer if whenever the object template is not named and it's noncompliant, you just set `shouldAddCondensedRelatedObj = true`. Perhaps there are no objects, but that's okay. It makes the behavior consistent.

Also, I'm not sure if `reasonWantFoundNoMatch` is better than `reasonWantFoundDNE`, so if that's why you are duplicating the list query by calling `getNamesOfKind` again, then it's not worth it in my opinion.

yiraeChristineKim: um... to me, `reasonWantFoundNoMatch` is better. Because the name is * in relatedObj and users only see the `reason` on the console. `reasonWantFoundNoMatch` is for relatedObj.. isn't it? @JustinKuli  how about yoiu

controllers/configurationpolicy_controller.go
```diff
@@ -1970,30 +2007,35 @@ func getNamesOfKind(
 	dclient dynamic.Interface,
 	complianceType string,
 	zeroValueEqualsNil bool,
-) (kindNameList []string) {
+) (kindNameList []string, allResourceList []string) {
```
yiraeChristineKim, line 2010: Refactored

controllers/configurationpolicy_controller.go
yiraeChristineKim: I did

controllers/configurationpolicy_controller.go
JustinKuli: Sorry for slow response, if you're still looking for input, I think `reasonWantFoundNoMatch` makes more sense when the name is `*`. It doesn't make as much sense to say something "resource *, not found" when some instances of that resource do exist.

So for `mustnothave`, same kind of thing: if the names would be `*` or something similar, because some resources were found they just didn't fully match the policy criteria, then use a `NoMatch` kind of reason.

JustinKuli (dismissed):
/hold

Looks good to me! I'm not sure if Matt had any remaining requirements

controllers/configurationpolicy_controller.go
```diff
@@ -1598,6 +1601,15 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 			} else {
 				resultEvent.compliant = false
 				resultEvent.reason = reasonWantFoundDNE
+				// Length of objNames = 0, complianceType == musthave and the name field is empty
```
mprahl, line 1604:
```suggestion
				// Length of objNames = 0, complianceType == musthave or mustonlyhave and the name field is empty
```

controllers/configurationpolicy_controller.go
```diff
@@ -1598,6 +1601,15 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
...
+				// Find Noncompliant resources to add to the status.relatedObjects
+				if objDetails.kind != "" && objDetails.name == "" {
+					// Change reason to Resource found but does not match
+					if len(allResourceNames) > 0 {
+						resultEvent.reason = reasonWantFoundNoMatch
+						shouldAddCondensedRelatedObj = true
```
mprahl, line 1610: I think you want to always set this to true if the object template was unnamed except the special case of only one object returned from `getNamesOfKind` which is handled by `handleSingleObj` above.

I think this current code won't add the correct related object entry when no objects were returned from `getNamesOfKind`.

controllers/configurationpolicy_controller.go
```diff
@@ -1606,21 +1618,42 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 			} else {
 				resultEvent.compliant = true
 				resultEvent.reason = reasonWantNotFoundDNE
+
+				// Compliant, complianceType == mustnothave and the name field is empty
+				// Find resources in the same namespace to add to the status.relatedObjects
+				if objDetails.kind != "" && objDetails.name == "" {
+					if len(allResourceNames) > 0 {
+						shouldAddCondensedRelatedObj = true
```
mprahl, line 1626: Same comment as above. I don't think this accounts for unnamed object templates when `getNamesOfKind` returns no objects. In this case, just removing the `if len(allResourceNames) > 0 {` conditional should fix that.

controllers/configurationpolicy_controller.go
```diff
@@ -1606,21 +1618,42 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
...
-			mapping.Resource,
-			objDetails.kind,
-			namespace,
-			objDetails.isNamespaced,
-			objNames,
-			resultEvent.reason,
-			nil,
-		)
+		if shouldAddCondensedRelatedObj {
+			// Length of objNames = 0, complianceType == musthave and the name field is empty
```
mprahl, line 1635: This comment is out of date since `mustnothave` can also set `shouldAddCondensedRelatedObj` to `true`.

controllers/configurationpolicy_controller.go
mprahl: Could you please update the documentation comment of `getNamesOfKind` to indicate what `allResourceList` is?

controllers/configurationpolicy_controller.go
```diff
@@ -1970,30 +2003,35 @@ func getNamesOfKind(
...
 	if err != nil {
 		log.Error(err, "Could not list resources", "rsrc", rsrc, "namespaced", namespaced)
 
-		return kindNameList
+		return kindNameList, allResourceList
+	}
+
+	if len(resList.Items) == 0 {
+		return kindNameList, allResourceList
+	}
```
mprahl, line 2028: Optional: This isn't needed.
```suggestion
```

controllers/configurationpolicy_utils.go
```diff
@@ -65,6 +65,46 @@ func addRelatedObjects(
 	return relatedObjects
 }
 
+// addCondensedRelatedObjs does not include all of relatedObjs.
+// The Name field is "*". The list of objects will be presented on the console.
+// This function is used when it is "musthave", object-template doesn't have "name" of metadata and NonCompliant.
```
mprahl, line 70:
```suggestion
// This function is used when it is "musthave", object-template doesn't have "name" in metadata and NonCompliant.
```

controllers/configurationpolicy_utils.go
```diff
@@ -65,6 +65,46 @@ func addRelatedObjects(
...
+	kind string,
+	namespace string,
+	namespaced bool,
+	reason string,
+) (relatedObjects []policyv1.RelatedObject) {
+	// Initialize the related object from the object handling
+	var relatedObject policyv1.RelatedObject
+
+	if compliant {
+		relatedObject.Compliant = string(policyv1.Compliant)
```
mprahl, line 83: The description of this function indicates you shouldn't call this if it's compliant. So should this code not exist in the function?

controllers/configurationpolicy_utils.go
```diff
@@ -65,6 +65,46 @@ func addRelatedObjects(
...
+	} else {
+		relatedObject.Compliant = string(policyv1.NonCompliant)
+	}
+
+	relatedObject.Properties = nil
```
mprahl, line 88: Isn't it already `nil`?

controllers/configurationpolicy_utils.go
```diff
@@ -65,6 +65,46 @@ func addRelatedObjects(
...
+
+	relatedObject.Reason = reason
+	metadata := policyv1.ObjectMetadata{}
+	metadata.Name = "*"
```
mprahl, line 92:
```suggestion
	metadata := policyv1.ObjectMetadata{Name: "*"}
```

controllers/configurationpolicy_controller.go
yiraeChristineKim: So if there is no noncompliant resource, we present as * then when the user clicks *, query nothing?

controllers/configurationpolicy_controller.go
yiraeChristineKim: hm.. but it goes to the unnecessary process `buildNameList`  I can remove it. It is not a big deal

controllers/configurationpolicy_controller.go
yiraeChristineKim: reason is `Resource found but does not match` To me, it doesn't make sense if there is 0 found resource

mprahl: I was specifically referring to `shouldAddCondensedRelatedObj`.

controllers/configurationpolicy_controller.go
mprahl: Yeah, I think so. It just shows the query that was used for considering objects.

controllers/configurationpolicy_controller.go
yiraeChristineKim: When `allResourceNames = 0`, don't need to add related obj `Resource found but does not match` doesn't match at all

mprahl: Right... but in that case you won't have a related objects entry. I'm saying that when `allResourceNames != 1` (1 object is handled by `handleSingleObj` above), then `shouldAddCondensedRelatedObj = true` should be set to true when doing `musthave` and it's not compliant. The reason is different based on if `allResourceNames > 1`

yiraeChristineKim: I thought `resultEvent.compliant = false` mean objNames is 0, allResourceNames doesn't include matched obj

test/e2e/case37_no_name_test.go
```diff
@@ -0,0 +1,372 @@
...
+	})
+	Describe("Test a mustnothave", Ordered, func() {
+		BeforeAll(func() {
+			By("Creating a policy on managed")
+			utils.Kubectl("apply", "-f", case37PolicyNotHavePath, "-n", testNamespace)
+			plc := utils.GetWithTimeout(clientManagedDynamic, gvrConfigPolicy,
+				case37PolicyMustnothaveName, testNamespace, true, defaultTimeoutSeconds)
+			Expect(plc).ShouldNot(BeNil())
+		})
+		It("should have 0 Compliant relatedObject under the policy's status", func() {
```
mprahl, line 126: I don't think this is correctly describing the test since it's expecting related objects to have a length of 1. Could you please reword it?

test/e2e/case37_no_name_test.go
```diff
@@ -0,0 +1,372 @@
...
+			var managedPlc *unstructured.Unstructured
+
+			Eventually(func() interface{} {
+				managedPlc = utils.GetWithTimeout(clientManagedDynamic, gvrConfigPolicy,
+					case37PolicyMustnothaveName, testNamespace, true, defaultTimeoutSeconds)
+
+				return utils.GetComplianceState(managedPlc)
+			}, defaultTimeoutSeconds, 1).Should(Equal("Compliant"))
+
+			By("relatedObjects should not exist")
```
mprahl, line 136:
```suggestion
			By("relatedObjects should exist")
```

test/e2e/case37_no_name_test.go
```diff
@@ -0,0 +1,372 @@
...
+
+				var err error
+				relatedObjects, _, err = unstructured.NestedSlice(managedPlc.Object, "status", "relatedObjects")
+				g.Expect(err).ShouldNot(HaveOccurred())
+
+				return relatedObjects
+				// Getting relatedObjects takes longer time
+			}, defaultTimeoutSeconds*2, 1).Should(HaveLen(1))
+
+			By("Check the reasons of relatedObjects")
```
mprahl, line 150:
```suggestion
			By("Check the reason of the related object")
```

test/e2e/case37_no_name_test.go
```diff
@@ -0,0 +1,372 @@
...
+			}, defaultTimeoutSeconds*2, 1).Should(HaveLen(1))
+
+			//  NonCompliant relatedObjects Should be deleted and only the Compliant relatedObject remain
+			By("Check the kind of relatedObjects")
+			kind := relatedObjects[0].(map[string]interface{})["object"].(map[string]interface{})["kind"].(string)
+			Expect(kind).Should(Equal("Ingress"))
+
+			By("Check the reason of relatedObjects")
+			reason := relatedObjects[0].(map[string]interface{})["reason"].(string)
+			Expect(reason).Should(Equal("Resource found but should not exist"))
```
mprahl, line 234: Could you please also test that the name is not `*` here?

mprahl (approved):
Looks great!

/hold for @JustinKuli to take another look

JustinKuli (approved):
/unhold

__debug_bin2514188118
dhaiducek, line 1: @yiraeChristineKim I think this file shouldn't have been merged. Can you revert this, please?

PR #183, opened 2023-12-05
Require `objectDefinition` and `remediationAction`

dhaiducek:
Without these fields, the policy is meaningless. When they're optional, a ConfigurationPolicy can be created without an `objectDefinition` and, if there was an attempt to put the object at the `objectDefinition` level (i.e. apiVersion, kind, metadata), all those fields get wiped out without notice for the user. It'd be a better experience to make these required and surface the error sooner.

I tested it out, and the status is returned properly:
>NonCompliant; template-error; Failed to create policy template: ConfigurationPolicy.policy.open-cluster-management.io "7799-policy-crd" is invalid: spec.object-templates[0].objectDefinition: Required value

### DIFF
--- a/api/v1/configurationpolicy_types.go
+++ b/api/v1/configurationpolicy_types.go
@@ -152,8 +152,8 @@ func (e EvaluationInterval) GetNonCompliantInterval() (time.Duration, error) {
 
 // ConfigurationPolicySpec defines the desired state of ConfigurationPolicy
 type ConfigurationPolicySpec struct {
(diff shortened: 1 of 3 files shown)


dhaiducek:
We could alternatively set a default for `remediationAction`?

JustinKuli (approved):
¯\_(ツ)_/¯ seems good to me

mprahl (approved):

mprahl (approved):

PR #189, opened 2024-01-12
ACM-6652: Add installation of specific operators via OperatorPolicy

JeffeyL:
Allows using an OperatorPolicy to install an operator from scratch, without pre-existing resources. This PR also incorporates transition to a `Reconcile` pattern rather than the previous periodic pattern that utilized `PeriodicallyExecOperatorPolicies`. Caching was also added using the kubernetes-dependency-watches library, which required updating some packages.

Ref: https://issues.redhat.com/browse/ACM-6652


yiraeChristineKim:
Oh! Test failed Let me know after the tests fixed

Makefile
```diff
@@ -192,6 +184,11 @@ kind-tests: kind-delete-cluster kind-bootstrap-cluster-dev kind-deploy-controlle
 
 .PHONY: install-crds
 install-crds:
+	@echo installing olm
+	# operator-sdk olm install
+	curl -L https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.26.0/install.sh -o install.sh
+	chmod +x install.sh
+	./install.sh v0.26.0
```
JustinKuli, line 191:
```suggestion
	curl -L https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.26.0/install.sh -o $(LOCAL_BIN)/install.sh
	chmod +x $(LOCAL_BIN)/install.sh
	$(LOCAL_BIN)/install.sh v0.26.0
```

I think most stuff like this already goes into `$(LOCAL_BIN)`, and it prevents them from being committed to the repo.

Makefile
```diff
@@ -40,12 +40,6 @@ REGISTRY ?= quay.io/open-cluster-management
 TAG ?= latest
 IMAGE_NAME_AND_VERSION ?= $(REGISTRY)/$(IMG)
 
-# Fix sed issues on mac by using GSED
-SED="sed"
-ifeq ($(GOOS), darwin)
-  SED="gsed"
-endif
-
```
JustinKuli, line 48: I'm surprised this would need to be removed... it might've been a while ago, but do you remember why this change was made?

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +65,201 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
+	watcher := depclient.ObjectIdentifier{
+		Group:     operatorv1.GroupVersion.Group,
+		Version:   operatorv1.GroupVersion.Version,
+		Kind:      "OperatorPolicy",
+		Namespace: req.Namespace,
+		Name:      req.Name,
+	}
+
+	// Start query batch for caching and watching related objects
+	err := r.DynamicWatcher.StartQueryBatch(watcher)
```
JustinKuli, line 77: It might be good to `defer` stopping the query batch, but that's a change we can consider in a future PR.

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +65,201 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
-		// Future implementation: Possibly emit a success event
-
-		return nil
 	}
 
-	// Object exists, now need to validate field to make sure they match
-	if exists {
-		OpLog.Info("The object already exists. Checking fields to verify matching specs")
-		// Future implementation: Verify the specs of the object matches the one on the cluster
+	r.setCompliance(policy, policyv1.Compliant)
```
JustinKuli, line 252: I think this is so that it could be marked as Compliant in inform mode, but actually if `createPolicyResources` is called in inform mode, then I think it should be marked as NonCompliant (which something above already does). tl;dr I think this line should be removed.

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +65,201 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
-) {
-	OpLog.Info("Waiting for leader election before periodically evaluating operator policies")
-	<-elected
+	// Check if specified Subscription exists
+	subscriptiongvk := schema.GroupVersionKind{Group: "operators.coreos.com", Version: "v1alpha1", Kind: "Subscription"}
 
-	exiting := false
-	for !exiting {
-		start := time.Now()
+	cachedItem, err := r.DynamicWatcher.Get(watcher, subscriptiongvk, policy.Namespace,
```
JustinKuli, line 107: I think the namespace should be `policy.Spec.Subscription.Namespace`. When created by an ocm policy, the operator policy namespace will just be the cluster namespace, and probably isn't where the OperatorGroup and Subscription will be defined.

I think there are other places with this same issue, please take look through all occurrences of `policy.Namespace`.

JustinKuli (changes requested):
Looking good overall! I think this will be a good base to work on once a few details here are ironed out

Makefile
zyjjay: I don't think Jeffrey and I made this change (unless it was by accident). I'll look into this and get back to you on that.

zyjjay:
@JustinKuli I made the suggested changes, let me know if there's any I missed!

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +65,205 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
-
-		return nil
+	err = r.DynamicWatcher.EndQueryBatch(watcher)
+	if err != nil {
+		panic(err)
 	}
 
-	return nil
+	// Will only reach this if in inform mode
+	return reconcile.Result{}, nil
```
zyjjay, line 263: It seems like there's an 'extra' reconcile that occurs after this line is called in inform mode. Not sure if I'm missing some details but would appreciate some insight on why this happens.

zyjjay: This also happens in enforce mode, but I'm assuming that's because the subscription object changes during the process of installing the operator.

JustinKuli: Two thoughts, one or both might be incorrect though 😆 

1. The controller watches for changes to the OperatorPolicy. If a reconcile loop causes a status change on the OperatorPolicy, that might trigger another reconcile immediately. In some controllers, we deliberately set things up so that status-only changes do not cause a new reconcile, but I don't think we've done that here.

2. The "first" reconcile is caused by the creation of the OperatorPolicy. During that reconcile, regardless of inform/enforce or anything else it does, it creates watches on the OperatorGroups and Subscriptions (ideally just the ones concerning this policy, but I think the current implementation is not that specific). When those watches are created, they sent an initial event that may cause `Reconcile` to be called.

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +65,205 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
+	operatorgroupgvk := schema.GroupVersionKind{Group: "operators.coreos.com", Version: "v1", Kind: "OperatorGroup"}
 
-		elapsed := time.Since(start).Seconds()
-		if float64(freq) > elapsed {
-			remainingSleep := float64(freq) - elapsed
-			sleepTime := time.Duration(remainingSleep) * time.Second
-			OpLog.V(2).Info("Sleeping before reprocessing the operator policies", "seconds", sleepTime)
-			time.Sleep(sleepTime)
-		}
+	cachedList, err := r.DynamicWatcher.List(watcher, operatorgroupgvk, policy.Spec.OperatorGroup.Namespace, nil)
```
JustinKuli, line 123: If the OperatorGroup is nil, this could have a problem.

Makefile
JustinKuli: I don't see the change for this in the current commit.

JustinKuli (commented):
Mostly I think this is now just waiting on test coverage.

Makefile
yiraeChristineKim: This is important. Plz don't remove. For mac this is important

controllers/operatorpolicy_controller.go
zyjjay: Would changing this to the subscription namespace be fine? Since that the CSV is installed the in same namespace as the subscription, maybe we can use the subscription namespace to check for existing operatorGroups.

Makefile
```diff
@@ -71,7 +71,7 @@ test-coverage: TESTARGS = -json -cover -covermode=atomic -coverprofile=coverage_
 test-coverage: test
 
 .PHONY: gosec-scan
-gosec-scan: GOSEC_ARGS = -exclude-generated
```
JustinKuli, line 74: Any clue why this was removed? If not, I say let's put it back.

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +65,215 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
+	OpLog.Info("Reconciling OperatorPolicy", "policy", policy.Name)
 
-// PeriodicallyExecOperatorPolicies loops through all operatorpolicies in the target namespace and triggers
-// template handling for each one. This function drives all the work the operator policy controller does.
-func (r *OperatorPolicyReconciler) PeriodicallyExecOperatorPolicies(freq uint, elected <-chan struct{},
-) {
-	OpLog.Info("Waiting for leader election before periodically evaluating operator policies")
-	<-elected
+	// Check if specified Subscription exists
+	subscriptiongvk := schema.GroupVersionKind{Group: "operators.coreos.com", Version: "v1alpha1", Kind: "Subscription"}
```
JustinKuli, line 110: Optional Nit: make this a constant and we could use it in a few places.

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +65,215 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
-					// handle policy
-					err := r.handleSinglePolicy(&policy)
-					if err != nil {
-						OpLog.Error(err, "Error while evaluating operator policy")
-					}
-				}
-			}
-		}
+	// Check if any OperatorGroups exist in the namespace
+	operatorgroupgvk := schema.GroupVersionKind{Group: "operators.coreos.com", Version: "v1", Kind: "OperatorGroup"}
```
JustinKuli, line 121: Optional Nit: same as above, this could be a constant and get used in a few places.

JustinKuli (changes requested):
My only required change is the `install.sh` file.

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +70,211 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
+	OpLog.Info("Reconciling OperatorPolicy", "policy", policy.Name)
 
-// PeriodicallyExecOperatorPolicies loops through all operatorpolicies in the target namespace and triggers
-// template handling for each one. This function drives all the work the operator policy controller does.
-func (r *OperatorPolicyReconciler) PeriodicallyExecOperatorPolicies(freq uint, elected <-chan struct{},
-) {
-	OpLog.Info("Waiting for leader election before periodically evaluating operator policies")
-	<-elected
+	// Check if specified Subscription exists
+	cachedItem, err := r.DynamicWatcher.Get(watcher, subscriptiongvk, policy.Spec.Subscription.Namespace,
```
mprahl, line 115: From what I can tell, `policy.Spec.Subscription` is not a required field:
https://github.com/open-cluster-management-io/config-policy-controller/blob/7549739f09cafec987a292aac32925718f8b1561/api/v1beta1/operatorpolicy_types.go#L107

Should it be? If not, then this code needs to guard against that not being set.

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +70,211 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
+		policy.Spec.Subscription.SubscriptionSpec.Package)
+	if err != nil {
+		panic(err)
```
mprahl, line 118: You shouldn't panic here since there could be a legitimate error and we don't want the whole propagator to crash.

I suggest just returning the error.

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +70,211 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
+		Version:   operatorv1.GroupVersion.Version,
+		Kind:      "OperatorPolicy",
+		Namespace: req.Namespace,
+		Name:      req.Name,
+	}
+
+	// Start query batch for caching and watching related objects
+	err := r.DynamicWatcher.StartQueryBatch(watcher)
+	if err != nil {
+		panic(err)
```
mprahl, line 84: Based on the current implementation, as long as you always call `r.DynamicWatcher.EndQueryBatch(watcher)` before returning, this is safe.

That being said, I don't think you should crash the controller in the event a newer version of the dependency watches libary can return a legitimate error. My suggestion is log and return the error so controller-runtime requeues the request.

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +70,211 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
-		}
+	ogExists := len(cachedList) != 0
+
+	// Exists indicates if a Subscription or Operatorgroup need to be created
+	exists := subExists && ogExists
+	shouldExist := strings.EqualFold(string(policy.Spec.ComplianceType), string(policyv1.MustHave))
+
+	// Case 1: policy has just been applied and related resources have yet to be created
+	if !exists && shouldExist {
+		return r.createPolicyResources(policy, watcher)
```
mprahl, line 142: Could you address @JustinKuli's comment about deferring `EndQueryBatch`? I find it hard to follow if the function starting the query batch isn't the one responsible for ending it.

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +70,211 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
-	err := r.Get(context.TODO(),
-		types.NamespacedName{Namespace: subscriptionSpec.Namespace, Name: subscriptionSpec.Name},
-		subscriptionSpec)
-	exists := !errors.IsNotFound(err)
-	shouldExist := strings.EqualFold(string(policy.Spec.ComplianceType), string(policyv1.MustHave))
+	ogList, err := r.DynamicWatcher.ListFromCache(
+		operatorgroupgvk,
+		ogNamespace, nil)
+	if err != nil {
+		panic(err)
```
mprahl, line 188: I don't think it's safe to panic in this case. I'd suggest returning the error.

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +70,211 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
+			err = r.Create(context.TODO(), ogSpec)
 			if err != nil {
+				OpLog.Error(err, "Error while creating OperatorGroup")
 				r.setCompliance(policy, policyv1.NonCompliant)
-				OpLog.Error(err, "Could not handle missing musthave object")
 
-				return err
+				err = r.DynamicWatcher.EndQueryBatch(watcher)
+				if err != nil {
+					panic(err)
```
mprahl, line 203: I don't think it's safe to panic in this case. I'd suggest returning the error.

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +70,211 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
 			if err != nil {
-				r.setCompliance(policy, policyv1.NonCompliant)
-				OpLog.Error(err, "Could not handle missing musthave object")
-
-				return err
+				panic(err)
 			}
 
-			r.setCompliance(policy, policyv1.Compliant)
+			return reconcile.Result{Requeue: true}, err
```
mprahl, line 217: Why do you have to requeue in this case?

controllers/operatorpolicy_controller.go
JustinKuli: I was figuring we would replace most (if not all) panics as we proceeded with implementation.

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +70,211 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
+			r.setCompliance(policy, policyv1.NonCompliant)
 		}
+	}
 
-		// Inform
-		r.setCompliance(policy, policyv1.NonCompliant)
-
-		return nil
+	// Retrieve subscription from cache
+	cachedSubscription, err := r.DynamicWatcher.GetFromCache(
```
mprahl, line 229: Why not pass this as an argument to this function since you already retrieved it from the cache?

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +70,211 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
-	if exists && !shouldExist {
+	// Create new subscription
+	if cachedSubscription == nil {
 		if strings.EqualFold(string(policy.Spec.RemediationAction), string(policyv1.Enforce)) {
-			OpLog.Info("deleting kind " + subscriptionSpec.Kind + " in ns " + subscriptionSpec.Namespace)
-			err = r.Delete(context.TODO(), subscriptionSpec)
+			subscriptionSpec := buildSubscription(policy)
+			OpLog.Info("creating kind " + subscriptionSpec.Kind + " in ns " + subscriptionSpec.Namespace)
 
+			err := r.Create(context.TODO(), subscriptionSpec)
```
mprahl, line 242: You should be using the context from the `Reconcile` method here and all other API calls in this method.

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +70,211 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
+				err = r.DynamicWatcher.EndQueryBatch(watcher)
+				if err != nil {
+					panic(err)
+				}
 
-				return err
+				return reconcile.Result{}, err
 			}
 
 			r.setCompliance(policy, policyv1.Compliant)
```
mprahl, line 255: Is this actually compliant? Don't you need to wait for the status checks to confirm it's compliant?

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +70,211 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
 
-			return nil
-		}
+			err = r.DynamicWatcher.EndQueryBatch(watcher)
+			if err != nil {
+				panic(err)
+			}
 
-		// Inform
+			return reconcile.Result{Requeue: true}, nil
```
mprahl, line 262: Why are you requeuing in this case?

controllers/operatorpolicy_controller.go
```diff
@@ -300,16 +333,16 @@ func buildOperatorGroup(
 ) *operatorv1.OperatorGroup {
 	operatorGroup := new(operatorv1.OperatorGroup)
 
-	gvk := schema.GroupVersionKind{
-		Group:   "operators.coreos.com",
-		Version: "v1",
-		Kind:    "OperatorGroup",
+	if policy.Spec.OperatorGroup == nil {
+		operatorGroup.SetGroupVersionKind(operatorgroupgvk)
+		operatorGroup.ObjectMeta.SetName(policy.Name + "-default-og")
```
mprahl, line 338: You might run into the 63 character limit if the policy name is long. I'd suggest guarding against this.

controllers/operatorpolicy_controller.go
```diff
@@ -300,16 +333,16 @@ func buildOperatorGroup(
 ) *operatorv1.OperatorGroup {
 	operatorGroup := new(operatorv1.OperatorGroup)
 
-	gvk := schema.GroupVersionKind{
-		Group:   "operators.coreos.com",
-		Version: "v1",
-		Kind:    "OperatorGroup",
+	if policy.Spec.OperatorGroup == nil {
+		operatorGroup.SetGroupVersionKind(operatorgroupgvk)
```
mprahl, line 337: Please use camel case for the `operatorgroupgvk` variable (e.g. `operatorGroupGVK`).

controllers/operatorpolicy_controller.go
```diff
@@ -26,18 +28,26 @@ const (
 	OperatorControllerName string = "operator-policy-controller"
 )
 
+var (
+	subscriptiongvk  = schema.GroupVersionKind{Group: "operators.coreos.com", Version: "v1", Kind: "OperatorGroup"}
+	operatorgroupgvk = schema.GroupVersionKind{Group: "operators.coreos.com", Version: "v1", Kind: "OperatorGroup"}
```
mprahl, line 33:
```suggestion
	subscriptionGVK  = schema.GroupVersionKind{Group: "operators.coreos.com", Version: "v1", Kind: "OperatorGroup"}
	operatorGroupGVK = schema.GroupVersionKind{Group: "operators.coreos.com", Version: "v1", Kind: "OperatorGroup"}
```

controllers/operatorpolicy_controller.go
mprahl: I think the first thought is correct because they are setting `DisableInitialReconcile: true` on the dependency watcher.

You should be able to use a predicate to ignore status updates. I think controller-runtime has a built-in one for just checking if the "generation" changes (i.e. `spec` has changed).

controllers/operatorpolicy_controller.go
mprahl: I don't want to risk forgetting to replace these.

controllers/operatorpolicy_controller.go
JeffeyL: @zyjjay After pulling changes I noticed that these are duplicates as well, subscription should be `schema.GroupVersionKind{Group: "operators.coreos.com", Version: "v1alpha1", Kind: "Subscription"}`

controllers/operatorpolicy_controller.go
zyjjay: That's an good point. I think you mentioned the capability of the controller to be able to handle operator installations through means other than the OperatorPolicy. In that case, and in the general case, I think the subscription field must be set to ensure the subscription is either being created, and/or watched. I'll make this change.

controllers/operatorpolicy_controller.go
zyjjay: So the pattern Justin, Jeffrey and I agreed upon is to create one resource at a time and return early (and requeue) so that each reconcile is more cleanly divided into different logical chunks and it's purpose is clear. For example: First reconcile is to create the OperatorGroup, second one for Subscription.

Prior to this, the controller would try to create all the resources in one reconcile.

controllers/operatorpolicy_controller.go
zyjjay: After a bit of thinking about the flow, I don't actually think requeueing is necessary.

controllers/operatorpolicy_controller.go
zyjjay: Currently, as long as the controller is able to create the necessary resources specified in the OperatorPolicy, it will be considered 'Complaint'. In future PRs, specifically after adding the health checks for the deployments, the compliance state of the policy should be more definitive and better represented with respect to the state of the resources it will watch. Hope that makes sense!

controllers/operatorpolicy_controller.go
zyjjay: Would enforcing a specific length (i.e., enforce the policy name to be 63 - length of default OperatorGroup suffix) be a good approach? Another way I'm thinking of is dynamically renaming the default OperatorGroup but that might make it difficult to look for it on the cluster.

controllers/operatorpolicy_controller.go
mprahl: Could you explain why this pattern was chosen?

I'm thinking that it will not be efficient and it doesn't really add to be better readability/maintainability in my opinion. Why I say it's not efficient is that the first reconcile creates the operator group. Then after the requeue, it'll check to see if the operator group matches, and so on and so forth.

controllers/operatorpolicy_controller.go
mprahl: Sure, that's fine. Could you please add a comment in the code to that effect?

controllers/operatorpolicy_controller.go
mprahl: Could you reuse the subscription name since that is a required field?

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +70,211 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
-
-					// handle policy
-					err := r.handleSinglePolicy(&policy)
-					if err != nil {
-						OpLog.Error(err, "Error while evaluating operator policy")
-					}
-				}
-			}
-		}
+	cachedList, err := r.DynamicWatcher.List(watcher, operatorgroupgvk, ogNamespace, nil)
```
mprahl, line 129: Is it common to have multiple operator groups in the same namespace? I'm wondering if this will cause a lot of extra reconciles since you are using the dynamic watch library.

controllers/operatorpolicy_controller.go
```diff
@@ -60,166 +70,211 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
 
-			// Currently creates an OperatorGroup for every Subscription
-			// in the same ns, and defaults to targeting all ns.
-			// Future implementations will enable targeting ns based on
-			// installModes supported by the CSV. Also, only one OperatorGroup
-			// should exist in each ns
-			operatorGroup := buildOperatorGroup(policy)
-			err = r.Create(context.TODO(), operatorGroup)
+			// Created successfully, requeue result
+			r.setCompliance(policy, policyv1.NonCompliant)
```
mprahl, line 210: Why is this non compliant if it was successful?

controllers/operatorpolicy_controller.go
```diff
@@ -300,16 +333,16 @@ func buildOperatorGroup(
...
+		operatorGroup.ObjectMeta.SetNamespace(policy.Spec.OperatorGroup.Namespace)
+		operatorGroup.Spec.TargetNamespaces = policy.Spec.OperatorGroup.Target.Namespace
 	}
 
-	operatorGroup.SetGroupVersionKind(gvk)
-	operatorGroup.ObjectMeta.SetName(policy.Spec.Subscription.Package + "-operator-group")
-	operatorGroup.ObjectMeta.SetNamespace(policy.Spec.Subscription.Namespace)
-	operatorGroup.Spec.TargetNamespaces = []string{"*"}
-
 	return operatorGroup
```
mprahl, line 347: @JustinKuli , do you think it'd be worthwhile to add a label indicating ownership by the policy? I'm thinking it might help future reconciles to find the operator group. The same for the subscription.

controllers/operatorpolicy_controller.go
zyjjay: I wouldn't say it's common but it is possible to have multiple operator groups. However, having multiple operator groups in the same ns will transition any CSV within that ns to a failure state with the reason `TooManyOperatorGroups`. Using a `List` guards against the case where there is a preexisting operator group on the cluster (either created manually or by another operator policy) and you want to apply another operator policy that has a different operator group spec in the same ns. Since we don't know the name of other operator groups except the one we're currently reconciling, we just check that resource type within the ns.

controllers/operatorpolicy_controller.go
zyjjay: At this point the subscription has not been created yet, so the `NonCompliant` is like a placeholder until the subscription is created. My logic is that the policy should be `NonCompliant` until every object in the spec is created.

controllers/operatorpolicy_controller.go
JustinKuli: I suggested this pattern. I had a dream of what I wanted it to look like, and I still believe the idea would help with modularization and readability, but I concede it would be less efficient than an alternative.

I was hoping the main reconcile could simplify down to something like this:
```golang
opGroup, ogNeedsAction := r.CheckOperatorGroup(policy)
if ogNeedsAction {
    return r.HandleOperatorGroup(policy, opGroup)
}

sub, subNeedsAction := r.CheckSubscription(policy)
if subNeedsAction {
    return r.HandleSubscription(policy, sub)
}

installPlan, ipNeedsAction := r.CheckInstallPlan(policy, sub)
if ipNeedsAction {
    return r.HandleInstallPlan(policy, installPlan, sub)
}
```

If each `Check` function could be read-only API calls, and therefore be likely to just use the cache, they would be quite fast/cheap and I wouldn't mind running them potentially many times due to repeated reconciles.

Then the `Handle` functions would have some write step, creating or updating a resource, possibly just the updating policy's status. I feel like there are a lot of different possible outcomes of the `Handle` functions, with some outcomes blocking future work, and figuring out how to handle all of the outcomes would become a maze of logic. I feel it is simpler to just restart at that point, and if the actions were correctly taken (and have "stuck" on the cluster) than the next corresponding `Check` will report that no further action is needed.

Additionally, since the `Handle` functions do some API calls, they would take more time, and things fetched from previous steps might no longer be up to date. Bugs, or at least possible update errors, could then appear. Restarting, and just doing "fast" things until another action needs to be taken might avoid those.

mprahl: I think you'll wind up having to check everything anyways in the inform case since you probably want to provide an exhaustive list of the non-compliant "related objects", so then the code flows would be very different for inform/enforce.

I'm also not quite seeing the complexity that this simplifies. I just find it to be an unusual pattern and don't think it'd be hard to write this without the requeuing.

If you want to handle that in a separate PR to unblock this one, that'd be fine.

controllers/operatorpolicy_controller.go
mprahl: As mentioned in Slack, this wasn't addressed. You either need to move this after retrieving the operator policy or move the defer up since it looks like in the unknown error case, you currently aren't ending the query batch which will cause an issue on the next reconcile.

controllers/operatorpolicy_controller.go
```diff
@@ -60,180 +78,194 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
-	if !exists && shouldExist {
-		if strings.EqualFold(string(policy.Spec.RemediationAction), string(policyv1.Enforce)) {
-			OpLog.Info("creating kind " + subscriptionSpec.Kind + " in ns " + subscriptionSpec.Namespace)
-			subscriptionSpec := buildSubscription(policy, subscriptionSpec)
-			err = r.Create(context.TODO(), subscriptionSpec)
+	// Check if specified Subscription exists
+	cachedSubscription, err := r.DynamicWatcher.Get(watcher, subscriptionGVK, policy.Spec.Subscription.Namespace,
+		policy.Spec.Subscription.SubscriptionSpec.Package)
+	if err != nil {
+		OpLog.Error(err, "Could not get Subscription "+policy.Spec.Subscription.SubscriptionSpec.Package+
```
mprahl, line 135: Like my other comment, could you please make the log statement a key-value pair format?

controllers/operatorpolicy_controller.go
```diff
@@ -60,180 +78,189 @@ var _ reconcile.Reconciler = &OperatorPolicyReconciler{}
...
 		if errors.IsNotFound(err) {
 			OpLog.Info("Operator policy could not be found")
 
-			return reconcile.Result{}, nil
+			err = r.DynamicWatcher.RemoveWatcher(watcher)
+			if err != nil {
+				OpLog.Error(err, "Error updating dependency watcher. Ignoring the failure.")
+			}
+
+			return reconcile.Result{}, err
```
mprahl, line 100:
```suggestion
			return reconcile.Result{}, nil
```

controllers/operatorpolicy_controller.go
```diff
@@ -243,51 +270,31 @@ func (r *OperatorPolicyReconciler) updatePolicyStatus(
...
-		Kind:    "Subscription",
-	}
+	subscription := new(operatorv1alpha1.Subscription)
 
-	subscription.SetGroupVersionKind(gvk)
+	subscription.SetGroupVersionKind(subscriptionGVK)
 	subscription.ObjectMeta.Name = policy.Spec.Subscription.Package
 	subscription.ObjectMeta.Namespace = policy.Spec.Subscription.Namespace
 	subscription.Spec = policy.Spec.Subscription.SubscriptionSpec.DeepCopy()
+	OpLog.Info("Creating kind " + subscription.Kind + " in ns " + subscription.Namespace)
```
mprahl, line 284: Like my other comment, could you please make these key-value pairs?

controllers/operatorpolicy_controller.go
```diff
@@ -300,16 +307,22 @@ func buildOperatorGroup(
...
 
-	gvk := schema.GroupVersionKind{
-		Group:   "operators.coreos.com",
-		Version: "v1",
-		Kind:    "OperatorGroup",
+	// Create a default OperatorGroup if one wasn't specified in the policy
+	// Future work: Prevent creating multiple OperatorGroups in the same ns
+	if policy.Spec.OperatorGroup == nil {
+		operatorGroup.SetGroupVersionKind(operatorGroupGVK)
+		operatorGroup.ObjectMeta.SetName(policy.Spec.Subscription.Package + defaultOGSuffix)
```
mprahl, line 314: After talking with Justin, the name length maximum is 253 characters but the label values are 63. So I was mistaken.

I think you could possibly run into this but as long as you're logging the error, I think this is fine for now.

controllers/operatorpolicy_controller.go
```diff
@@ -300,16 +307,22 @@ func buildOperatorGroup(
...
+		operatorGroup.ObjectMeta.SetName(policy.Spec.OperatorGroup.Name)
+		operatorGroup.ObjectMeta.SetNamespace(policy.Spec.OperatorGroup.Namespace)
+		operatorGroup.Spec.TargetNamespaces = policy.Spec.OperatorGroup.Target.Namespace
 	}
 
-	operatorGroup.SetGroupVersionKind(gvk)
-	operatorGroup.ObjectMeta.SetName(policy.Spec.Subscription.Package + "-operator-group")
-	operatorGroup.ObjectMeta.SetNamespace(policy.Spec.Subscription.Namespace)
-	operatorGroup.Spec.TargetNamespaces = []string{"*"}
+	OpLog.Info("Creating kind " + operatorGroup.Kind +
```
mprahl, line 324: Could you please use key-value pairs in the log?

controllers/operatorpolicy_controller.go
JustinKuli: I think I'd prefer recording this information in the status of the OperatorPolicy. We'd already basically want that for RelatedObjects, I think.

In `enforce` mode, it would certainly be ok to add a label like that. But we'd have to consider if that kind of "write" operation would be ok in `inform` mode - and I think there would need to be similar considerations for `mustnothave` mode. If in any of those situations we *don't* want to add the label, then each situation might need to be handled differently... whereas if we always just record it in the status, and use it from the status, it's all somewhat consistent.

mprahl: That's fair. Since we will have a finalizer preventing the OperatorPolicy from being deleted and thus presumably, the status from being lost, I'm good with this.

mprahl (approved):
/hold for Justin to review

JustinKuli (approved):
LGTM
/unhold

PR #190, opened 2024-01-19
Use `OLM_VERSION`

dhaiducek:
...also test `magic-mirror` syncing now that the large file is (hopefully) removed from the commit history.


gparvin (approved):
Thanks!

PR #191, opened 2024-01-22
Add diff logging

dhaiducek:
Adds a `recordDiff` enum parameter to the `ConfigurationPolicy`. When set to `Log`, it uses the `go-difflib` package to compare the YAML marshalled into strings, with unified output similar to the `diff` command in the shell. While `go-difflib` is unmaintained, it's extensively imported, in particular by the `stretchr/testify` package here.

For simplicity, the diff for objectDefinitions without a name specified are not logged.

Further enhancements could be made by making the diff object-aware. For instance, the ArgoCD diff goes so far as to hack around the Kubernetes library that implements `apply`. However, I think this is a good, simple implementation as far as an MVP goes and is likely more performant than an object-aware implementation.

ref: https://issues.redhat.com/browse/ACM-9072

### DIFF
--- a/api/v1/configurationpolicy_types.go
+++ b/api/v1/configurationpolicy_types.go
@@ -187,7 +187,19 @@ type ObjectTemplate struct {
 	// ObjectDefinition defines required fields for the object
 	// +kubebuilder:pruning:PreserveUnknownFields
 	ObjectDefinition runtime.RawExtension `json:"objectDefinition"`
+
+	// RecordDiff specifies whether (and where) to log the diff between the object on the
+	// cluster and the objectDefinition in the policy. Defaults to "None".
+	RecordDiff RecordDiff `json:"recordDiff,omitempty"`
 }
+
+// +kubebuilder:validation:Enum=Log;None
+type RecordDiff string
+
+const (
+	RecordDiffLog  RecordDiff = "Log"
+	RecordDiffNone RecordDiff = "None"
+)
 
 // ConfigurationPolicyStatus defines the observed state of ConfigurationPolicy
 type ConfigurationPolicyStatus struct {
@@ -210,9 +222,6 @@ type CompliancePerClusterStatus struct {
 
 // ComplianceMap map to hold CompliancePerClusterStatus objects
 type ComplianceMap map[string]*CompliancePerClusterStatus
-
-// ResourceState genric description of a state
-type ResourceState string
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -1775,11 +1775,8 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 
 			throwSpecViolation = !compliant
 		} else {
-			compType := strings.ToLower(string(objectT.ComplianceType))
-			mdCompType := strings.ToLower(string(objectT.MetadataComplianceType))
-
 			throwSpecViolation, msg, triedUpdate, updatedObj = r.checkAndUpdateResource(
-				obj, compType, mdCompType, remediation,
+				obj, objectT, remediation,
 			)
 		}
 
@@ -2549,10 +2546,12 @@ type cachedEvaluationResult struct {
 // successfully.
 func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
 	obj singleObject,
-	complianceType string,
-	mdComplianceType string,
+	objectT *policyv1.ObjectTemplate,
 	remediation policyv1.RemediationAction,
 ) (throwSpecViolation bool, message string, updateNeeded bool, updateSucceeded bool) {
+	complianceType := strings.ToLower(string(objectT.ComplianceType))
+	mdComplianceType := strings.ToLower(string(objectT.MetadataComplianceType))
+
 	log := log.WithValues(
 		"policy", obj.policy.Name, "name", obj.name, "namespace", obj.namespace, "resource", obj.gvr.Resource,
 	)
@@ -2633,16 +2632,6 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
 		}
 
 		if keyUpdateNeeded {
-			isInform := strings.EqualFold(string(remediation), string(policyv1.Inform))
-
-			// If a key didn't match but the cluster supports dry run mode, then continue merging the object
-			// and then run a dry run update request to see if the Kubernetes API agrees with the assesment.
-			if isInform && !r.DryRunSupported {
-				r.setEvaluatedObject(obj.policy, obj.existingObj, false)
-
-				return true, "", false, false
-			}
-
 			if isStatus {
 				throwSpecViolation = true
 				statusMismatch = true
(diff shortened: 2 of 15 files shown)


dhaiducek:
Particular special thanks to @JustinKuli for the initial legwork on this PR. 😄

controllers/configurationpolicy_utils_test.go
```diff
@@ -197,3 +198,113 @@ func TestEqualObjWithSortEmptyMap(t *testing.T) {
...
+			existingObj := &unstructured.Unstructured{
+				Object: test.existingObj,
+			}
+			updatedObj := &unstructured.Unstructured{
+				Object: test.updatedObj,
+			}
+
+			diff, err := generateDiff(existingObj, updatedObj)
+			if err != nil {
+				assert.EqualError(t, err, test.expectedErr)
```
gparvin, line 298: Is there a way to test the error cases?

gparvin:
This looks good. I only had one minor comment.  Thanks Dale!

controllers/configurationpolicy_utils_test.go
dhaiducek: I'll take a look. I don't think I can hit all of them, but I might be able to pass it a `nil` somewhere...

dhaiducek: I haven't worked out a way to hit the errors. I think the likelihood that the `generateDiff` function is passed a `nil` pointer is minimal/nonexistent (and would even require another check). Beyond that, it seems `yaml.Marshal()` will always (?) work on a `map[string]interface{}` Go object, and I think the `go-difflib` would likely only fail with system errors if the buffer fails, which also seems unlikely.

Unless anyone has other ideas on how to hit them, I'll remove this assertion and use `Fatal()` instead.

api/v1/configurationpolicy_types.go
```diff
@@ -187,8 +187,21 @@ type ObjectTemplate struct {
 	// ObjectDefinition defines required fields for the object
 	// +kubebuilder:pruning:PreserveUnknownFields
 	ObjectDefinition runtime.RawExtension `json:"objectDefinition"`
+
+	// RecordDiff specifies whether (and where) to log the diff between the object on the
+	// cluster and the objectDefinition in the policy. Defaults to "None".
+	// +kubebuilder:default:=None
+	RecordDiff RecordDiff `json:"recordDiff,omitempty"`
```
mprahl, line 194: Does `omitempty` apply here? I think it'd have to be a pointer for that to work since the default is set to `None`.

mprahl: I think we probably want to remove `// +kubebuilder:default:=None` and treat an empty string or this set to `None` as equivalent in the code.

controllers/configurationpolicy_controller.go
```diff
@@ -2725,6 +2724,16 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
 				return false, "", false, false
 			}
 
+			// Generate and log the diff
+			if objectT.RecordDiff == policyv1.RecordDiffLog {
```
mprahl, line 2728: By having this here, I guess inform on OCP 3.11 will not log the diff since DryRun is not supported on OCP 3.11.

dhaiducek: That sounds like acceptable risk to me?

dhaiducek: ...or should make a best effort to log something?

test/e2e/case39_diff_generation_test.go
```diff
@@ -0,0 +1,87 @@
...
+				configPolicyName, testNamespace, true, defaultTimeoutSeconds)
+
+			return utils.GetStatusMessage(managedPlc)
+		}, 30, 0.5).Should(Equal("configmaps [case39-map] was updated successfully in namespace default"))
+	})
+
+	It("diff should be logged by the controller", func() {
+		By("Checking the controller logs")
+		Eventually(func(g Gomega) string {
+			diff, err := exec.Command("awk", "/Logging the diff:/,/^\t{/", logPath).Output()
```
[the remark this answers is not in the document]
dhaiducek: Sure--I'll work out how to parse it with Go.

controllers/configurationpolicy_controller.go
mprahl: So in OCP 3.11, we return right after the first mismatch if it's inform:
https://github.com/open-cluster-management-io/config-policy-controller/blob/77048df3ca61b976d61f75ec9b1b93b48400c168/controllers/configurationpolicy_controller.go#L2638-L2644

I'm fine with not supporting OCP 3.11 if we document that. If you want to support it, you basically need to remove the linked code and then add `else if isInform` to the `if r.DryRunSupported {` statement and log the diff and return.

mprahl (dismissed):
/hold depending on if you want to support OCP 3.11

controllers/configurationpolicy_controller.go
dhaiducek: > ...and log the diff and return.

But what's the diff? The diff between the `objectDefinition` and the object on the cluster? Seems there'd need to be some extra logic there to only diff the fields that are defined if it's `musthave`, right?

mprahl: I think `checkAndUpdateResource` is only called if it's musthave or mustonlyhave and the object exists.

JustinKuli (dismissed):
LGTM! (With the understanding that it won't work on managed clusters running OCP 3.11)

controllers/configurationpolicy_controller.go
```diff
@@ -2653,6 +2644,8 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
 
 				if !isInform {
 					log.Info("Queuing an update for the object due to a value mismatch", "key", key)
+				} else {
+					log.Info("Detected value mismatch for object key: " + key)
```
dhaiducek, line 2648: This seemed like a nice little add to the logging since we're already doing it for `enforce`, but I can take it out (or move it to level 2) if we think it'd be too noisy.

controllers/configurationpolicy_controller.go
```diff
@@ -2653,6 +2644,8 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
 
 				if !isInform {
```
mprahl, line 2645: As @dhaiducek pointed out, this only shows the top-level key that didn't match so it's not very useful. So these log messages can be removed entirely if there is a log message that says the update is taking place.

controllers/configurationpolicy_controller.go
```diff
@@ -2725,14 +2718,36 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
...
+			// Generate and log the diff
+			if objectT.RecordDiff == policyv1.RecordDiffLog {
+				diff, err := generateDiff(existingObjectCopy, dryRunUpdatedObj)
+				if err != nil {
+					log.Info("Failed to generate the diff: " + err.Error())
+				} else {
+					log.Info("Logging the diff:\n" + diff)
+				}
+			}
+		} else if isInform && objectT.RecordDiff == policyv1.RecordDiffLog {
```
mprahl, line 2730:
```suggestion
		} else if objectT.RecordDiff == policyv1.RecordDiffLog {
```

mprahl, line 2730: As discussed in Slack, it'd be nice to have an info log message that says the diff could be shown if the new recordDiff field is set. I can't think of a nice way to specify it but maybe `spec.object-templates[*].recordDiff`  or something?

dhaiducek: I like that path spec. Updated!

controllers/configurationpolicy_controller.go
```diff
@@ -2651,16 +2640,18 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
 			} else {
 				updateNeeded = true
 
-				if !isInform {
-					log.Info("Queuing an update for the object due to a value mismatch", "key", key)
+				mismatchLog := "Detected value mismatch for object key: " + key
+				// Add a configuration breadcrumb for users that might be looking in the logs for a diff
+				if objectT.RecordDiff != policyv1.RecordDiffLog {
+					mismatchLog += " (Diff disabled. To log the diff, " +
+						"set 'spec.object-tempates[].objectDefinition.recordDiff' to 'Log' for this object-template.)"
```
mprahl, line 2647:
```suggestion
						"set 'spec.object-tempates[].recordDiff' to 'Log' for this object-template.)"
```

mprahl (approved):
/hold for fixing the typo in the log message

controllers/configurationpolicy_controller.go
dhaiducek: Oops--Thanks for catching that!

mprahl (approved):
/unhold

PR #192, opened 2024-01-25
Handle preexisting operator

JustinKuli:
For https://issues.redhat.com/browse/ACM-9283.

In addition to handling more situations around whether the Subscription or OperatorGroup already exist, this adds condition reporting, related objects, and compliance events.

This PR also slims the CRD to what we believe can be functional in an initial release.


controllers/configurationpolicy_controller.go
```diff
@@ -1741,7 +1741,7 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 
 	if exists && !obj.shouldExist {
 		// it is a mustnothave but it exist, so it must be deleted
-		if strings.EqualFold(string(remediation), string(policyv1.Enforce)) {
+		if remediation.IsEnforce() {
```
mprahl, line 1744: @JustinKuli is the plan to keep mustnothave to remove what the operator created but nothing beyond that or is the plant to still remove mustnothave until the whole uninstall logic is added?

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	desiredOpGroup, err := buildOperatorGroup(policy)
 	if err != nil {
-		OpLog.Error(err, "Could not list operator group", "kind", operatorGroupGVK.Kind,
-			"namespace", ogNamespace)
+		return fmt.Errorf("error building operator group: %w", err)
+	}
 
-		return reconcile.Result{}, err
+	foundOpGroups, err := r.DynamicWatcher.List(
+		watcher, operatorGroupGVK, desiredOpGroup.Namespace, labels.Everything())
```
mprahl, line 145: Is `labels.Everything()` basically no selector? If so, you can optionally provide `nil` instead for the label selector.

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+					// The names match, but the specs don't: report NonCompliance
+					err := r.updateStatus(ctx, policy, mismatchCond("OperatorGroup"), mismatchedObj(&opGroup))
+					if err != nil {
+						return fmt.Errorf("error updating the status for an OperatorGroup that does not match: %w", err)
+					}
+
+					if policy.Spec.RemediationAction.IsEnforce() {
+						desiredOpGroup.ResourceVersion = opGroup.GetResourceVersion()
+
+						err := r.Update(ctx, desiredOpGroup)
```
mprahl, line 208: This seems like "must only have" since I don't see any checking to see if the desired operator group is a subset of the existing operator group.

mprahl, line 208: Same comment as above, please ensure field validation is strict on the update.

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			// For an initial implementation, assume if an OperatorGroup already exists, then it's a good one.
+			// Only update the condition if it reflects a NonCompliant state, to prevent repeats.
+			idx, existingCond := policy.Status.GetCondition(opGroupConditionType)
 
-	// Case 4: Resources exist, and should exist (i.e. update)
-	if exists && shouldExist {
-		// Future implementation: Verify the specs of the object matches the one on the cluster
-		OpLog.Info("The object already exists. Checking fields to verify matching specs")
+			if idx == -1 || existingCond.Status == metav1.ConditionFalse {
+				err := r.updateStatus(ctx, policy, matchesCond("OperatorGroup"), matchedObj(&opGroup))
```
mprahl, line 243: Would it be possible to include in a message that there is a preexisting operator group that was not verified since it was not specified in the policy?

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 
-			// Created successfully, requeue result
-			r.setCompliance(ctx, policy, policyv1.NonCompliant)
+	sub := make(map[string]interface{})
 
-			return reconcile.Result{Requeue: true}, err
-		} else if policy.Spec.OperatorGroup != nil {
-			// If inform mode, keep going and return at the end without requeue. Before
-			// continuing, should check if required og spec is created to set compliance state
+	err := json.Unmarshal(policy.Spec.Subscription.Raw, &sub)
```
mprahl, line 273: I suggest this instead so you can use things like `sub.GetNamespace()`.
```go
sub := unstructured.Unstructured{}

err := sub.UnmarshalJSON(policy.Spec.Subscription.Raw)
```

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	if err != nil {
+		return nil, fmt.Errorf("error unmarshalling subscription: %w", err)
+	}
 
-			// Set to non compliant because operatorgroup does not exist on cluster, but
-			// should still go to check subscription
-			r.setCompliance(ctx, policy, policyv1.NonCompliant)
-		}
+	subNamespace, ok := sub["namespace"].(string)
+	if !ok {
```
mprahl, line 279: Do you also want to check the length too?

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-			err := r.Create(ctx, subscriptionSpec)
-			if err != nil {
-				OpLog.Error(err, "Could not handle missing musthave object")
-				r.setCompliance(ctx, policy, policyv1.NonCompliant)
+		return operatorGroup, nil
+	}
 
-				return reconcile.Result{}, err
-			}
+	opGroup := make(map[string]interface{})
```
mprahl, line 292: Same comment as for the subscription. I recommend using unstructured. It's the same amount of code and you don't have to type assert.

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	}
 
-			return reconcile.Result{}, nil
-		}
-		// inform
-		r.setCompliance(ctx, policy, policyv1.NonCompliant)
+	// Fallback to the Subscription namespace if the OperatorGroup namespace is not specified in the policy.
+	ogNamespace := subNamespace
+
+	if specifiedNS, ok := opGroup["namespace"].(string); ok {
```
mprahl, line 302:
```suggestion
	if specifiedNS, ok := opGroup["namespace"].(string); ok || specifiedNS == "" {
```

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	// Will only reach this if in inform mode
-	return reconcile.Result{}, nil
+	name, ok := opGroup["name"].(string)
+	if !ok {
+		return nil, fmt.Errorf("name is required in operatorGroup.spec")
+	}
+
+	spec := new(operatorv1.OperatorGroupSpec)
+
+	err = json.Unmarshal(policy.Spec.OperatorGroup.Raw, spec)
```
mprahl, line 313: If you follow my other comments, you can use unstructured for the conversion but it does create an extra json.Marshal that your implementation does not.
```
err = k8sruntime.DefaultUnstructuredConverter.FromUnstructured(opgroup.UnstructuredContent(), &spec)
```

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	}
 
-				return reconcile.Result{}, err
-			}
+	opGroup := make(map[string]interface{})
 
-			// Future work: Check availability/status of all resources
-			// managed by the OperatorPolicy before setting compliance state
-			r.setCompliance(ctx, policy, policyv1.Compliant)
+	err = json.Unmarshal(policy.Spec.OperatorGroup.Raw, &opGroup)
```
mprahl, line 294: Does this fail if you directly unmarshal it to the `operatorv1.OperatorGroupSpec`? If not, it'd allow you to skip a step.

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	} else {
-		policy.Status = updatedStatus
+		return fmt.Errorf("error building subscription: %w", err)
 	}
 
-	err = r.Status().Update(ctx, policy)
+	foundSub, err := r.DynamicWatcher.Get(watcher, subscriptionGVK, desiredSub.Namespace, desiredSub.Name)
 	if err != nil {
-		OpLog.Info(fmt.Sprintf("Failed to update policy status: %s", err))
+		return fmt.Errorf("error listing OperatorGroups: %w", err)
```
mprahl, line 335:
```suggestion
		return fmt.Errorf("error getting the subscription: %w", err)
```

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+	if foundSub == nil {
+		// Missing Subscription: report NonCompliance
+		err := r.updateStatus(ctx, policy, missingWantedCond("Subscription"), missingWantedObj(desiredSub))
+		if err != nil {
+			return fmt.Errorf("error updating status for a missing Subscription: %w", err)
+		}
+
+		if policy.Spec.RemediationAction.IsEnforce() {
+			err := r.Create(ctx, desiredSub)
```
mprahl, line 346: Same comment about field validation.

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+			desiredSub.SetGroupVersionKind(subscriptionGVK) // Create stripped this information
+
+			// Now it should match, so report Compliance
+			err = r.updateStatus(ctx, policy, createdCond("Subscription"), createdObj(desiredSub))
+			if err != nil {
+				return fmt.Errorf("error updating the status for a created Subscription: %w", err)
+			}
+		}
+	} else {
```
mprahl, line 359: Optional: You could return nil at the end of the if block to prevent the indentation of else.

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		// Subscription found; check if specs match
+		desiredUnstruct, err := runtime.DefaultUnstructuredConverter.ToUnstructured(desiredSub)
+		if err != nil {
+			return fmt.Errorf("error converting desired Subscription to an Unstructured: %w", err)
+		}
+
+		if reflect.DeepEqual(desiredUnstruct["spec"], foundSub.Object["spec"]) {
```
mprahl, line 366: This seems like a "must only have" check. Is the intention to only support "must only have"?

mprahl, line 366: Also, are there any `omitempty` that the user could specify in `desiredSub` but not be returned in `foundSub`?

mprahl: Since we are dealing with unstructured here, I think you could move out some of the config-policy-controller code from `checkAndUpdateResource` to do the comparison for both must have and must only have.

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			// FUTURE: Check more details about the *status* of the Subscription
+			// For now, (conditionally) mark it as compliant
+			idx, existingCond := policy.Status.GetCondition(subConditionType)
+
+			if idx == -1 || existingCond.Status == metav1.ConditionFalse {
```
mprahl, line 371: Does updateStatus already check this? If so, you can remove this and I believe there's another place where this comment is relevant.

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			err := r.updateStatus(ctx, policy, mismatchCond("Subscription"), mismatchedObj(foundSub))
+			if err != nil {
+				return fmt.Errorf("error updating status for a mismatched Subscription: %w", err)
+			}
 
-		return err
+			if policy.Spec.RemediationAction.IsEnforce() {
+				desiredSub.ResourceVersion = foundSub.GetResourceVersion()
+
+				err := r.Update(ctx, desiredSub)
```
mprahl, line 386: Same comment about field validation.

controllers/operatorpolicy_status.go
```diff
@@ -0,0 +1,429 @@
...
+// condition and related objects can match what is already in the status - in that case, no API
+// calls are made. The `lastTransitionTime` on a condition is not considered when checking if the
+// condition has changed. If not provided, the `lastTransitionTime` will use "now". It also handles
+// preserving the `CreatedByPolicy` property on relatedObjects.
+//
+// This function assumes that all given related objects are of the same kind.
+//
+// Note that only changing the related objects will not emit a new compliance event, but will update
+// the status.
+func (r *OperatorPolicyReconciler) updateStatus(
```
mprahl, line 31: I think the controller should send one compliance event per compliance change like the config-policy-controller to prevent so many compliance changes on the policy.

The config-policy-controller does it with "event batches" at the end of `handleObjectTemplates`. I imagine it'd be a lot simpler in the operator policy case since there is no namespace selector and not multiple object templates.

controllers/operatorpolicy_controller.go
```diff
@@ -278,57 +404,42 @@ func (r *OperatorPolicyReconciler) updatePolicyStatus(
...
 
-	subscription.SetGroupVersionKind(subscriptionGVK)
-	subscription.ObjectMeta.Name = policy.Spec.Subscription.Package
-	subscription.ObjectMeta.Namespace = policy.Spec.Subscription.Namespace
-	subscription.Spec = policy.Spec.Subscription.SubscriptionSpec.DeepCopy()
+	sub := make(map[string]interface{})
 
-	OpLog.Info("Creating the subscription", "kind", subscription.Kind,
-		"namespace", subscription.Namespace)
+	err := json.Unmarshal(policy.Spec.Subscription.Raw, &sub)
```
mprahl, line 412: Optional: Similar comments as in `buildOperatorGroup` for using unstructured.

controllers/operatorpolicy_controller.go
```diff
@@ -278,57 +404,42 @@ func (r *OperatorPolicyReconciler) updatePolicyStatus(
...
-func (r *OperatorPolicyReconciler) setCompliance(
-	ctx context.Context,
-	policy *policyv1beta1.OperatorPolicy,
-	compliance policyv1.ComplianceState,
-) {
-	policy.Status.ComplianceState = compliance
+	spec := new(operatorv1alpha1.SubscriptionSpec)
 
-	err := r.updatePolicyStatus(ctx, policy)
+	err = json.Unmarshal(policy.Spec.Subscription.Raw, spec)
```
mprahl, line 424: Optional: Like I mentioned in `buildOperatorGroup`, are you able to just use this immediately rather than go the map route and then the typed route?

mprahl, line 424: I don't think this will catch invalid/unknown fields. It looks like the unstructured package's converter to typed has some unknown field validation.

controllers/operatorpolicy_controller.go
```diff
@@ -278,57 +404,42 @@ func (r *OperatorPolicyReconciler) updatePolicyStatus(
...
+	subscription.ObjectMeta.Namespace = ns
+	subscription.Spec = spec
 
-	OpLog.Info("Creating the operator group", "kind", operatorGroup.Kind,
-		"namespace", operatorGroup.Namespace)
+	return subscription, nil
+}
 
-	return operatorGroup
+func opPolIdentifier(namespace, name string) depclient.ObjectIdentifier {
```
mprahl, line 437: Optional: This seems like a good candidate as a method on the OperatorPolicy struct.

controllers/operatorpolicy_status.go
```diff
@@ -0,0 +1,429 @@
...
+	} else if conditionChanged(updatedCondition, existingCondition) {
+		condChanged = true
+
+		policy.Status.Conditions[condIdx] = updatedCondition
+	}
+
+	if condChanged {
+		updatedComplianceCondition := calculateComplianceCondition(policy)
+
+		compCondIdx, _ := policy.Status.GetCondition(updatedComplianceCondition.Type)
```
mprahl, line 58: Is this different than `condIdx`?

controllers/operatorpolicy_status.go
```diff
@@ -0,0 +1,429 @@
...
+		if compCondIdx == -1 {
+			policy.Status.Conditions = append(policy.Status.Conditions, updatedComplianceCondition)
+		} else {
+			policy.Status.Conditions[compCondIdx] = updatedComplianceCondition
+		}
+
+		// Sort the conditions based on their type.
+		sort.SliceStable(policy.Status.Conditions, func(i, j int) bool {
```
mprahl, line 66: Nice!

controllers/operatorpolicy_status.go
```diff
@@ -0,0 +1,429 @@
...
+			return policy.Status.Conditions[i].Type < policy.Status.Conditions[j].Type
+		})
+
+		if updatedComplianceCondition.Status == metav1.ConditionTrue {
```
mprahl, line 70: Doesn't this need to check all conditions including existing ones?

controllers/operatorpolicy_status.go
```diff
@@ -0,0 +1,429 @@
...
+
+		err := r.emitComplianceEvent(ctx, policy, updatedComplianceCondition)
+		if err != nil {
+			return err
+		}
+	}
+
+	relObjsChanged := false
+
+	prevRelObjs := policy.Status.RelatedObjsOfKind(updatedRelatedObjs[0].Object.Kind)
```
mprahl, line 84: Could you guard against an empty `updatedRelatedObjs`?

controllers/operatorpolicy_status.go
```diff
@@ -0,0 +1,429 @@
...
+//
+// This function assumes that all given related objects are of the same kind.
+//
+// Note that only changing the related objects will not emit a new compliance event, but will update
+// the status.
+func (r *OperatorPolicyReconciler) updateStatus(
+	ctx context.Context,
+	policy *policyv1beta1.OperatorPolicy,
+	updatedCondition metav1.Condition,
+	updatedRelatedObjs ...policyv1.RelatedObject,
```
mprahl, line 35: Could you document that you only support one kind in the provided related objects?

controllers/operatorpolicy_status.go
```diff
@@ -0,0 +1,429 @@
...
+
+	relObjsChanged := false
+
+	prevRelObjs := policy.Status.RelatedObjsOfKind(updatedRelatedObjs[0].Object.Kind)
+	if len(prevRelObjs) == len(updatedRelatedObjs) {
+		for _, prevObj := range prevRelObjs {
+			nameFound := false
+
+			for i, updatedObj := range updatedRelatedObjs {
+				if prevObj.Object.Metadata.Name == updatedObj.Object.Metadata.Name {
```
mprahl, line 90: Optional: You can make this if condition the opposite and `continue` make this less indented. The same goes for below.

controllers/operatorpolicy_status.go
```diff
@@ -0,0 +1,429 @@
...
+							updatedRelatedObjs[i].Properties.CreatedByPolicy = prevObj.Properties.CreatedByPolicy
+						}
+					}
+
+					if prevObj.Compliant != updatedObj.Compliant {
+						relObjsChanged = true
+					} else if prevObj.Reason != updatedObj.Reason {
+						relObjsChanged = true
+					}
+				}
```
mprahl, line 107: Is it safe to `break` if `relObjsChanged = true` here?

controllers/operatorpolicy_status.go
```diff
@@ -0,0 +1,429 @@
...
+			}
+
+			if !nameFound {
+				relObjsChanged = true
+			}
```
mprahl, line 112: Is it safe to `break` if `relObjsChanged = true` here?

controllers/operatorpolicy_controller.go
JustinKuli: This is slightly different than what `updateStatus` checks. The code comment in the OperatorGroup section explains it more, and I should copy it down here.

This check prevents a previous "created" message from being overwritten with a "matches" message. My feeling was that the created message is more context and is generally more useful, but now I'm thinking that it might be confusing this way, because that's not how the config-policy status and events do it...

controllers/operatorpolicy_status.go
JustinKuli: The number of status updates and compliance events is a valid critique. I wanted to send the updates and event immediately, because I felt it simplified error handling and limited the amount of information that needed to be returned by other functions.

I find what the config policy does with event batches very difficult to reason through... Can you carefully define "one compliance event per compliance change"?

Imagine a fresh enforced policy. With this implementation we get 4 status updates and events that look kinda like this:
- NonCompliant; the operator group is missing (subscription details omitted because they are still unknown)
- NonCompliant; the operator group was created (subscription details omitted because they are still unknown)
- NonCompilant; the operator group was created and the subscription is missing
- Compliant; the operator group was created and the subscription was created

Are you asking for just two events like these?
- NonCompliant; the operator group is missing, and the subscription is missing, 
- Compliant; the operator group was created, and the subscription was created

controllers/operatorpolicy_status.go
JustinKuli: Yes, `compCondIdx` is the index of the (overall) Compliance condition; `condIdx` is the index of the more specific condition that was passed in to be updated. When the passed-in condition does not match what is currently in the status, the overall compliance needs to be recalculated and possibly updated.

controllers/operatorpolicy_status.go
JustinKuli: That is what `calculateComplianceCondition` does. This `updatedComplianceCondition` is the overall compliance.

controllers/operatorpolicy_status.go
JustinKuli: Yes, good idea.

controllers/operatorpolicy_status.go
JustinKuli: Yes.

controllers/operatorpolicy_status.go
JustinKuli: Yes.

JustinKuli (commented):
I'll work on implementing most of these changes tomorrow.

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +101,300 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		err := r.updateStatus(ctx, policy, missingWantedCond("OperatorGroup"), missingWantedObj(desiredOpGroup))
+		if err != nil {
+			return fmt.Errorf("error updating the status for a missing OperatorGroup: %w", err)
+		}
 
-	// Exists indicates if a Subscription or Operatorgroup need to be created
-	exists := subExists && ogExists
-	shouldExist := strings.EqualFold(string(policy.Spec.ComplianceType), string(policyv1.MustHave))
+		if policy.Spec.RemediationAction.IsEnforce() {
+			err = r.Create(ctx, desiredOpGroup)
```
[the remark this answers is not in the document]
mprahl: @JustinKuli the oldest version we need to support is OCP 4.12 (https://access.redhat.com/articles/7027073). This seems to map to OLM API v0.17.3:
https://github.com/openshift/operator-framework-olm/blob/release-4.12/go.mod#L16

If you go to this link to compare v0.17.3 to the latest version, you can see that a couple of minor fields were added to the subscription:
https://github.com/operator-framework/api/compare/v0.17.3...v0.21.0

It's easiest to check the crds/operators.coreos.com_subscriptions.yaml file.

So this would mean that using the latest github.com/operator-framework/api is fine but you need to error if those new fields were provided on OCP 4.12.

api/v1beta1/operatorpolicy_types.go
```diff
@@ -97,30 +64,66 @@ type StatusConfig struct {
...
-	// Subscription defines an Application that can be installed
+	OperatorGroup *runtime.RawExtension `json:"operatorGroup,omitempty"`
+
+	// Include the namespace, and any `spec` fields for the Subscription.
+	// For more info, see `kubectl explain subscription.spec` or
+	// https://olm.operatorframework.io/docs/concepts/crds/subscription/
 	// +kubebuilder:validation:Required
-	Subscription SubscriptionSpec `json:"subscription"`
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Subscription runtime.RawExtension `json:"subscription"`
```
mprahl, line 81: Would it be possible to have Subscription be a custom struct with the name and namespace fields and have the subscription fields as `runtime.RawExtension`? I think using the `inline` JSON tag may make this work.

controllers/operatorpolicy_controller.go
mprahl: I see, thanks! I posted a comment on the types to continue this discussion.

controllers/operatorpolicy_status.go
mprahl: Yeah, exactly. If I have time today, I can try to make the event batching logic more generic since I wrote that code.

The gist of the approach is that each `handle` method would return a list of events. So say operator group returned three events (two noncompliant) and subscription returned two event (one noncompliant), you would end up with three events in this way:

1. noncompliant operator group + existing state
1. noncompliant operator group and noncompliant subscription
1. compliant operator group and compliant subscription

controllers/operatorpolicy_status.go
```diff
@@ -0,0 +1,429 @@
...
+
+	// FUTURE: check additional conditions
+
+	if foundNonCompliant {
+		return metav1.Condition{
+			Type:               compliantConditionType,
+			Status:             metav1.ConditionFalse,
+			LastTransitionTime: metav1.Now(),
+			Reason:             "NonCompliant",
+			Message:            "NonCompliant; " + strings.Join(messages, ", "),
```
mprahl, line 205: Optional: would a semicolon as the join separator be better?

deploy/crds/kustomize_operatorpolicy/kustomization.yaml
```diff
@@ -1,9 +1,9 @@
 resources:
 - policy.open-cluster-management.io_operatorpolicies.yaml
 
-# Add validation more complicated than Kubebuilder markers can provide
 patches:
-- path: ns-selector-validation.json
+# The OperatorPolicy currently only supports "musthave"
+- path: allowed-compliance-types.json
```
mprahl, line 6: This is a good workaround.

controllers/operatorpolicy_status.go
JustinKuli: I think the comma more closely matches what config-policy does.

api/v1beta1/operatorpolicy_types.go
JustinKuli: I can check.

JustinKuli: I think this was a neat idea but I couldn't get it to work. I tried like this:
```golang
type Subscription struct {
	Namespace string `json:"namespace"`
	Spec runtime.RawExtension `json:",inline"`
}
// ...
func buildSubscription() {
// ...
	spec := new(operatorv1alpha1.SubscriptionSpec)

	err := json.Unmarshal(policy.Spec.Subscription.Spec.Raw, spec)
	if err != nil {
		return nil, fmt.Errorf("error unmarshalling subscription: %w", err)
	}
// ...
}
```

The `json.Unmarshal` always returned an error, it looks like the RawExtension is always empty.

controllers/operatorpolicy_controller.go
JustinKuli: I have created a condition for this case with the message `the policy does not specify an OperatorGroup but one already exists in the namespace - assuming that OperatorGroup is correct`

controllers/operatorpolicy_controller.go
JustinKuli: I have removed the `if` preventing the “matches” condition from being added if the previous condition was “created”. This should match how the config-policy works. The tests will need some changes to check this.

controllers/operatorpolicy_status.go
JustinKuli: It does on line 27 (GitHub wouldn't give me a url for it):
```
// This function requires that all given related objects are of the same kind.
```

Or did you want the warning somewhere else

controllers/operatorpolicy_status.go
JustinKuli: Actually no. It needs to continue looping in order to ensure that the `createdByPolicy` property is correctly set. But this comment made me notice that it wasn’t always doing that, so it should be better at it now.

JustinKuli:
I'm expecting the e2e tests to be failing. One of my first revisions to the PR was to add debugging to the test, but actually the way I tried to do that made it so the test could never register its failure 😆

Work Item 1: As pointed out, some of the logic currently works more like "mustonlyhave" than "musthave", and with some of the type/untyped changes more recently, defaulted fields need to be considered more carefully. I will bring in some machinery from config-policy to help with this, and I expect it will also bring options for strict field validation. The `build` functions might return Unstructureds in the future.

Work Item 2: fewer compliance events / status update API calls. @mprahl in your opinion, is this required for this implementation, or is it something we could look at improving in the future?

mprahl:
> I'm expecting the e2e tests to be failing. One of my first revisions to the PR was to add debugging to the test, but actually the way I tried to do that made it so the test could never register its failure 😆
> 
> Work Item 1: As pointed out, some of the logic currently works more like "mustonlyhave" than "musthave", and with some of the type/untyped changes more recently, defaulted fields need to be considered more carefully. I will bring in some machinery from config-policy to help with this, and I expect it will also bring options for strict field validation. The `build` functions might return Unstructureds in the future.
> 
> Work Item 2: fewer compliance events / status update API calls. @mprahl in your opinion, is this required for this implementation, or is it something we could look at improving in the future?

I think work item 2 can be delayed. It shouldn't change QE's tests so it should be safe to handle early next sprint and get it in by the release. I'd be happy to do it if you create a ticket.

api/v1beta1/operatorpolicy_types.go
```diff
@@ -97,30 +64,66 @@ type StatusConfig struct {
...
+
+	return objs
+}
+
+// Searches the conditions of the policy, and returns the index and condition matching the
+// given condition Type. It will return -1 as the index if no condition of the specified
+// Type is found.
+func (status OperatorPolicyStatus) GetCondition(condType string) (int, metav1.Condition) {
+	for i, cond := range status.Conditions {
+		if cond.Type == condType {
```
yiraeChristineKim, line 121: What if multiple conditions match? is it okay to return only the first match?

controllers/configurationpolicy_controller.go
```diff
@@ -547,7 +547,7 @@ func (r *ConfigurationPolicyReconciler) cleanUpChildObjects(plc policyv1.Configu
 ) []string {
 	deletionFailures := []string{}
 
-	if !strings.EqualFold(string(plc.Spec.RemediationAction), "enforce") {
+	if !plc.Spec.RemediationAction.IsEnforce() {
```
yiraeChristineKim, line 550: I love this

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +104,345 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		if err != nil {
-			OpLog.Error(err, "Could not end query batch for the watcher", "watcher", watcher.Name,
-				"watcherKind", watcher.Kind)
+			OpLog.Error(err, "Could not end query batch for the watcher")
 		}
 	}()
 
 	// handle the policy
-	OpLog.Info("Reconciling OperatorPolicy", "policy", policy.Name)
+	OpLog.Info("Reconciling OperatorPolicy")
```
yiraeChristineKim, line 120: why `"policy", policy.Name`  are removed?

controllers/operatorpolicy_controller.go
```diff
@@ -108,167 +104,345 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 
-			return reconcile.Result{}, nil
+		// The names match, but the specs don't: report NonCompliance
+		err = r.updateStatus(ctx, policy, mismatchCond("OperatorGroup"), mismatchedObj(&opGroup))
+		if err != nil {
+			return fmt.Errorf("error updating the status for an OperatorGroup that does not match: %w", err)
+		}
+
+		if policy.Spec.RemediationAction.IsEnforce() {
+			desiredOpGroup.ResourceVersion = opGroup.GetResourceVersion()
```
yiraeChristineKim, line 264: Why newly updated opgroup should have the same resourceversion as old one?

api/v1beta1/operatorpolicy_types.go
JustinKuli: I think it's ok to just return the first match. Ideally we would have some way of enforcing that there are not multiple conditions of the same type (the actual k8s APIs can use listType=map [like this](https://github.com/kubernetes/apimachinery/blob/v0.27.7/pkg/apis/meta/v1/types.go#L1475), but that didn't seem to be available for CRDs). We'll just need to be careful not to accidentally put the same condition twice.

controllers/operatorpolicy_controller.go
JustinKuli: The same information (and more) is included in the fields of the logger passed in via the context from controller-runtime. This log message now looks like this:
```
2024-01-30T16:57:21.964-0500	info	controllers/operatorpolicy_controller.go:120	Reconciling OperatorPolicy	{"controller": "operator-policy-controller", "controllerGroup": "policy.open-cluster-management.io", "controllerKind": "OperatorPolicy", "OperatorPolicy": {"name":"oppol-no-group","namespace":"operator-policy-testns"}, "namespace": "operator-policy-testns", "name": "oppol-no-group", "reconcileID": "9b961756-90ff-4df1-8e75-5a13ca4c1493"}
```

controllers/operatorpolicy_controller.go
JustinKuli: It won't really, but when sending an update request, the API requires what version you were looking at. That way, you don't submit an update that might conflict with something that changed in the meantime.

yiraeChristineKim: oh I see Thanks!!

api/v1beta1/operatorpolicy_types.go
yiraeChristineKim: good to learn Thanks

test/e2e/case38_install_operator_test.go
```diff
@@ -1,126 +1,518 @@
...
+				g.Expect(foundMatchingName).To(BeTrue())
+			}
+
+			idx, actualCondition := policy.Status.GetCondition(expectedCondition.Type)
+			g.Expect(idx).NotTo(Equal(-1))
+			g.Expect(actualCondition.Status).To(Equal(expectedCondition.Status))
+			g.Expect(actualCondition.Reason).To(Equal(expectedCondition.Reason))
+			g.Expect(actualCondition.Message).To(Equal(expectedCondition.Message))
+
+			g.Expect(utils.GetMatchingEvents(
```
yiraeChristineKim, line 74: Is this better wrap with Eventually?

yiraeChristineKim, line 74: Isn't it better to wrap it with Eventually? it may cause flaky issue

JustinKuli: Yes - this function is now called in an `Eventually`, and then also a `Consistently`; the idea is to check that it occurs at some point, and then when it happens, it is the "final" state of the resource. Without the `Consistently`, a bug I had in the code was causing it to pass the tests incorrectly sometimes. With the `Consistently`, the bug was always causing a failure (until I fixed it)

test/e2e/case38_install_operator_test.go
```diff
@@ -1,126 +1,518 @@
...
+					Type:    "OperatorGroupCompliant",
+					Status:  metav1.ConditionFalse,
+					Reason:  "OperatorGroupMismatch",
+					Message: "the OperatorGroup found on the cluster does not match the policy",
+				},
+				"the OperatorGroup found on the cluster does not match the policy",
+			)
+		})
+		It("Should match when the OperatorGroup is manually corrected", func() {
+			utils.Kubectl("delete", "operatorgroup", incorrectOpGroupName, "-n", opPolTestNS)
```
yiraeChristineKim, line 267: OperatorGroup is specified and another OperatorGroup also exists, then what happens?

yiraeChristineKim: Can we add one more case(it)

dhaiducek (commented):
Taking a look at the new PR also, but I wanted to give this one a look first. 🙂

PR #195, opened 2024-02-01
Enable status reporting for CatalogSource in OperatorPolicy

zyjjay:
Following the schema of implementing multiple `handle` functions, the CatalogSource status reporting logic utilizes the same `updateStatus` function to facilitate the `conditions` and `relatedObject` field updates on the related OperatorPolicy. The controller reads the conditions field on the watched Subscription to determine the health of the CatalogSource. In the event a Subscription is not created due to either an `inform` OperatorPolicy or an error, the controller directly checks if the CatalogSource object exists on the cluster to propagate relevant health information up to the OperatorPolicy.

ref: https://issues.redhat.com/browse/ACM-9285


controllers/operatorpolicy_controller.go
```diff
@@ -133,9 +146,96 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	desiredSub, err := buildSubscription(policy)
+	if err != nil {
+		return fmt.Errorf("error building Subscription: %w", err)
+	}
+
+	// Retrieve Subscription
+	foundSub, err := r.DynamicWatcher.Get(watcher, subscriptionGVK, desiredSub.Namespace, desiredSub.Name)
+	if err != nil {
+		return fmt.Errorf("error getting Subscription: %w", err)
+	}
```
JustinKuli, line 174: I think it will be better if we use a subscription returned by `handleSubscription`. I'll share a snippet that we can either bring into this PR, or it might be ok to accept this as-is and adjust it in another PR.

controllers/operatorpolicy_controller.go
```diff
@@ -133,9 +146,96 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+	if err != nil {
+		return fmt.Errorf("error decoding object into array type: %w", err)
+	}
+
+	// Iterate over all the conditions to determine the state of the CatalogSource
+	for _, condition := range condList {
+		condMap, ok := condition.(map[string]interface{})
+		if !ok {
+			return fmt.Errorf("conditions field is of the wrong type")
```
JustinKuli, line 208: This makes me realize that we might need to keep track of 2 kinds of errors: this one here is something that re-reconciling can not fix. So if this occurs, any `handle` functions after it won't be run, which means the status might always be incomplete. That's probably not ideal, I think it's something we'll need to deal with eventually.

On the other hand, it's possible to have an error when reading from the API server, which does just need a retry, so I think re-reconciling in that case is fine.

Anyway, I think this is fine for this PR, I was just mentioning it for awareness.

controllers/operatorpolicy_controller.go
```diff
@@ -133,9 +146,96 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		}
+
+		if fmt.Sprintf("%v", condMap["type"]) == catalogSrcConditionType {
```
JustinKuli, line 211: It took me a minute to find a reason for using `fmt.Sprintf` throughout this section... did you choose this because it saves a bunch of error-checking for types?

I think I'm warming up to the idea, I would just like a comment explaining it once somewhere.

controllers/operatorpolicy_controller.go
```diff
@@ -133,9 +146,96 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			if fmt.Sprintf("%v", condMap["status"]) == string(metav1.ConditionFalse) {
+				// Emit healthy event using the defined status template
+				err := r.updateStatus(ctx, policy, catalogSourceFindCond(false), catalogSourceObj(desiredSub, false))
```
JustinKuli, line 214: So if the Subscription says that the CatalogSource isn't unhealthy, this just reports it as found... Does the Subscription have a reason and message in this case, can we report that instead?

I'm thinking it would be more complete if, when the condition is found, it copies over all of the information from that condition, like it currently does in the `else` case below.

controllers/operatorpolicy_controller.go
```diff
@@ -133,9 +146,96 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+				if err != nil {
+					return fmt.Errorf("error updating the status for an existing CatalogSource: %w", err)
+				}
+			} else {
+				// Emit unhealthy event by inheriting condition fields from the Subscription
+				failureCond := metav1.Condition{
+					Type:               fmt.Sprintf("%v", condMap["type"]),
+					Status:             metav1.ConditionTrue,
+					LastTransitionTime: metav1.Now(),
```
JustinKuli, line 223: Is this not in the condition on the subscription?

controllers/operatorpolicy_controller.go
```diff
@@ -414,7 +514,7 @@ func (r *OperatorPolicyReconciler) handleSubscription(ctx context.Context, polic
 		// For now, just mark it as compliant
 		err := r.updateStatus(ctx, policy, matchesCond("Subscription"), matchedObj(foundSub))
 		if err != nil {
-			return fmt.Errorf("error updating the status for an OperatorGroup that matches: %w", err)
+			return fmt.Errorf("error updating the status for an Subscription that matches: %w", err)
```
JustinKuli, line 517: Good catch! Fix the `an` at the same time?
```suggestion
			return fmt.Errorf("error updating the status for a Subscription that matches: %w", err)
```

controllers/operatorpolicy_status.go
```diff
@@ -373,6 +386,51 @@ var opGroupTooManyCond = metav1.Condition{
...
+	return metav1.Condition{
+		Type:    "CatalogSourcesUnhealthy",
+		Status:  status,
+		Reason:  reason,
+		Message: message,
+	}
+}
+
+// catalogSourceObj returns a conditionally compliant RelatedObject with reason based on the `isMissing` parameter
+func catalogSourceObj(sub *operatorv1alpha1.Subscription, isMissing bool) policyv1.RelatedObject {
```
JustinKuli, line 411: I think I'd prefer if this just took a name and namespace, not a whole subscription.

test/e2e/case40_catalogsource_status_report_test.go
```diff
@@ -0,0 +1,157 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	policyv1 "open-cluster-management.io/config-policy-controller/api/v1"
+	policyv1beta1 "open-cluster-management.io/config-policy-controller/api/v1beta1"
+	"open-cluster-management.io/config-policy-controller/test/utils"
+)
+
+var _ = Describe("Test status reporting for catalogsource", Ordered, func() {
```
JustinKuli, line 15: Is there a problem if we merge this into the big describe in the case38 file? I think it's kind of nice to be able to quickly run all of the OperatorPolicy tests at once right now.

JustinKuli (commented):
You've got the right idea overall!

controllers/operatorpolicy_controller.go
zyjjay: I think @JeffeyL's PR was going to modify `handleSubscription` to return the subscription. I wanted to draft this one up to get some reviews first but I can always adjust it later.

controllers/operatorpolicy_controller.go
zyjjay: Yea, I think we probably want some type of error message that can be displayed/logged, but not re-reconcile when this type of error happens.

controllers/operatorpolicy_controller.go
zyjjay: Forgot to mention but this was one of the implementation details that I wanted feedback on. Is this the best way to type cast `interface{}` to `string`? I chose this way because it was one that worked, but I was wondering if this was the best way to approach this.

controllers/operatorpolicy_controller.go
zyjjay: I think originally I didn't have separate logic to handle when the `Subscription` did not exist so we couldn't read from the conditions field. With the most recent changes I think I can change this logic to inherit the `Subscription`'s condition instead.

controllers/operatorpolicy_controller.go
zyjjay: Haha missed that!

test/e2e/case40_catalogsource_status_report_test.go
zyjjay: Yea I could do that. I was thinking of moving the `check` function to a common file since it's used by a few files, but moving the entire test into one file works too!

controllers/operatorpolicy_controller.go
zyjjay: @JustinKuli

controllers/operatorpolicy_controller.go
zyjjay: @JustinKuli

controllers/operatorpolicy_controller.go
JustinKuli: I would've expected something like
```golang
condType, ok := condMap["type"].(string)
if !ok {
    // handle the situation where it isn't a string or the field wasn't in the map
}
```

Without the `, ok`, I think it could panic which we would not want.

I don't think the `fmt.Sprintf` like this will ever panic, probably gives a usefulresult even in odd situations, and at worst might just be an empty string.

controllers/operatorpolicy_controller.go
JustinKuli: I think you would need to marshal it to json and then unmarshal it again, or maybe you can use an unstructured converter.

I think a better approach here overall would be to convert the Subscription itself out of Unstructured, and then you get all the conditions and other details more easily.

controllers/operatorpolicy_controller.go
```diff
@@ -134,9 +147,74 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	ctx context.Context,
+	policy *policyv1beta1.OperatorPolicy,
+	subscription *operatorv1alpha1.Subscription,
+) error {
+	watcher := opPolIdentifier(policy.Namespace, policy.Name)
+
+	catalogName := subscription.Spec.CatalogSource
+	catalogNS := subscription.Spec.CatalogSourceNamespace
+
+	if len(subscription.Status.Conditions) == 0 {
```
JustinKuli, line 170:
```suggestion
	if subscription == nil || len(subscription.Status.Conditions) == 0 {
```

controllers/operatorpolicy_controller.go
```diff
@@ -134,9 +147,74 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 
+func (r *OperatorPolicyReconciler) handleCatalogSource(
+	ctx context.Context,
+	policy *policyv1beta1.OperatorPolicy,
+	subscription *operatorv1alpha1.Subscription,
+) error {
+	watcher := opPolIdentifier(policy.Namespace, policy.Name)
+
+	catalogName := subscription.Spec.CatalogSource
+	catalogNS := subscription.Spec.CatalogSourceNamespace
```
JustinKuli, line 168: I think we need to guard against the situation where the subscription is nil here.

controllers/operatorpolicy_controller.go
```diff
@@ -134,9 +147,74 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		if err != nil {
+			return fmt.Errorf("error updating the status for CatalogSource when Subscription DNE on cluster: %w", err)
+		}
+
+		return nil
+	}
+
+	// Iterate over all the conditions to determine the state of the CatalogSource
+	for _, condition := range subscription.Status.Conditions {
+		if condition.Type == catalogSrcConditionType {
```
JustinKuli, line 192: Does this need this kind of conversion? Maybe you mentioned this in a separate thread, but now I feel like I'm missing something.
```suggestion
		if string(condition.Type) == string(catalogSrcConditionType) {
```

controllers/operatorpolicy_controller.go
```diff
@@ -134,9 +147,74 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			status := metav1.ConditionFalse
+			if string(condition.Status) == string(metav1.ConditionTrue) {
+				status = metav1.ConditionTrue
+			}
+
+			isMissing := string(condition.Status) == string(metav1.ConditionTrue)
```
JustinKuli, line 198: I think it can be unhealthy without it being "missing", for example if it's pointing to an image that it can't pull, or an image that isn't the right format for a catalog source. So I think that this variable is misnamed, and that the condition/related obj that we put in based on it shouldn't just be "found" or "missing".

controllers/operatorpolicy_controller.go
```diff
@@ -134,9 +147,74 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+			healthCondition := metav1.Condition{
+				Type:               string(condition.Type),
```
JustinKuli, line 201: I think it would be more clear here to just put:
```suggestion
				Type:               string(catalogSrcConditionType),
```
because that is true based on an earlier condition, and the reader doesn't need to check anywhere else to find that information.

controllers/operatorpolicy_status.go
```diff
@@ -373,6 +385,51 @@ var opGroupTooManyCond = metav1.Condition{
...
+
+	if !isMissing {
+		compliance = string(policyv1.Compliant)
+		reason = reasonWantFoundExists
+	}
+
+	return policyv1.RelatedObject{
+		Object: policyv1.ObjectResource{
+			Kind:       "CatalogSource",
+			APIVersion: "operators.coreos.com/v1alpha1",
```
JustinKuli, line 422: We can use the catalogSrcGVK here.

test/e2e/case38_install_operator_test.go
```diff
@@ -550,4 +550,110 @@ var _ = Describe("Test installing an operator from OperatorPolicy", Ordered, fun
...
+			utils.Kubectl("create", "ns", opPolTestNS)
+			DeferCleanup(func() {
+				utils.Kubectl("delete", "ns", opPolTestNS)
+			})
+
+			createObjWithParent(parentPolicyYAML, parentPolicyName,
+				OpPlcYAML, opPolTestNS, gvrPolicy, gvrOperatorPolicy)
+		})
+
+		It("Should initially be Compliant", func() {
```
JustinKuli, line 571: Nit:
```suggestion
		It("Should initially show the CatalogSource is compliant", func() {
```
just to be more accurate, since the policy overall is not compliant.

controllers/operatorpolicy_controller.go
zyjjay: Would the subscription ever be nil? As far as I can tell, `handleSubscription` only returns a nil subscription with an error, which would return before it can hit the `handleCatalogSource` block.

controllers/operatorpolicy_controller.go
zyjjay: I was thinking since `condition.Type`'s underlying type is the a string, comparing them won't be an issue. However, I'm not opposed to type casting both to strings for robustness.

controllers/operatorpolicy_controller.go
JustinKuli: Right now, it shouldn't ever be nil when it gets here. But things can change in the future, and if it does get here as `nil`, I think the controller would panic. I'd feel better to be on the safe side.

controllers/operatorpolicy_status.go
```diff
@@ -461,11 +480,98 @@ func buildDeploymentCond(
...
+
+// catalogSourceUnknownCond is a NonCompliant condition
+func catalogSourceUnknownCond() metav1.Condition {
+	return metav1.Condition{
+		Type:    "CatalogSourcesUnknownState",
+		Status:  metav1.ConditionTrue,
+		Reason:  "LastObservedUnknown",
+		Message: "Could not determine last observed state of CatalogSource",
+	}
+}
```
JustinKuli, line 528: Nit:
```suggestion
var catalogSourceUnknownCond = metav1.Condition{
		Type:    "CatalogSourcesUnknownState",
		Status:  metav1.ConditionTrue,
		Reason:  "LastObservedUnknown",
		Message: "Could not determine last observed state of CatalogSource",
}
```

controllers/operatorpolicy_status.go
```diff
@@ -461,11 +480,98 @@ func buildDeploymentCond(
...
-	Message: "The ClusterServiceVersion is missing, thus meaning there are no relevant deployments",
+func getCatalogCondAndObj(
+	catalogName string,
+	catalogNS string,
+	isUnhealthy bool,
+	isMissing bool,
+) (metav1.Condition, policyv1.RelatedObject) {
+	return catalogSourceFindCond(isUnhealthy, isMissing),
+		catalogSourceObj(catalogName, catalogNS, isUnhealthy, isMissing)
+}
```
JustinKuli, line 491: I think this function just adds an unnecessary extra layer when reading the code. I'd rather just have both of the functions in here used individually.

controllers/operatorpolicy_controller.go
```diff
@@ -165,9 +171,89 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			return nil
+		}
+
+		CatalogSrcState := catalogSrc.Status.GRPCConnectionState.LastObservedState
+
+		if CatalogSrcState == CatalogSourceReady {
+			isUnhealthy = false
+		} else {
+			isUnhealthy = true
+		}
```
JustinKuli, line 244: Nit:
```suggestion
		isUnhealthy = (CatalogSrcState != CatalogSourceReady)
```

controllers/operatorpolicy_status.go
```diff
@@ -427,6 +439,13 @@ var noCSVCond = metav1.Condition{
 	Message: "A relevant installed ClusterServiceVersion could not be found",
 }
 
+var noDeploymentsCond = metav1.Condition{
```
JustinKuli, line 442: I wasn't expecting to see this in this PR's diff, it might indicate a merge conflict gone wrong?

test/e2e/case38_install_operator_test.go
```diff
@@ -597,79 +597,161 @@ var _ = Describe("Test installing an operator from OperatorPolicy", Ordered, fun
...
+	Describe("Test status reporting for CatalogSource", Ordered, func() {
 		const (
-			opPolYAML = "../resources/case38_operator_install/operator-policy-no-group-csv-fail.yaml"
-			opPolName = "oppol-no-allnamespaces"
+			OpPlcYAML  = "../resources/case38_operator_install/operator-policy-with-group.yaml"
+			OpPlcName  = "oppol-with-group"
+			subName    = "project-quay"
+			catSrcName = "operatorhubio-catalog"
 		)
+
```
JustinKuli, line 620: The GitHub diff here makes me suspect there was a merge conflict and we've ended up losing a test scenario.

test/resources/case38_operator_install/case38_OpPlc.yaml
JustinKuli, line 1: I don't see any tests that use this file. Can you double check (since it seems like you modified it for something), and if it isn't used, remove it?

JustinKuli (changes requested):
This mostly looks good! There are a few places where I'm worried there was a merge conflict that didn't quite get resolved correctly. Let me know if you need help looking at that.

JustinKuli:
The first test run had a failure on an unrelated test. I'm re-running the test.

test/e2e/case38_install_operator_test.go
zyjjay: I think I forgot to add in one of the tests from the CSV/deployments PR, I'll add that in now.

test/e2e/case38_install_operator_test.go
```diff
@@ -672,4 +669,148 @@ var _ = Describe("Test installing an operator from OperatorPolicy", Ordered, fun
...
+					Reason:  "CatalogSourcesNotFound",
+					Message: "CatalogSource was not found",
+				},
+				"CatalogSource was not found",
+			)
+
+			By("Patching the policy to point to an existing CatalogSource for next test")
+			utils.Kubectl("patch", "operatorpolicy", OpPlcName, "-n", opPolTestNS, "--type=json", "-p",
+				`[{"op": "replace", "path": "/spec/subscription/source", "value": "operatorhubio-catalog"}]`)
+		})
```
JustinKuli, line 780: Nit: I think this should just go in the next test.

test/e2e/case38_install_operator_test.go
```diff
@@ -672,4 +669,148 @@ var _ = Describe("Test installing an operator from OperatorPolicy", Ordered, fun
...
+					Status:  metav1.ConditionTrue,
+					Reason:  "CatalogSourcesFoundUnhealthy",
+					Message: "CatalogSource was found but is unhealthy",
+				},
+				"CatalogSource was found but is unhealthy",
+			)
+
+			By("Patching the CatalogSource to reference its original valid image link")
+			utils.Kubectl("patch", "catalogsource", catSrcName, "-n", "olm", "--type=json", "-p",
+				`[{"op": "replace", "path": "/spec/image", "value": "quay.io/operatorhubio/catalog:latest"}]`)
```
JustinKuli, line 813: I think this should be moved into a `DeferCleanup` function, at the top of the test, so that the catalog source is still reverted if the test fails.

JustinKuli (changes requested):
Thanks for addressing all of the comments, I think this is functionally ready to go! I noticed two small things in the test now that the GitHub diff was more readable, and I think after that it will be ready to merge.

JustinKuli (approved):
Thanks, LGTM!

PR #196, opened 2024-02-01
Implement OperatorPolicy health checks for CSV

JeffeyL:
This change follows the previous pattern of using `handle` functions in order to handle different resources, in this case for ClusterServiceVersions (CSVs). The `handleSubscription` function was modified in order to return the unstructured subscription to make accessing CSV metadata easier. Since CSVs are managed by OLM, the controller reports conditions directly from the CSV in order to monitor its health. Changes also include updating `relatedObjects` accordingly.


test/e2e/case40_operator_health_checks_test.go
```diff
@@ -0,0 +1,117 @@
...
+	. "github.com/onsi/gomega"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	policyv1 "open-cluster-management.io/config-policy-controller/api/v1"
+	policyv1beta1 "open-cluster-management.io/config-policy-controller/api/v1beta1"
+	"open-cluster-management.io/config-policy-controller/test/utils"
+)
+
+var _ = FDescribe("Test health checks on OLM resources after installation via OperatorPolicy", Ordered, func() {
```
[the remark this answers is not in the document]
JeffeyL: I should be able to do that. My original thought was to have all the OLM resource health check tests in one file but seeing as case38 already tests health checks for operatorgroups and subscriptions I think it should fit nicely as well.

controllers/operatorpolicy_status.go
```diff
@@ -356,6 +372,21 @@ func updatedCond(kind string) metav1.Condition {
 	}
 }
 
+// buildCSVCond takes a csv and returns a shortened version of its most recent Condition
+func buildCSVCond(csv operatorv1alpha1.ClusterServiceVersion) metav1.Condition {
+	status := metav1.ConditionFalse
+	if csv.Status.Phase == operatorv1alpha1.CSVPhaseSucceeded {
```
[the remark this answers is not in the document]
JeffeyL: There are some phases that are "in the middle" such as `Installing`, `Pending`, `Replacing`, etc but I'm not sure how I would categorize those. `Succeeded` to me feels like the only one that should have a `True` condition status, although I could be wrong.

controllers/operatorpolicy_controller.go
```diff
@@ -125,13 +135,21 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	if err := r.handleSubscription(ctx, policy); err != nil {
+	subscription, err := r.handleSubscription(ctx, policy)
+	if err != nil {
 		OpLog.Error(err, "Error handling Subscription")
 
 		return reconcile.Result{}, err
 	}
 
-	// FUTURE: more resource checks
+	csv, err := r.handleCSV(ctx, policy, subscription)
```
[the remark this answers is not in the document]
JeffeyL: Yes, it would be used in checks for deployments.

controllers/operatorpolicy_status.go
```diff
@@ -451,3 +482,45 @@ func opGroupTooManyObjs(opGroups []unstructured.Unstructured) []policyv1.Related
...
+
+	compliance := policyv1.NonCompliant
+	if csv.Status.Phase == operatorv1alpha1.CSVPhaseSucceeded {
+		compliance = policyv1.Compliant
+	}
+
+	return policyv1.RelatedObject{
+		Object:    obj,
+		Compliant: string(compliance),
+		Reason:    string(csv.Status.Reason),
```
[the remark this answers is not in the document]
JeffeyL: This does sound like it would be useful, I'll test it out.

controllers/operatorpolicy_status.go
```diff
@@ -356,6 +372,21 @@ func updatedCond(kind string) metav1.Condition {
...
+	status := metav1.ConditionFalse
+	if csv.Status.Phase == operatorv1alpha1.CSVPhaseSucceeded {
+		status = metav1.ConditionTrue
+	}
+
+	return metav1.Condition{
+		Type:    condType(csv.Kind),
+		Status:  status,
+		Reason:  string(csv.Status.Reason),
+		Message: csv.Status.Message,
```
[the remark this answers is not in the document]
JeffeyL: I'll try a few things - the messages within the CSV are pretty varied so I may end up just prepending something generic to indicate that the message is a CSV message if that's alright.

controllers/operatorpolicy_status.go
```diff
@@ -451,3 +523,82 @@ func opGroupTooManyObjs(opGroups []unstructured.Unstructured) []policyv1.Related
 
 	return objs
 }
+
+func missingCSVObj(sub *operatorv1alpha1.Subscription) policyv1.RelatedObject {
```
JustinKuli, line 527: Nit: I think I would prefer this to just take `name` and `namespace`

controllers/operatorpolicy_status.go
```diff
@@ -451,3 +523,82 @@ func opGroupTooManyObjs(opGroups []unstructured.Unstructured) []policyv1.Related
...
+
+	return policyv1.RelatedObject{
+		Object: policyv1.ObjectResource{
+			Kind:       deploymentGVK.Kind,
+			APIVersion: deploymentGVK.GroupVersion().String(),
+			Metadata: policyv1.ObjectMetadata{
+				Name:      dep.Name,
+				Namespace: dep.Namespace,
+			},
+		},
```
JustinKuli, line 597:
```suggestion
		Object: policyv1.ObjectResourceFromObj(dep),
```
Unless the kind information is missing from the object, sometimes that happens 😅

test/e2e/case38_install_operator_test.go
```diff
@@ -17,7 +17,7 @@ var _ = Describe("Test installing an operator from OperatorPolicy", Ordered, fun
 		opPolTestNS      = "operator-policy-testns"
 		parentPolicyYAML = "../resources/case38_operator_install/parent-policy.yaml"
 		parentPolicyName = "parent-policy"
-		opPolTimeout     = 10
+		opPolTimeout     = 60
```
JustinKuli, line 20: 👀 what's taking up to 60 seconds? I'd rather we adjust it for that specific test somehow. Maybe a separate `Eventually` for whatever that test is waiting a while for, before the main `check`.

It looks like this is only (currently) used in `Eventually` kinds of checks, so keeping this change wouldn't be the end of the world.

controllers/operatorpolicy_controller.go
```diff
@@ -125,17 +141,95 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 
+func (r *OperatorPolicyReconciler) handleDeployment(
+	ctx context.Context,
+	policy *policyv1beta1.OperatorPolicy,
+	csv *operatorv1alpha1.ClusterServiceVersion,
+) error {
+	// case where csv is nil
+	if csv == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
JeffeyL: Wow, I didn't consider that at all.

controllers/operatorpolicy_controller.go
```diff
@@ -486,6 +598,59 @@ func buildSubscription(
...
+	var sub operatorv1alpha1.Subscription
+
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(unstructured, &sub)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get the CSV related to the object
+	foundCSV, err := r.DynamicWatcher.Get(watcher, clusterServiceVersionGVK, sub.Namespace,
+		sub.Status.CurrentCSV)
```
JustinKuli, line 622: We probably need to consider the case where this field is not set in the Subscription. It might be similar to the `subscription == nil` case... right after the subscription is created, the status won't have been populated by OLM yet.

JustinKuli: I'm looking at some other details from my notes, and I think we might want `InstalledCSV` instead. The difference is when there is an InstallPlan pending (for example for an upgrade), I think the `CurrentCSV` points to the "new" version, which won't exist if the plan hasn't been approved yet.

controllers/operatorpolicy_controller.go
```diff
@@ -495,6 +526,143 @@ func buildSubscription(
...
+) error {
+	// case where csv is nil
+	if csv == nil {
+		// need to remove all relatedobjects
+		err := r.removeObjsOfType(ctx, policy, deploymentGVK)
+		if err != nil {
+			return fmt.Errorf("error removing relatedObjects: %w", err)
+		}
+
+		err = r.updateStatus(ctx, policy, noDeploymentsCond, noExistingDeploymentObj)
```
JustinKuli, line 604: Same comment as for CSVs: if `updateStatus` gets `noExistingDeploymentObj`, then the `removeObjsOfType` is unnecessary.

controllers/operatorpolicy_controller.go
```diff
@@ -495,6 +526,143 @@ func buildSubscription(
...
+
+		return nil, nil
+	}
+
+	watcher := opPolIdentifier(policy.Namespace, policy.Name)
+
+	// case where status has not been populated yet
+	if sub.Status.InstalledCSV == "" {
+		return nil, nil
+	}
```
JustinKuli, line 554: I think this should be handled the same as if the subscription was `nil`, it needs to kind of "reset" the condition and relatedObjects.

controllers/operatorpolicy_controller.go
```diff
@@ -495,6 +526,143 @@ func buildSubscription(
...
+			return fmt.Errorf("error getting the Deployment: %w", err)
+		}
+
+		// report missing deployment in relatedObjects list
+		if foundDep == nil {
+			relatedObjects = append(relatedObjects, missingDeploymentObj(dep.Name, csv.Namespace))
+
+			if err != nil {
+				return fmt.Errorf("error updating the status for a missing Deployment: %w", err)
+			}
```
JustinKuli, line 632: Is this error check left over from something else? I don't think it's necessary.

controllers/operatorpolicy_controller.go
```diff
@@ -495,6 +526,143 @@ func buildSubscription(
...
+
+			continue
+		}
+
+		unstructured := foundDep.UnstructuredContent()
+		var dep appsv1.Deployment
+		err = runtime.DefaultUnstructuredConverter.FromUnstructured(unstructured, &dep)
+
+		if err != nil {
+			OpLog.Error(err, "Unable to convert unstructured Deployment to typed")
```
JustinKuli, line 642:
```suggestion
			OpLog.Error(err, "Unable to convert unstructured Deployment to typed", "Deployment.Name", dep.name)
```

controllers/operatorpolicy_status.go
```diff
@@ -373,6 +424,63 @@ var opGroupTooManyCond = metav1.Condition{
...
+	status := metav1.ConditionTrue
+	reason := "DeploymentsAvailable"
+	message := "All operator Deployments have their minimum availability"
+
+	if !depsExist {
+		reason = "NoExistingDeployments"
+		message = "No existing operator Deployments"
+	}
+
+	if unavailable {
```
JustinKuli, line 456: Nit: it seems like this boolean could be replaced with `len(unavailableDeps) != 0`

controllers/operatorpolicy_status.go
```diff
@@ -273,10 +300,30 @@ func (r *OperatorPolicyReconciler) emitComplianceEvent(
...
+
+	for _, relObj := range policy.Status.RelatedObjects {
+		if relObj.Object.Kind != gvk.Kind {
+			newRelObjs = append(newRelObjs, relObj)
+		}
+	}
+
+	policy.Status.RelatedObjects = newRelObjs
+
+	return r.Status().Update(ctx, policy)
```
JustinKuli, line 318: This whole function is not needed now, but it's worth mentioning that if it was used alongside `updateStatus`, this would result in 2 API calls, which would not be optimal. I would've preferred it to be a method on the policy status, and the caller would be expected to do this update call themselves.

JustinKuli (changes requested):
Looking good! It would be nice to have additional test cases for some of the other possible conditions. I know it's difficult to find an operator that will definitely not just work... In one of my OpenShift experiments I found that a subscription like this:
```yaml
spec:
  channel: singlenamespace-alpha
  installPlanApproval: Manual
  name: etcd
  source: community-operators
  sourceNamespace: openshift-marketplace
  startingCSV: etcdoperator.v0.9.0
```
has some problems - have you tried that?

If we really can't find a "broken" situation to test, that's ok for now. Eventually I think we'll need to create our own fake operator that we package so that we can test out odd upgrade situations, including when the deployment is broken, or the CSV gets into a weird status.

controllers/operatorpolicy_controller.go
JeffeyL: Whoops.

controllers/operatorpolicy_controller.go
JeffeyL: I also used the wrong `Cond`, my bad

JeffeyL:
Still working on the additional test cases, will update here when done.

JustinKuli (approved):
LGTM! Thanks for dealing with all of the comments back and forth :)

PR #198, opened 2024-02-05
Return a subscription from `handleSubscription`

JustinKuli:
Some of the additional checks will need pieces of information from the Subscription. Converting it before returning it (as opposed to using an Unstructured) should help reduce other error checking.

### DIFF
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -125,7 +125,8 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return reconcile.Result{}, err
 	}
 
-	if err := r.handleSubscription(ctx, policy); err != nil {
+	_, err = r.handleSubscription(ctx, policy)
+	if err != nil {
 		OpLog.Error(err, "Error handling Subscription")
 
 		return reconcile.Result{}, err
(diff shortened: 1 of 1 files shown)


JustinKuli:
I think this might be useful for @JeffeyL's https://github.com/open-cluster-management-io/config-policy-controller/pull/196 and @zyjjay's https://github.com/open-cluster-management-io/config-policy-controller/pull/195.

controllers/operatorpolicy_controller.go
```diff
@@ -358,98 +359,106 @@ func buildOperatorGroup(
...
 			}
 
 			desiredSub.SetGroupVersionKind(subscriptionGVK) // Create stripped this information
 
 			// Now it should match, so report Compliance
 			err = r.updateStatus(ctx, policy, createdCond("Subscription"), createdObj(desiredSub))
 			if err != nil {
-				return fmt.Errorf("error updating the status for a created Subscription: %w", err)
+				return desiredSub, fmt.Errorf("error updating the status for a created Subscription: %w", err)
 			}
```
zyjjay, line 396: I think returning a subscription inside of the logic that handles the `enforce` remediationAction type would be useful for future `handler` functions. It would make sense for this function to return a subscription only if it was created successfully on the cluster.

controllers/operatorpolicy_controller.go
```diff
@@ -358,98 +359,106 @@ func buildOperatorGroup(
...
 		}
 
-		return nil
+		return desiredSub, nil
```
zyjjay, line 399: I think returning something like `nil, nil` would be better here if we can return the subscription in the `enforce` case. Then in my PR, I could use the return value to determine if the desired subscription exists on the cluster (`nil` vs a subscription object) and skip a call to `r.DynamicWatcher.Get` to retrieve the subscription.

JustinKuli: It's possible... but I think that if `inform` mode returned `nil, nil` here, it would artificially limit some of the information we can get. For example, it is generally possible to check if the CatalogSource specified in the policy exists on the cluster, even if the Subscription doesn't exist.

zyjjay: I might be overlooking something but is the point of this to reduce calls to `buildSubscription`?

JustinKuli: Mostly. This would reduce calls to `buildSubscription` and `dynamicWatcher.Get` for the Subscription.

This can also return the "merged" subscription, regardless of inform/enforce, which hypothetically could be useful for some other healthcheck... I don't see a need for that yet, but it feels like it's easy enough to "do it right" now, so we don't have to worry about it later. For example if some other check used the `channel` in the subscription, this function would always return the *desired* setting based on the policy, which is what I think the user would be expecting.

zyjjay: I see, I think that would be useful in general. For my specific case I'm only calling `dynamicWatcher.Get` on a catalogsource when the subscription doesn't exist yet, so I might need to end up calling `dynamicWatcher.Get` on the subscription anyways in `handleCatalogSource` because the return values of `handleSubscription` doesn't seem to indicate whether or not a subscription exists on the cluster.

controllers/operatorpolicy_controller.go
```diff
@@ -358,98 +359,106 @@ func buildOperatorGroup(
...
 	merged := foundSub.DeepCopy() // Copy it so that the value in the cache is not changed
 
 	updateNeeded, skipUpdate, err := r.mergeObjects(ctx, desiredUnstruct, merged, string(policy.Spec.ComplianceType))
 	if err != nil {
-		return fmt.Errorf("error checking if the Subscription needs an update: %w", err)
+		return nil, fmt.Errorf("error checking if the Subscription needs an update: %w", err)
+	}
+
+	mergedSub := new(operatorv1alpha1.Subscription)
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(merged.Object, mergedSub); err != nil {
```
mprahl, line 416: I just checked and I don't believe this will produce an error if there are unknown fields (e.g. version drift of different OLM versions). So this looks to be safe if you are just reading from it.

PR #199, opened 2024-02-06
Handle InstallPlan approval based on spec.versions

JustinKuli:
When enforcing a policy that specifies which versions should be allowed, the controller will now set the Subscription.InstallPlanApproval to Manual, and will approve only the InstallPlans that match the versions specified in the policy.

The controller now also reports the status of related InstallPlans.

Refs:
 - https://issues.redhat.com/browse/ACM-9286

### DIFF
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -14,6 +14,7 @@ import (
 	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	depclient "github.com/stolostron/kubernetes-dependency-watches/client"
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -63,6 +64,11 @@ var (
 		Group:   "operators.coreos.com",
 		Version: "v1alpha1",
 		Kind:    "CatalogSource",
+	}
+	installPlanGVK = schema.GroupVersionKind{
+		Group:   "operators.coreos.com",
+		Version: "v1alpha1",
+		Kind:    "InstallPlan",
 	}
 )
 
@@ -154,6 +160,12 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	subscription, err := r.handleSubscription(ctx, policy)
 	if err != nil {
 		OpLog.Error(err, "Error handling Subscription")
+
+		return reconcile.Result{}, err
+	}
+
+	if err := r.handleInstallPlan(ctx, policy, subscription); err != nil {
+		OpLog.Error(err, "Error handling InstallPlan")
 
 		return reconcile.Result{}, err
 	}
@@ -447,7 +459,7 @@ func buildOperatorGroup(
 	// Fallback to the Subscription namespace if the OperatorGroup namespace is not specified in the policy.
 	ogNamespace := subNamespace
 
-	if specifiedNS, ok := opGroup["namespace"].(string); ok || specifiedNS == "" {
+	if specifiedNS, ok := opGroup["namespace"].(string); ok && specifiedNS != "" {
 		ogNamespace = specifiedNS
 	}
 
@@ -529,8 +541,28 @@ func (r *OperatorPolicyReconciler) handleSubscription(
 	}
 
 	if !updateNeeded {
-		// FUTURE: Check more details about the *status* of the Subscription
-		// For now, just mark it as compliant
+		subResFailed := mergedSub.Status.GetCondition(operatorv1alpha1.SubscriptionResolutionFailed)
+
+		if subResFailed.Status == corev1.ConditionTrue {
+			cond := metav1.Condition{
+				Type:    subConditionType,
+				Status:  metav1.ConditionFalse,
+				Reason:  subResFailed.Reason,
+				Message: subResFailed.Message,
+			}
+
+			if subResFailed.LastTransitionTime != nil {
+				cond.LastTransitionTime = *subResFailed.LastTransitionTime
+			}
+
+			err := r.updateStatus(ctx, policy, cond, nonCompObj(foundSub, subResFailed.Reason))
+			if err != nil {
+				return nil, fmt.Errorf("error setting the ResolutionFailed status for a Subscription: %w", err)
+			}
+
+			return mergedSub, nil
+		}
+
 		err := r.updateStatus(ctx, policy, matchesCond("Subscription"), matchedObj(foundSub))
 		if err != nil {
 			return nil, fmt.Errorf("error updating the status for a Subscription that matches: %w", err)
@@ -602,6 +634,12 @@ func buildSubscription(
 	subscription.ObjectMeta.Name = spec.Package
 	subscription.ObjectMeta.Namespace = ns
 	subscription.Spec = spec
+
+	// If the policy is in `enforce` mode and the allowed CSVs are restricted,
+	// the InstallPlanApproval will be set to Manual so that upgrades can be controlled.
+	if policy.Spec.RemediationAction.IsEnforce() && len(policy.Spec.Versions) > 0 {
+		subscription.Spec.InstallPlanApproval = operatorv1alpha1.ApprovalManual
+	}
 
 	return subscription, nil
 }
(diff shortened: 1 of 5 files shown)


controllers/operatorpolicy_controller.go
```diff
@@ -790,3 +828,228 @@ func (r *OperatorPolicyReconciler) mergeObjects(
...
+		}
+
+		// A `reason` should be added to most InstallPlans based on the phase.
+		// Note that a `compliant` value will only be added in certain cases.
+		if phase != "" {
+			relObj.Reason = "The InstallPlan is " + phase
+		}
+
+		// handle some special phases
+		switch phase {
```
JeffeyL, line 908: From my understanding `handleCSV` didn't need to consider the `Failed` phase since it took the condition directly from the CSV itself, but in this case should we not report a failed InstallPlan?

JeffeyL (commented):
Looks good overall, a few minor comments that are mostly cosmetic in nature.

controllers/operatorpolicy_controller.go
```diff
@@ -790,3 +828,228 @@ func (r *OperatorPolicyReconciler) mergeObjects(
 
 	return updateNeeded, false, nil
 }
+
+func (r *OperatorPolicyReconciler) handleInstallPlan(
+	ctx context.Context, policy *policyv1beta1.OperatorPolicy, sub *operatorv1alpha1.Subscription,
+) error {
+	if sub == nil {
```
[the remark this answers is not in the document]
JustinKuli: You are right, it needs to do that

controllers/operatorpolicy_controller.go
```diff
@@ -790,3 +828,228 @@ func (r *OperatorPolicyReconciler) mergeObjects(
...
+
+		if err != nil {
+			OpLog.Error(err, "Unable to determine the phase of the related InstallPlan",
+				"InstallPlan.Name", installPlan.GetName())
+
+			// The InstallPlan will be added as unknown
+			phase = ""
+		}
+
+		relObj := policyv1.RelatedObject{
```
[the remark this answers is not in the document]
JustinKuli: That kind of function would probably be nice for consistency. Because of the other stuff using `phase`, it's not a huge benefit, but I don't think it would be any messier.

controllers/operatorpolicy_controller.go
JustinKuli: I should add a comment for this, because it's a great question.

My thinking was: we don't always want to report NonCompliance if there's a `Failed` InstallPlan, because InstallPlans can be left over from previous installations or upgrades. So if the "current" InstallPlan is good, reporting NonCompliant for an old Failed one would be bad/confusing.

But right now, this implementation doesn't even report that a current `Failed` InstallPlan would cause NonCompliance... that's probably not perfect, although I'd expect the Subscription or CSV in this situation would have some status that still causes the policy to be NonCompliant overall. The subscription does point to a current InstallPlan, maybe it can check that one specifically.

controllers/operatorpolicy_controller.go
```diff
@@ -790,3 +828,228 @@ func (r *OperatorPolicyReconciler) mergeObjects(
...
+
+			approvableInstallPlans = append(approvableInstallPlans, installPlan)
+		}
+	}
+
+	if len(approvableInstallPlans) == 0 {
+		// FUTURE: check policy.spec.statusConfig.upgradesAvailable to determine `compliant`.
+		// For now this condition assumes it is set to 'NonCompliant'
+		cond := installPlanUpgradeCond(allUpgradeVersions)
+		cond.Message += " but not allowed by the specified versions in the policy"
```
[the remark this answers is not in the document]
JustinKuli: That would probably be nicer.

controllers/operatorpolicy_controller.go
```diff
@@ -790,3 +828,228 @@ func (r *OperatorPolicyReconciler) mergeObjects(
...
+		if !ok && err == nil {
+			err = errors.New("the phase of the InstallPlan was not found")
+		}
+
+		if err != nil {
+			OpLog.Error(err, "Unable to determine the phase of the related InstallPlan",
+				"InstallPlan.Name", installPlan.GetName())
+
+			// The InstallPlan will be added as unknown
+			phase = ""
```
zyjjay, line 891: Might be worth mentioning in the relevant relatedObj if the `phase` field wasn't found. A message in the `relObj.Reason` field looks like a good place to put it, unless this is required for functionality in another place in the controller.

controllers/operatorpolicy_controller.go
zyjjay: +1, curious about this too.

controllers/operatorpolicy_controller.go
```diff
@@ -603,6 +635,12 @@ func buildSubscription(
 	subscription.ObjectMeta.Namespace = ns
 	subscription.Spec = spec
 
+	// If the policy is in `enforce` mode and the allowed CSVs are restricted,
+	// the InstallPlanApproval will be set to Manual so that upgrades can be controlled.
+	if policy.Spec.RemediationAction.IsEnforce() && len(policy.Spec.Versions) > 0 {
+		subscription.Spec.InstallPlanApproval = operatorv1alpha1.ApprovalManual
```
mprahl, line 641: Do you need to set an `else` here to set automatic in case `policy.Spec.Versions` gets unset later on?

controllers/operatorpolicy_controller.go
```diff
@@ -790,3 +828,211 @@ func (r *OperatorPolicyReconciler) mergeObjects(
...
+			"spec", "clusterServiceVersionNames")
+		if !ok && err != nil {
+			err = errors.New("the clusterServiceVersionNames field of the InstallPlan was not found")
+		}
+
+		if err != nil {
+			OpLog.Error(err, "Unable to determine the csv names of the related InstallPlan",
+				"InstallPlan.Name", installPlan.GetName())
+
+			csvNames = []string{"???"}
```
mprahl, line 957:
```suggestion
			csvNames = []string{"unknown"}
```

controllers/operatorpolicy_controller.go
JustinKuli: If the policy is edited and `policy.Spec.Versions` is removed, then it will just use whatever is in `policy.Spec.Subscription`... in that case either Manual or Automatic is valid.

When a user makes an enforced policy specifying versions, we're interpreting that as them wanting "automatic" updates. And so it would make sense if they set Automatic in the subscription spec. But from OLM's perspective, it needs to be Manual, so that the operator policy controller can approve specific versions (which it will do autonomously).

controllers/operatorpolicy_controller.go
```diff
@@ -790,3 +828,211 @@ func (r *OperatorPolicyReconciler) mergeObjects(
...
+		if err != nil {
+			return fmt.Errorf("error updating status when no relevant InstallPlans were found: %w", err)
+		}
+
+		return nil
+	}
+
+	OpLog := ctrl.LoggerFrom(ctx)
+	relatedInstallPlans := make([]policyv1.RelatedObject, len(ownedInstallPlans))
+	ipsRequiringApproval := make([]unstructured.Unstructured, 0)
```
yiraeChristineKim, line 881: Why this is size 0?

controllers/operatorpolicy_controller.go
```diff
@@ -790,3 +828,211 @@ func (r *OperatorPolicyReconciler) mergeObjects(
...
+	anyInstalling := false
+	currentPlanFailed := false
+
+	// Construct the relevant relatedObjects, and collect any that might be considered for approval
+	for i, installPlan := range ownedInstallPlans {
+		phase, ok, err := unstructured.NestedString(installPlan.Object, "status", "phase")
+		if !ok && err == nil {
```
yiraeChristineKim, line 888: THis should be || ? no?

controllers/operatorpolicy_controller.go
JustinKuli: I just prefer this to an alternate syntax like `ipsRequiringApproval := [] unstructured.Unstructured{}`.

It needs to be length 0 to start so that we can `append` an unknown number of items to it.

controllers/operatorpolicy_controller.go
```diff
@@ -790,3 +828,211 @@ func (r *OperatorPolicyReconciler) mergeObjects(
...
+		return nil
+	}
+
+	approvedVersion := "" // this will only be accurate when there is only one approvable InstallPlan
+	approvableInstallPlans := make([]unstructured.Unstructured, 0)
+
+	for _, installPlan := range ipsRequiringApproval {
+		ipCSVs, ok, err := unstructured.NestedStringSlice(installPlan.Object,
+			"spec", "clusterServiceVersionNames")
+		if !ok && err != nil {
```
dhaiducek, line 982:
```suggestion
		if !ok && err == nil {
```

controllers/operatorpolicy_controller.go
```diff
@@ -790,3 +828,211 @@ func (r *OperatorPolicyReconciler) mergeObjects(
...
+
+		return nil
+	}
+
+	allUpgradeVersions := make([]string, len(ipsRequiringApproval))
+
+	for i, installPlan := range ipsRequiringApproval {
+		csvNames, ok, err := unstructured.NestedStringSlice(installPlan.Object,
+			"spec", "clusterServiceVersionNames")
+		if !ok && err != nil {
```
dhaiducek, line 949:
```suggestion
		if !ok && err == nil {
```

controllers/operatorpolicy_status.go
```diff
@@ -578,6 +594,82 @@ func catalogSrcUnknownObj(catalogName string, catalogNS string) policyv1.Related
...
+// installPlanInstallingCond is a NonCompliant condition with Reason 'InstallPlansInstalling'
+// and message 'a relevant InstallPlan is actively installing'
+var installPlanInstallingCond = metav1.Condition{
+	Type:    installPlanConditionType,
+	Status:  metav1.ConditionFalse,
+	Reason:  "InstallPlansInstalling",
+	Message: "a relevant InstallPlan is actively installing",
+}
+
+// installPlansFineCond is a Compliant condition with Reason 'InstallPlansFine'
```
dhaiducek, line 624: "Fine" is too abstract/non-specific and I feel like it wouldn't be widely understood when translated. Unless this is some alignment with OLM, is there a better word like "Succeeded", "Complete", or "Installed"?

dhaiducek (changes requested):

controllers/operatorpolicy_status.go
JustinKuli: I agree, I had no better ideas when I made it up 😆 

I think originally I thought this would need to cover more cases than it does - so I deliberately avoided something like "Succeeded" because I expected it would cover a case where nothing had been installed yet... but now I think this situation is really just "there are no InstallPlans considered for approval", so I can report that as the condition.

test/e2e/case38_install_operator_test.go
```diff
@@ -576,6 +584,19 @@ var _ = Describe("Test installing an operator from OperatorPolicy", Ordered, fun
 		})
 
 		It("Should generate conditions and relatedobjects of CSV", func() {
+			Eventually(func() string {
+				csv, _ := clientManagedDynamic.Resource(gvrClusterServiceVersion).Namespace(opPolTestNS).
+					Get(context.TODO(), "quay-operator.v3.8.13", metav1.GetOptions{})
```
yiraeChristineKim, line 589:
```suggestion
			Eventually(func(ctx SpecContext) string {
				csv, _ := clientManagedDynamic.Resource(gvrClusterServiceVersion).Namespace(opPolTestNS).
					Get(ctx, "quay-operator.v3.8.13", metav1.GetOptions{})
```

yiraeChristineKim: Good for close test with force

test/e2e/case38_install_operator_test.go
```diff
@@ -635,6 +656,13 @@ var _ = Describe("Test installing an operator from OperatorPolicy", Ordered, fun
 		})
 
 		It("Should generate conditions and relatedobjects of CSV", func() {
+			Eventually(func() []unstructured.Unstructured {
+				csvList, _ := clientManagedDynamic.Resource(gvrClusterServiceVersion).Namespace(opPolTestNS).
+					List(context.TODO(), metav1.ListOptions{})
```
yiraeChristineKim, line 661: same here `func(ctx SpecContext)`

test/e2e/case38_install_operator_test.go
JustinKuli: Good idea

controllers/operatorpolicy_controller.go
JustinKuli: Right. Also, from a quick test on a cluster, it looks like the CRD doesn't require it to be set (even though its description says it's required...) and it seems to default to `Automatic` behavior when it isn't specified.

mprahl (approved):
/hold for other reviews

yiraeChristineKim (approved):

dhaiducek (approved):
LGTM!

PR #204, opened 2024-02-15
Allow configuring a default namespace for operators

JustinKuli:
This adds a command line flag for setting a default namespace for subscriptions/operatorgroups when not specified in an OperatorPolicy spec. This will allow (for example) OpenShift clusters to default to the "openshift-operators" namespace, which is conventionally used for cluster-wide operator installations.

Note to reviewers: the first commit is just re-arranging code blocks to a more consistent order; going through the other 2 commit by commit is probably easier to read.

### DIFF
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -75,8 +75,9 @@ var (
 // OperatorPolicyReconciler reconciles a OperatorPolicy object
 type OperatorPolicyReconciler struct {
 	client.Client
-	DynamicWatcher depclient.DynamicWatcher
-	InstanceName   string
+	DynamicWatcher   depclient.DynamicWatcher
+	InstanceName     string
+	DefaultNamespace string
 }
 
 // SetupWithManager sets up the controller with the Manager and will reconcile when the dynamic watcher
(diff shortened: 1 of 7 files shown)


controllers/operatorpolicy_controller.go
```diff
@@ -193,79 +200,142 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-		if catalogSrc.Status.GRPCConnectionState == nil {
-			// Unknown State
-			err := r.updateStatus(ctx, policy, catalogSourceUnknownCond, catalogSrcUnknownObj(catalogName, catalogNS))
-			if err != nil {
-				return fmt.Errorf("error retrieving the status for a CatalogSource: %w", err)
-			}
+// buildOperatorGroup bootstraps the OperatorGroup spec defined in the operator policy
+// with the apiversion and kind in preparation for resource creation
+func buildOperatorGroup(
+	policy *policyv1beta1.OperatorPolicy, defaultNS string,
```
mprahl, line 282:
```suggestion
	policy *policyv1beta1.OperatorPolicy, namespace string,
```

JustinKuli: It would still use the namespace specified by the user if one is provided, so I feel like `defaultNS` is a more accurate name. But, thinking about it, we should probably stop and alert the user if the specified namespace for the operator group does not match what the subscription will use.

JustinKuli: In which case, the parameter should just be called `namespace` 😆

JustinKuli:
I adjusted the way the relatedObjects is marked as optional, so the controller shouldn't get stuck if the cluster still has the old version of the CRD where the field is required: https://github.com/open-cluster-management-io/config-policy-controller/compare/4187afd2e0bd400b5160e0104c093675187ca9d1..d76d6b5eee52b5f3d7ed8789c0958468e7a8ad75#diff-c1ba6e7a77ddcb50f7ecdd75bd3556d38d1d58414b685c17a84953b947e5cb1cR103-R104

controllers/operatorpolicy_status.go
```diff
@@ -178,7 +182,19 @@ func calculateComplianceCondition(policy *policyv1beta1.OperatorPolicy) metav1.C
 	foundNonCompliant := false
 	messages := make([]string, 0)
 
-	idx, cond := policy.Status.GetCondition(opGroupConditionType)
+	idx, cond := policy.Status.GetCondition(validPolicyConditionType)
+	if idx == -1 {
+		messages = append(messages, "whether the policy is valid or not is unknown")
```
mprahl, line 187: Optional:
```suggestion
		messages = append(messages, "the policy validity is unknown")
```

mprahl (approved):
/hold for the optional minor wording change

PR #206, opened 2024-02-15
Upgrade `controller-gen`

dhaiducek:


deploy/crds/policy.open-cluster-management.io_configurationpolicies.yaml
```diff
@@ -29,14 +27,19 @@ spec:
           API
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation
-              of an object. Servers should convert recognized schemas to the latest
-              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            description: |-
```
JustinKuli, line 30: It's weird they changed the behavior so that it now has newlines. But it shouldn't make a difference for the API, I guess.

JustinKuli (approved):

gparvin (approved):

PR #208, opened 2024-02-20
Emit fewer OperatorPolicy events

JustinKuli:
Previously, every time the controller determined a piece of the status for an OperatorPolicy, it updated the policy status on the cluster, and emitted a compliance event. This made each condition's logic more self- contained, but leads to more API calls than necessary.

Now, each `handle*` function returns whether the status in the cluster needs to be updated, and the controller will make a maximum of one status update pre reconcile. That status update will update all of the conditions and related objects at once. It will also send one compliance event at that point, reflecting the "final" state of the policy.

Some `handle*` functions *should* emit multiple events: for example an enforced policy should report when something is missing, and then separately report if that resource was successfully created. So, some of the functions return "early" conditions, which the controller emits before the end of the reconcile. Compare this to "event batches" in ConfigurationPolicy.

### DIFF
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/validation"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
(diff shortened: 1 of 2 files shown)


controllers/operatorpolicy_controller.go
```diff
@@ -158,59 +159,110 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+func (r *OperatorPolicyReconciler) handleResources(ctx context.Context, policy *policyv1beta1.OperatorPolicy) (
+	earlyComplianceEvents []metav1.Condition, condChanged bool, err error,
+) {
+	OpLog := ctrl.LoggerFrom(ctx)
+
+	earlyComplianceEvents = make([]metav1.Condition, 0)
+
+	desiredSub, desiredOG, changed, err := r.buildResources(policy)
+	condChanged = condChanged || changed
```
mprahl, line 194:
```suggestion
	condChanged = changed
```

controllers/operatorpolicy_controller.go
```diff
@@ -158,59 +159,110 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	return earlyComplianceEvents, condChanged, nil
 }
 
 // buildResources builds desired states for the Subscription and OperatorGroup, and
 // checks if the policy's spec is valid. It returns an error if it couldn't update the
 // validation condition in the policy's status.
-func (r *OperatorPolicyReconciler) buildResources(ctx context.Context, policy *policyv1beta1.OperatorPolicy) (
-	*operatorv1alpha1.Subscription, *operatorv1.OperatorGroup, error,
+func (r *OperatorPolicyReconciler) buildResources(policy *policyv1beta1.OperatorPolicy) (
+	*operatorv1alpha1.Subscription, *operatorv1.OperatorGroup, bool, error,
```
mprahl, line 265: Could you document the returned bool in the function doc comment?

controllers/operatorpolicy_controller.go
```diff
@@ -158,59 +159,110 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	desiredSub, desiredOG, err := r.buildResources(ctx, policy)
+	errs := make([]error, 0)
+
+	conditionsToEmit, conditionChanged, err := r.handleResources(ctx, policy)
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	if conditionChanged {
+		conditionsToEmit = append(conditionsToEmit, calculateComplianceCondition(policy))
```
mprahl, line 170: Could you please add a comment indicating that this is the "final" compliance event to emit of the current policy state? It wasn't immediately obvious to me and that explains why `conditionsToEmit` is named as it is rather than `earlyComplianceEvents`.

mprahl (approved):
/hold for minor comments to be addressed

controllers/operatorpolicy_controller.go
```diff
@@ -158,59 +159,110 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+	for _, cond := range conditionsToEmit {
+		if err := r.emitComplianceEvent(ctx, policy, cond); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return reconcile.Result{}, utilerrors.NewAggregate(errs)
+}
+
+func (r *OperatorPolicyReconciler) handleResources(ctx context.Context, policy *policyv1beta1.OperatorPolicy) (
```
yiraeChristineKim, line 186: Comment Missed? It would be good to me if this had a description

yiraeChristineKim: Looks good to me, very clear to understand! Thanks!

yiraeChristineKim (approved):

controllers/operatorpolicy_controller.go
JustinKuli: Good idea

yiraeChristineKim (approved):

mprahl (approved):
/unhold

PR #209, opened 2024-02-28
Reduce debug logs for operatorpolicy test

JustinKuli:
Instead of printing the log every time the function fails in one of the Eventually runs, it saves a message and will only print it if the test overall fails. This should reduce noise during the tests.


JustinKuli:
As a quick comparison, I see only 8k lines in the E2E tests here, compared to https://github.com/open-cluster-management-io/config-policy-controller/actions/runs/7993535507/job/21829566080 where I see 76k lines.

mprahl (approved):

PR #214, opened 2024-03-01
Filter out unrelated subscription resolution failures

mprahl:
Since OLM includes a subscription error on all subscriptions in the namespace, even if that subscription itself is unaffected, operator policy needs to know if the error is related to the subscription. This adds some regex filtering to parse the messages since the message format is consistent.

Relates:
https://issues.redhat.com/browse/ACM-10195

### DIFF
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"regexp"
 
 	operatorv1 "github.com/operator-framework/api/pkg/operators/v1"
 	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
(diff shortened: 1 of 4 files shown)


controllers/operatorpolicy_controller.go
```diff
@@ -685,6 +704,24 @@ func (r *OperatorPolicyReconciler) handleSubscription(
...
 }
 
+// isApplicableConstraintsNotSatisfiable checks if the ConstraintsNotSatisfiable message includes the input
+// subscription or package. Some examples that it catches:
+// https://github.com/operator-framework/operator-lifecycle-manager/blob/dc0c564f62d526bae0467d53f439e1c91a17ed8a/pkg/controller/registry/resolver/resolver.go#L257-L267
+// - no operators found from catalog %s in namespace %s referenced by subscription %s
+// - no operators found in package %s in the catalog referenced by subscription %s
+// - no operators found in channel %s of package %s in the catalog referenced by subscription %s
+// - no operators found with name %s in channel %s of package %s in the catalog referenced by subscription %s
+func isApplicableConstraintsNotSatisfiable(subscriptionName, packageName, message string) (bool, error) {
```
JustinKuli, line 714: This function name is quite a mouthful. I would suggest `are` instead of `is`, to match the `Constraints` since that is plural, but that doesn't really make it feel better. What about something like `messageIncludesSubscription`?

JustinKuli (commented):
Seems pretty good! It's just quite a mouthful.

JustinKuli (dismissed):
/hold

LGTM. Holding in case you want anyone else to look.

JustinKuli (approved):

PR #216, opened 2024-03-11
Run framework tests on OCM PRs

zyjjay:
ref: https://issues.redhat.com/browse/ACM-4811

### DIFF
--- a/.github/workflows/framework-kind.yml
+++ b/.github/workflows/framework-kind.yml
@@ -1,0 +1,14 @@
+name: Framework
+
+on:
+  push:
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  fw-kind-tests:
+    name: KinD
+    uses: stolostron/governance-policy-framework/.github/workflows/kind.yml@main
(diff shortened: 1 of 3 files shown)


.github/workflows/framework-kind.yml
```diff
@@ -0,0 +1,14 @@
...
+    tags:
+      - v0.**
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  fw-kind-tests:
+    name: KinD
+    uses: stolostron/governance-policy-framework/.github/workflows/kind.yml@main
```
dhaiducek, line 14: nit: trailing newline

```suggestion
    uses: stolostron/governance-policy-framework/.github/workflows/kind.yml@main

```

.github/workflows/framework-kind.yml
```diff
@@ -0,0 +1,14 @@
+name: Framework
+
+on:
+  push:
+    tags:
+      - v0.**
```
dhaiducek, line 6: Good idea to run the tests on tag pushes! Is there a reason for this RegEx or could we make it more generic?

```suggestion
      - v[0-9]+.[0-9]+.[0-9]+
```

zyjjay: I didn't have anything particular in mind so I just used something that works. I'm open to making it more generic!

dhaiducek (approved):
Woohoo! Thanks for all your work on this, @zyjjay!

PR #220, opened 2024-03-25
Fix constraints not satisfiable message causing compliance flood

JeffeyL:
`ConstraintsNotSatisfiable` message types in the subscription status are formatted `"constraints not satisfiable: clause1, clause2, clause3 ..."`. However, the order of the clauses is nondeterministic, which causes compliance event flapping as each update leads to a different ordering of messages, which in turn triggers another update. This was fixed by sorting the message prior to evaluation.

Ref: https://issues.redhat.com/browse/ACM-10204

### DIFF
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -11,6 +11,8 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
+	"slices"
+	"strings"
 
 	operatorv1 "github.com/operator-framework/api/pkg/operators/v1"
 	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
@@ -653,19 +655,17 @@ func (r *OperatorPolicyReconciler) handleSubscription(
 			}
 
 			if includesSubscription {
-				cond := metav1.Condition{
-					Type:    subConditionType,
-					Status:  metav1.ConditionFalse,
-					Reason:  subResFailed.Reason,
-					Message: subResFailed.Message,
+				// a "constraints not satisfiable" message has nondeterministic clauses, and thus
+				// need to be sorted in order to check that they are not duplicates of the current message.
+				if constraintMessageMatch(policy, &subResFailed) {
+					return mergedSub, nil, false, nil
 				}
 
-				if subResFailed.LastTransitionTime != nil {
-					cond.LastTransitionTime = *subResFailed.LastTransitionTime
-				}
-
-				return mergedSub, nil, updateStatus(policy, cond, nonCompObj(foundSub, subResFailed.Reason)), nil
+				return mergedSub, nil, updateStatus(
+					policy, subResFailedCond(subResFailed), nonCompObj(foundSub, subResFailed.Reason)), nil
 			}
+
+			return mergedSub, nil, false, nil
 		}
 
 		return mergedSub, nil, updateStatus(policy, matchesCond("Subscription"), matchedObj(foundSub)), nil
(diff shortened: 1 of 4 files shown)


controllers/operatorpolicy_controller.go
```diff
@@ -727,6 +727,43 @@ func messageIncludesSubscription(subscription *operatorv1alpha1.Subscription, me
...
 }
 
+// constraintMessageMatch checks if the ConstraintsNotSatisfiable message is actually different
+// from the old one by sorting the clauses of the message
+func constraintMessageMatch(policy *policyv1beta1.OperatorPolicy, cond *operatorv1alpha1.SubscriptionCondition) bool {
+	const cnfPrefix = "constraints not satisfiable: "
+
+	var policyMessage, subMessage string
+
+	for _, cond := range policy.Status.Conditions {
```
JustinKuli, line 737: This `cond` shadows the function parameter. Maybe renaming it to `statusCond` would be better.

controllers/operatorpolicy_controller.go
```diff
@@ -727,6 +727,43 @@ func messageIncludesSubscription(subscription *operatorv1alpha1.Subscription, me
...
+	}
+
+	if strings.Contains(cond.Message, cnfPrefix) {
+		subMessage = cond.Message
+	} else {
+		return false
+	}
+
+	policyMessage = strings.TrimPrefix(policyMessage, cnfPrefix)
+	subMessage = strings.TrimPrefix(subMessage, cnfPrefix)
```
JustinKuli, line 754: Nit, saves kind of saves a branch:
```suggestion
	if !strings.Contains(cond.Message, cnfPrefix) {
		return false
	}

	policyMessage = strings.TrimPrefix(policyMessage, cnfPrefix)
	subMessage = strings.TrimPrefix(cond.Message, cnfPrefix)
```

JustinKuli (commented):
Looks good overall! Just want to help the readability a little bit

controllers/operatorpolicy_controller.go
```diff
@@ -727,6 +727,37 @@ func messageIncludesSubscription(subscription *operatorv1alpha1.Subscription, me
...
+
+	var policyMessage, subMessage string
+
+	for _, statusCond := range policy.Status.Conditions {
+		if strings.Contains(statusCond.Message, cnfPrefix) {
+			policyMessage = statusCond.Message
+		}
+	}
+
+	if policyMessage == "" || !strings.Contains(cond.Message, cnfPrefix) {
```
JustinKuli, line 743: Even better!

JustinKuli (approved):

JeffeyL:
Some tests might be flaky after this change... I'll check it out

JeffeyL:
Turns out I had to add the string sorting logic to the `check` function in the tests as well, should be fixed now!

JustinKuli (approved):

controllers/operatorpolicy_controller.go
```diff
@@ -727,6 +727,37 @@ func messageIncludesSubscription(subscription *operatorv1alpha1.Subscription, me
...
+// constraintMessageMatch checks if the ConstraintsNotSatisfiable message is actually different
+// from the old one by sorting the clauses of the message
+func constraintMessageMatch(policy *policyv1beta1.OperatorPolicy, cond *operatorv1alpha1.SubscriptionCondition) bool {
+	const cnfPrefix = "constraints not satisfiable: "
+
+	var policyMessage, subMessage string
+
+	for _, statusCond := range policy.Status.Conditions {
+		if strings.Contains(statusCond.Message, cnfPrefix) {
+			policyMessage = statusCond.Message
```
mprahl, line 739: Is there a reason not to break out of the loop after this assignment?

PR #221, opened 2024-03-28
Handle policy recreation race condition

mprahl:
When the configuration policy is recreated after the List call completes in the loop in the PeriodicallyExecConfigPolicies method, the old configuration policy status gets set on the new configuration policy which can lead to a compliance event not being set on the replicated parent policy.

Relates:
https://issues.redhat.com/browse/ACM-10500

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -2982,6 +2982,7 @@ func (r *ConfigurationPolicyReconciler) updatePolicyStatus(
 		"Updating configurationPolicy status", "status", policy.Status.ComplianceState, "policy", policy.GetName(),
 	)
 
+	evaluatedUID := policy.UID
 	updatedStatus := policy.Status
 
 	maxRetries := 3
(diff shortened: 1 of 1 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -2991,6 +2992,17 @@ func (r *ConfigurationPolicyReconciler) updatePolicyStatus(
 			log.Info(fmt.Sprintf("Failed to refresh policy; using previously fetched version: %s", err))
 		} else {
 			policy.Status = updatedStatus
+
+			// If the policy was recreated after it was retrieved from the cache, there could be no compliance
+			// set on the parent policy. Updating the status here may prevent a compliance event from being generated on
+			// future evaluations so skipping the status update is very important. If sendEvent was true, there is still
+			// value to send the compliance event to the parent policy so that this history is not lost before the
+			// policy was recreated.
```
JustinKuli, line 3000: Maybe: "If the UID has changed, then the policy has been deleted and created again. Do not update the status, because it was calculated based on a previous version. If sendEvent is true, that event might be useful and it can be emitted. By leaving the status blank, the policy will be reevaluated and send a new event.

JustinKuli (approved):
LGTM

PR #222, opened 2024-04-02
MustNotHave mode for OperatorPolicy

JustinKuli:
For https://issues.redhat.com/browse/ACM-9287


JustinKuli:
Tests passed twice, I'm finally feeling good about this 😆

api/v1beta1/operatorpolicy_types.go
```diff
@@ -36,20 +37,78 @@ const (
...
+
+	//+kubebuilder:default=Delete
+	//+kubebuilder:validation:Enum=Keep;Delete
+	// Specifies whether to delete the Subscription; defaults to 'Delete'
 	Subscriptions RemovalAction `json:"subscriptions,omitempty"`
-	// Kind ClusterServiceVersion
+
+	//+kubebuilder:default=Delete
+	//+kubebuilder:validation:Enum=Keep;Delete
+	// Specifies whether to delete the ClusterServiceVersion; defaults to 'Delete'
```
yiraeChristineKim, line 66: Can we add when this happens?

api/v1beta1/operatorpolicy_types.go
```diff
@@ -36,20 +37,78 @@ const (
...
+
+	//+kubebuilder:default=Keep
+	//+kubebuilder:validation:Enum=Keep;Delete
+	// Specifies whether to delete any CustomResourceDefinitions associated with the operator;
+	// defaults to 'Keep' because deleting them should be done deliberately
 	CRDs RemovalAction `json:"customResourceDefinitions,omitempty"`
-	// Kind APIServiceDefinitions
-	APIServiceDefinitions RemovalAction `json:"apiServiceDefinitions,omitempty"`
+
+	// Future?
```
yiraeChristineKim, line 81: TODO?!!

test/e2e/case38_install_operator_test.go
```diff
@@ -1195,6 +1200,103 @@ var _ = Describe("Test installing an operator from OperatorPolicy", Ordered, fun
...
 	})
+	Describe("Testing CustomResourceDefinition reporting", Ordered, func() {
+		const (
+			opPolYAML = "../resources/case38_operator_install/operator-policy-authorino.yaml"
+			opPolName = "oppol-authorino"
+		)
+		BeforeAll(func() {
+			utils.Kubectl("delete", "crd", "--selector=olm.managed=true")
+			utils.Kubectl("create", "ns", opPolTestNS)
+			DeferCleanup(func() {
```
yiraeChristineKim, line 1211: I am not 100% sure but DeferCleanup func will execute right after BeforeAll...

test/e2e/case38_install_operator_test.go
```diff
@@ -1331,4 +1433,836 @@ var _ = Describe("Test installing an operator from OperatorPolicy", Ordered, fun
...
+				false,
+				[]policyv1.RelatedObject{{
+					Object: policyv1.ObjectResource{
+						Kind:       "OperatorGroup",
+						APIVersion: "operators.coreos.com/v1",
+						Metadata: policyv1.ObjectMetadata{
+							Namespace: opPolTestNS,
+						},
+					},
+					Reason: "The OperatorGroup is attached to a mustnothave policy, but does not need to be removed",
```
yiraeChristineKim, line 1769: Why this doesn't have compliant?

test/e2e/case38_install_operator_test.go
```diff
@@ -1331,4 +1433,836 @@ var _ = Describe("Test installing an operator from OperatorPolicy", Ordered, fun
...
+				opPolTestNS, true, eventuallyTimeout)
+			utils.GetWithTimeout(clientManagedDynamic, gvrSubscription, subName,
+				opPolTestNS, true, eventuallyTimeout)
+			utils.GetWithTimeout(clientManagedDynamic, gvrCRD, "quayecosystems.redhatcop.redhat.io",
+				"", true, eventuallyTimeout)
+			utils.GetWithTimeout(clientManagedDynamic, gvrCRD, "quayregistries.quay.redhat.com",
+				"", true, eventuallyTimeout)
+		})
+		It("Should not remove anything when enforced while set to Keep everything", func(ctx SpecContext) {
+			// Change the removal behaviors from Delete to Keep
```
yiraeChristineKim, line 1908: Maybe keep to delete?

test/e2e/case38_install_operator_test.go
```diff
@@ -1331,4 +1433,836 @@ var _ = Describe("Test installing an operator from OperatorPolicy", Ordered, fun
...
+			utils.GetWithTimeout(clientManagedDynamic, gvrClusterServiceVersion, "quay-operator.v3.8.13",
+				opPolTestNS, true, eventuallyTimeout)
+			utils.GetWithTimeout(clientManagedDynamic, gvrSubscription, subName,
+				opPolTestNS, true, eventuallyTimeout)
+			utils.GetWithTimeout(clientManagedDynamic, gvrCRD, "quayecosystems.redhatcop.redhat.io",
+				"", true, eventuallyTimeout)
+			utils.GetWithTimeout(clientManagedDynamic, gvrCRD, "quayregistries.quay.redhat.com",
+				"", true, eventuallyTimeout)
+		})
+		It("Should not remove anything when enforced while set to Keep everything", func(ctx SpecContext) {
```
yiraeChristineKim, line 1907: `It `title  same as above

controllers/operatorpolicy_controller.go
```diff
@@ -929,55 +1130,170 @@ func (r *OperatorPolicyReconciler) handleInstallPlan(
...
+	for _, csv := range csvList {
+		if csv.GetName() == sub.Status.InstalledCSV {
+			matchedCSV := operatorv1alpha1.ClusterServiceVersion{}
+
+			err = runtime.DefaultUnstructuredConverter.FromUnstructured(csv.UnstructuredContent(), &matchedCSV)
+			if err != nil {
+				return nil, nil, false, err
+			}
+
+			foundCSV = &matchedCSV
```
mprahl, line 1211:
```suggestion
			foundCSV = &matchedCSV
			
			break
```

controllers/operatorpolicy_controller.go
```diff
@@ -1029,10 +1345,90 @@ func (r *OperatorPolicyReconciler) handleDeployment(
...
+	// Same condition for musthave and mustnothave
+	if len(crdList) == 0 {
+		return nil, updateStatus(policy, noCRDCond, noExistingCRDObj), nil
+	}
+
+	relatedCRDs := make([]policyv1.RelatedObject, len(crdList))
+
+	if policy.Spec.ComplianceType.IsMustHave() {
+		for i := range crdList {
+			relatedCRDs[i] = matchedObj(&crdList[i])
```
mprahl, line 1374: Nit: I believe this is the most efficient way since it avoids initializing policyv1.RelatedObject objects in the slice that go unused. This suggestion just sets the capacity of the underlying array. 
```suggestion
	relatedCRDs := make([]policyv1.RelatedObject, 0, len(crdList))

	if policy.Spec.ComplianceType.IsMustHave() {
		for i := range crdList {
			relatedCRDs = append(relatedCRDs, matchedObj(&crdList[i]))
```

api/v1beta1/operatorpolicy_types.go
mprahl, line 81: Let's just remove this

controllers/operatorpolicy_controller.go
```diff
@@ -576,6 +608,87 @@ func (r *OperatorPolicyReconciler) handleOpGroup(
...
+		return nil, changed, nil
+	}
+
+	foundOpGroupName := ""
+
+	for _, opGroup := range foundOpGroups {
+		emptyNameMatch := desiredOpGroup.Name == "" && opGroup.GetGenerateName() == desiredOpGroup.GenerateName
+
+		if opGroup.GetName() == desiredOpGroup.Name || emptyNameMatch {
+			foundOpGroupName = opGroup.GetName()
```
mprahl, line 630:
```suggestion
			foundOpGroupName = opGroup.GetName()

			break
```

api/v1beta1/operatorpolicy_types.go
```diff
@@ -36,20 +37,78 @@ const (
...
+func (ra RemovalAction) IsDeleteIfUnused() bool {
+	return strings.EqualFold(string(ra), string(DeleteIfUnused))
+}
+
 type RemovalBehavior struct {
-	// Kind OperatorGroup
+	//+kubebuilder:default=DeleteIfUnused
+	//+kubebuilder:validation:Enum=Keep;Delete;DeleteIfUnused
+	// Specifies whether to delete the OperatorGroup; defaults to 'DeleteIfUnused' which
+	// will only delete the OperatorGroup if there is not another Subscription using it.
```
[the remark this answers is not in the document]
JustinKuli: Right, it means "Delete, regardless of any consequences". Kind of like a mustnothave configuration policy.

api/v1beta1/operatorpolicy_types.go
JustinKuli: In the CRD description for `removalBehavior`, it says "RemovalBehavior defines what resources will be removed by enforced mustnothave policies ..."

Is that what you meant? Or what's another place to put it?

api/v1beta1/operatorpolicy_types.go
JustinKuli: 😆 the whole feature was a TODO ...

This type / option was something I thought would be good during the design, but I haven't found anyone asking for it, or any examples of it yet, so I want to hold off on it until we can understand it better.

test/e2e/case38_install_operator_test.go
JustinKuli: Good catch!

test/e2e/case38_install_operator_test.go
JustinKuli: Yep, you're right.

controllers/operatorpolicy_controller.go
JustinKuli: 👍 Using `append` is probably better in general in case anything would ever exclude items in the future. Unlikely in this spot, but a good habit.

controllers/operatorpolicy_controller.go
```diff
@@ -791,16 +974,32 @@ func (r *OperatorPolicyReconciler) handleInstallPlan(
...
+
+func (r *OperatorPolicyReconciler) musthaveInstallPlan(
+	ctx context.Context,
+	policy *policyv1beta1.OperatorPolicy,
+	sub *operatorv1alpha1.Subscription,
+	ownedInstallPlans []unstructured.Unstructured,
+) (bool, error) {
 	OpLog := ctrl.LoggerFrom(ctx)
 	relatedInstallPlans := make([]policyv1.RelatedObject, len(ownedInstallPlans))
 	ipsRequiringApproval := make([]unstructured.Unstructured, 0)
```
mprahl, line 999:
```suggestion
	relatedInstallPlans := make([]policyv1.RelatedObject, 0, len(ownedInstallPlans))
	ipsRequiringApproval := make([]unstructured.Unstructured)
```

controllers/operatorpolicy_controller.go
```diff
@@ -929,55 +1132,172 @@ func (r *OperatorPolicyReconciler) handleInstallPlan(
 	return updateStatus(policy, installPlanApprovedCond(approvedVersion), relatedInstallPlans...), nil
 }
 
+func (r *OperatorPolicyReconciler) mustnothaveInstallPlan(
+	ctx context.Context,
+	policy *policyv1beta1.OperatorPolicy,
+	ownedInstallPlans []unstructured.Unstructured,
+) ([]metav1.Condition, bool, error) {
+	relatedInstallPlans := make([]policyv1.RelatedObject, len(ownedInstallPlans))
```
mprahl, line 1140:
```suggestion
	relatedInstallPlans := make([]policyv1.RelatedObject, 0, len(ownedInstallPlans))
```

controllers/operatorpolicy_controller.go
```diff
@@ -929,55 +1132,172 @@ func (r *OperatorPolicyReconciler) handleInstallPlan(
...
+	}
+
+	earlyConds := []metav1.Condition{}
+
+	if changed {
+		earlyConds = append(earlyConds, calculateComplianceCondition(policy))
+	}
+
+	for i := range ownedInstallPlans {
+		err := r.Delete(ctx, &ownedInstallPlans[i])
```
mprahl, line 1167: Optional: Is there likely to be a lot of these install plans around? If so, it may be worth doing this concurrently since deletes can take a few seconds.

controllers/operatorpolicy_controller.go
```diff
@@ -929,55 +1132,172 @@ func (r *OperatorPolicyReconciler) handleInstallPlan(
...
+		changed := updateStatus(policy, missingNotWantedCond("ClusterServiceVersion"),
+			missingNotWantedCSVObj(namespace))
+
 		return nil, changed, nil
 	}
 
-	// Check CSV most recent condition
-	unstructured := foundCSV.UnstructuredContent()
-	var csv operatorv1alpha1.ClusterServiceVersion
+	relatedCSVs := make([]policyv1.RelatedObject, len(csvList))
```
mprahl, line 1249: Nit: Another spot :smile: 
```suggestion
	relatedCSVs := make([]policyv1.RelatedObject, 0, len(csvList))
```

controllers/operatorpolicy_controller.go
```diff
@@ -929,55 +1132,172 @@ func (r *OperatorPolicyReconciler) handleInstallPlan(
...
-		return nil, false, err
+	if policy.Spec.RemovalBehavior.ApplyDefaults().CSVs.IsKeep() {
+		for i := range csvList {
+			relatedCSVs[i] = leftoverObj(&csvList[i])
+		}
+
+		return nil, updateStatus(policy, keptCond("ClusterServiceVersion"), relatedCSVs...), nil
+	}
+
+	for i := range csvList {
```
mprahl, line 1259: Is it possible to have more than one CSV?

controllers/operatorpolicy_status.go
```diff
@@ -394,7 +415,29 @@ func missingWantedCond(kind string) metav1.Condition {
...
 
-// createdCond returns a Compliant condition, with a Reason like'____Created',
+// missingNotWantedCond returns a Compliant condition with a Reason like '____Excluded'
+// and a Message like 'the ____ checked by the mustnothave policy was not found'
+func missingNotWantedCond(kind string) metav1.Condition {
+	return metav1.Condition{
+		Type:    condType(kind),
+		Status:  metav1.ConditionTrue,
+		Reason:  kind + "Excluded",
+		Message: "the " + kind + " checked by the mustnothave policy was not found",
```
mprahl, line 425: I think this is easier to read.
```suggestion
		Message: "the " + kind + " checked by the must not have policy was not found",
```

controllers/operatorpolicy_status.go
```diff
@@ -394,7 +415,29 @@ func missingWantedCond(kind string) metav1.Condition {
...
+}
+
+// foundNotWantedCond returns a NonCompliant condition with a Reason like '____Included'
+// and a Message like 'the ____ checked by the mustnothave policy was found'
+func foundNotWantedCond(kind string) metav1.Condition {
+	return metav1.Condition{
+		Type:    condType(kind),
+		Status:  metav1.ConditionFalse,
+		Reason:  kind + "Included",
+		Message: "the " + kind + " checked by the mustnothave policy was found",
```
mprahl, line 436:
```suggestion
		Message: "the " + kind + " checked by the must not have policy was found",
```

controllers/operatorpolicy_status.go
mprahl: Instead of "checked" does "managed" sound better?

controllers/operatorpolicy_status.go
```diff
@@ -405,6 +448,30 @@ func createdCond(kind string) metav1.Condition {
...
 }
 
+// deletedCond returns a Compliant condition, with a Reason like '____Deleted',
+// and a Message like 'the ____ checked by the mustnothave policy was deleted'
+func deletedCond(kind string) metav1.Condition {
+	return metav1.Condition{
+		Type:    condType(kind),
+		Status:  metav1.ConditionTrue,
+		Reason:  kind + "Deleted",
+		Message: "the " + kind + " checked by the mustnothave policy was deleted",
```
mprahl, line 458:
```suggestion
		Message: "the " + kind + " checked by the must not have policy was deleted",
```

controllers/operatorpolicy_status.go
```diff
@@ -489,6 +560,17 @@ func subResFailedCond(subFailedCond operatorv1alpha1.SubscriptionCondition) meta
 	return cond
 }
 
+// irrelevantCond returns a Compliant condition, with a Reason like '____Irrelevant',
```
mprahl, line 563: Is NotApplicable a little better?

controllers/operatorpolicy_controller.go
JustinKuli: I've now done a search on `make([]` to find these instances 😆 thanks for noticing them.

controllers/operatorpolicy_controller.go
JustinKuli: OLM removes some old InstallPlans eventually, I think I read somewhere that it tries to keep a max of 5?

I don't think that waiting for several deletes in series would be a major inconvenience. I think I prefer the simplicity here.

JustinKuli: But this makes me wonder about situations where the `Delete` might get stuck (possibly indefinitely) waiting for finalizers or something. I think controller-runtime has a default deadline on the context it passes the Reconcile call, but I would bet it's longer than we'd really want. It might be worth looking at all of these Deletes, specifying a lower timeout, and specially reporting the status when they "get stuck". I would prefer to handle that in a separate PR.

controllers/operatorpolicy_controller.go
JustinKuli: In general, there might be more than one CSV: I think if one fails it might stay around even while another one is created. However, it's not clear to me if/when OLM will remove the label from an old CSV. So I would say it's very likely there would only be one CSV in this list, but it should handle the case where there are more.

controllers/operatorpolicy_controller.go
mprahl: I don't think the delete API request gets stuck since I don't think it's blocking like `kubectl`. It does seem like the delete calls will likely be repeated every reconcile. Perhaps you can guard deletes on if the deletion timestamp is set to avoid such API requests?

mprahl:
@JustinKuli does the order of operations matter? To me, it seems like we should wait for the subscription to be fully deleted before deleting anything else so that we don't conflict with OLM. What do you think?

controllers/operatorpolicy_status.go
JustinKuli: Finding the right phrasing is hard. Maybe we need to keep thinking about it, I don't think these suggestions are solidly better.

I prefer keeping "mustnothave", I think it's more searchable in the docs to find exactly what this behavior is. Separate words does look better, but doesn't reflect what the actual option is in the resource.

I think "managed" could imply that the policy created/edited the resource at some point, which is not necessarily the case.

We should probably also think about whether "Excluded" is good. I just wanted a different Reason than the other condition where not found is noncompliant.

controllers/operatorpolicy_status.go
JustinKuli: That does sound better

test/e2e/case38_install_operator_test.go
```diff
@@ -1195,6 +1200,103 @@ var _ = Describe("Test installing an operator from OperatorPolicy", Ordered, fun
 			)
 		})
 	})
+	Describe("Testing CustomResourceDefinition reporting", Ordered, func() {
+		const (
+			opPolYAML = "../resources/case38_operator_install/operator-policy-authorino.yaml"
+			opPolName = "oppol-authorino"
+		)
+		BeforeAll(func() {
+			utils.Kubectl("delete", "crd", "--selector=olm.managed=true")
```
mprahl, line 1209: Can this be more specific?

JustinKuli:
> @JustinKuli does the order of operations matter? To me, it seems like we should wait for the subscription to be fully deleted before deleting anything else so that we don't conflict with OLM. What do you think?

@mprahl that could be nice. I think that even if stuff was deleted in the wrong order, eventually this would get to the correct state, but it could be a lot of churn on the cluster: for example if the CSV was deleted before the Subscription was removed, OLM would probably re-create the CSV. The policy would reconcile several times and (probably) eventually get to a stable status.

This also makes me think that we might want to validate some relationships between the removalBehavior options: if Subscriptions is set to Keep, than it should probably be an error to set ClusterServiceVersions to Delete, since that would cause a loop with OLM.

controllers/operatorpolicy_status.go
```diff
@@ -394,7 +415,29 @@ func missingWantedCond(kind string) metav1.Condition {
...
 }
 
-// createdCond returns a Compliant condition, with a Reason like'____Created',
+// missingNotWantedCond returns a Compliant condition with a Reason like '____Excluded'
+// and a Message like 'the ____ checked by the mustnothave policy was not found'
+func missingNotWantedCond(kind string) metav1.Condition {
+	return metav1.Condition{
+		Type:    condType(kind),
+		Status:  metav1.ConditionTrue,
+		Reason:  kind + "Excluded",
```
mprahl, line 424: What about this?
```suggestion
		Reason:  kind + "NotPresent",
```

controllers/operatorpolicy_status.go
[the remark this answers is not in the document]
JustinKuli: Sure

controllers/operatorpolicy_status.go
JustinKuli: Updated this. "Included" is also now "Present" and the deleted and kept conditions are updated so that their messages are shorter and avoid the "checked by the mustnothave" language.

controllers/operatorpolicy_controller.go
mprahl: @JustinKuli what are your thoughts on this thread? Are you planning on addressing any of it in this PR or a separate PR?

JustinKuli: Guarding deletes if a deletion timestamp is set would be good. I think I want to do it in a separate PR.

mprahl (approved):
/hold for other reviews and assuming that another PR will change deletes to be guarded by if a deletion timestamp is not set.

controllers/operatorpolicy_status.go
```diff
@@ -600,13 +680,33 @@ func buildCSVCond(csv *operatorv1alpha1.ClusterServiceVersion) metav1.Condition
 	}
 }
 
+// noCSVCond is a NonCompliant condition with Reason 'RelevantCSVNotFound'
 var noCSVCond = metav1.Condition{
 	Type:    csvConditionType,
 	Status:  metav1.ConditionFalse,
```
zyjjay, line 686: Maybe I missed something, but wouldn't this imply that the reason "RelevantCSVNotFound" is false here given the status of the condition?

api/v1beta1/operatorpolicy_types.go
```diff
@@ -36,20 +37,75 @@ const (
...
+	// Specifies whether to delete any InstallPlans associated with the operator; defaults
+	// to 'Keep' because those objects are only for history
 	InstallPlan RemovalAction `json:"installPlans,omitempty"`
-	// Kind CustomResourceDefinitions
+
+	//+kubebuilder:default=Keep
+	//+kubebuilder:validation:Enum=Keep;Delete
+	// Specifies whether to delete any CustomResourceDefinitions associated with the operator;
+	// defaults to 'Keep' because deleting them should be done deliberately
 	CRDs RemovalAction `json:"customResourceDefinitions,omitempty"`
```
zyjjay, line 79: Do these CRDs refer to CRDs created by OLM as a prerequisite for installing the desired operator?

controllers/operatorpolicy_status.go
JustinKuli: Conditions and reasons can be confusing, and that's not limited to just our use of them. I think the status goes with the type, not the reason.

So I read them like `<Type>? <Status>, because <Reason>. Extra info: <Message>`

So in this case, "ClusterServiceVersionCompliant? False, because RelevantCSVNotFound. Extra info: A relevant installed ClusterServiceVersion could not be found"

api/v1beta1/operatorpolicy_types.go
JustinKuli: They are the CRDs created by OLM, as defined in the PackageManifest/ClusterServiceVersion for the operator.

controllers/operatorpolicy_status.go
zyjjay: Right, that makes sense.

PR #223, opened 2024-04-03
Fix handling of undefined fields for `mustonlyhave`

dhaiducek:
If a field exists in the existing object but wasn't defined for a `mustonlyhave` policy, we need to check the existing object's fields and set them accordingly, particularly to set them to null.

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -2750,7 +2750,11 @@ func handleKeys(
 	mdCompType string,
 	zeroValueEqualsNil bool,
 ) (throwSpecViolation bool, message string, updateNeeded bool, statusMismatch bool) {
+	handledKeys := map[string]bool{}
+
+	// Iterate over keys of the desired object to compare with the existing object on the cluster
 	for key := range desiredObj.Object {
+		handledKeys[key] = true
 		isStatus := key == "status"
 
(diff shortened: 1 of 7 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -2798,6 +2802,36 @@ func handleKeys(
...
+
+			// If the mergedObj is nil, we'll use the value to erase it from the existing object.
+			if skipped {
+				continue
+			}
+
+			if keyUpdateNeeded {
+				existingObj.Object[key] = mergedObj
+				updateNeeded = true
+			}
```
mprahl, line 2831: Can this be simplified to the following? I'm think that if the keys weren't present on `desiredObj.Object` but exist on `existingObj.Object`, then I think you always just want to delete the key.
```suggestion
			delete(existingObj.Object, key)
			updateNeeded = true
```

dhaiducek: Ah, true. That's way cleaner--thanks!

dhaiducek: I guess I was thinking of inheriting some of the logic in `handleSingleKey`, but deleting it seems right. The dry-run ought to hopefully take care of any edge cases.

mprahl (approved):

PR #228, opened 2024-04-11
Set default subscription values when not specified

mprahl:
This sets the default channel, source, and sourceNamespace when not provided in the policy.

Relates:
https://issues.redhat.com/browse/ACM-10561

### DIFF
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/client-go/dynamic"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -85,11 +86,18 @@ var (
 		Version: "v1alpha1",
 		Kind:    "InstallPlan",
 	}
+	packageManifestGVR = schema.GroupVersionResource{
+		Group:    "packages.operators.coreos.com",
+		Version:  "v1",
+		Resource: "packagemanifests",
+	}
+	ErrPackageManifest = errors.New("")
 )
 
 // OperatorPolicyReconciler reconciles a OperatorPolicy object
 type OperatorPolicyReconciler struct {
 	client.Client
+	DynamicClient    dynamic.Interface
 	DynamicWatcher   depclient.DynamicWatcher
 	InstanceName     string
 	DefaultNamespace string
@@ -209,7 +217,7 @@ func (r *OperatorPolicyReconciler) handleResources(ctx context.Context, policy *
 
 	earlyComplianceEvents = make([]metav1.Condition, 0)
 
-	desiredSub, desiredOG, changed, err := r.buildResources(policy)
+	desiredSub, desiredOG, changed, err := r.buildResources(ctx, policy)
 	condChanged = changed
 
 	if err != nil {
@@ -297,13 +305,28 @@ func (r *OperatorPolicyReconciler) handleResources(ctx context.Context, policy *
 //   - an error if an API call failed
 //
 // The built objects can be used to find relevant objects for a 'mustnothave' policy.
-func (r *OperatorPolicyReconciler) buildResources(policy *policyv1beta1.OperatorPolicy) (
+func (r *OperatorPolicyReconciler) buildResources(ctx context.Context, policy *policyv1beta1.OperatorPolicy) (
 	*operatorv1alpha1.Subscription, *operatorv1.OperatorGroup, bool, error,
 ) {
+	var returnedErr error
+
 	validationErrors := make([]error, 0)
 
 	sub, subErr := buildSubscription(policy, r.DefaultNamespace)
-	if subErr != nil {
+	if subErr == nil {
+		err := r.applySubscriptionDefaults(ctx, sub)
+		if err != nil {
+			sub = nil
+
+			// If it's a PackageManifest API error, then that means it should be returned for the Reconcile method
+			// to requeue the request. This is to workaround the PackageManifest API not supporting watches.
+			if errors.Is(err, ErrPackageManifest) {
+				returnedErr = err
+			}
+
+			validationErrors = append(validationErrors, err)
+		}
+	} else {
 		validationErrors = append(validationErrors, subErr)
 	}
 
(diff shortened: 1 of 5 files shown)


controllers/operatorpolicy_controller.go
```diff
@@ -332,6 +346,74 @@ func (r *OperatorPolicyReconciler) buildResources(policy *policyv1beta1.Operator
...
+) error {
+	subSpec := subscription.Spec
+
+	defaultsNeeded := subSpec.Channel == "" || subSpec.CatalogSource == "" || subSpec.CatalogSourceNamespace == ""
+
+	if !defaultsNeeded {
+		return nil
+	}
+
+	// PackageManifests come from an API server and not a Kubernetes resource, so the DynamicWatcher can't be used since
```
JustinKuli, line 362: Oh! Is that why you were saying the packagemanifest is inconsistent when you have multiple catalogsources for the same package? At least the resource isn't constantly changing in etcd I guess...

controllers/operatorpolicy_controller.go
```diff
@@ -332,6 +346,74 @@ func (r *OperatorPolicyReconciler) buildResources(policy *policyv1beta1.Operator
...
+
+		log.Error(err, "Failed to get the PackageManifest", "name", subSpec.Package)
+
+		return errors.New(
+			"the subscription defaults could not be determined because the PackageManifest API returned an error",
+		)
+	}
+
+	catalog, _, _ := unstructured.NestedString(packageManifest.Object, "status", "catalogSource")
+	catalogNamespace, _, _ := unstructured.NestedString(packageManifest.Object, "status", "catalogSourceNamespace")
```
JustinKuli, line 382: I'm surprised the linter doesn't mind this? But I think it's handled correctly, so I'm fine with it.

controllers/operatorpolicy_controller.go
```diff
@@ -332,6 +346,74 @@ func (r *OperatorPolicyReconciler) buildResources(policy *policyv1beta1.Operator
...
+		)
+	}
+
+	if (subSpec.CatalogSource != "" && subSpec.CatalogSource != catalog) ||
+		(subSpec.CatalogSourceNamespace != "" && subSpec.CatalogSourceNamespace != catalogNamespace) {
+		return errors.New(
+			"the subscription defaults could not be determined because the catalog doesn't match those in the " +
+				"PackageManifest",
+		)
+	}
```
JustinKuli, line 396: For a minute, I was thinking that this kind of message could be good to surface regardless of defaulting. But I expect that the subscription would give some kind of error if the catalog doesn't match the package, so the user would still get a message there. Do you think that's right?

controllers/operatorpolicy_controller.go
```diff
@@ -332,6 +346,74 @@ func (r *OperatorPolicyReconciler) buildResources(policy *policyv1beta1.Operator
...
+		return errors.New(
+			"the subscription defaults could not be determined because the PackageManifest didn't specify a catalog",
+		)
+	}
+
+	if (subSpec.CatalogSource != "" && subSpec.CatalogSource != catalog) ||
+		(subSpec.CatalogSourceNamespace != "" && subSpec.CatalogSourceNamespace != catalogNamespace) {
+		return errors.New(
+			"the subscription defaults could not be determined because the catalog doesn't match those in the " +
+				"PackageManifest",
```
JustinKuli, line 394:
```suggestion
			"the subscription defaults could not be determined because the catalog specified in the " +
				"policy does not match what was found in the PackageManifest on the cluster",
```

Especially if the user is not familiar with OLM, I think we should give some extra hints on where these values came from.

test/resources/case38_operator_install/operator-policy-all-defaults.yaml
```diff
@@ -0,0 +1,20 @@
...
+    name: parent-policy
+    uid: 12345678-90ab-cdef-1234-567890abcdef # must be replaced before creation
+spec:
+  remediationAction: enforce
+  severity: medium
+  complianceType: musthave
+  subscription:
+    name: project-quay
+    namespace: operator-policy-testns
+    installPlanApproval: Manual
```
JustinKuli, line 20: This is separate from this PR, but I think that the controller will approve the InstallPlan for this operator... Is that confusing, because the user specified "Manual"? What do you think?

JustinKuli (commented):
Looking clean overall!

controllers/operatorpolicy_controller.go
mprahl: Yes, exactly!

controllers/operatorpolicy_controller.go
mprahl: You're right. I think I'll just return without an error here.

mprahl: Yeah, that sounds right.

test/resources/case38_operator_install/operator-policy-all-defaults.yaml
mprahl: Based on our conversation about this, @JustinKuli will create a separate issue to address this.

controllers/operatorpolicy_controller.go
```diff
@@ -329,7 +352,79 @@ func (r *OperatorPolicyReconciler) buildResources(policy *policyv1beta1.Operator
...
+
+	// PackageManifests come from an API server and not a Kubernetes resource, so the DynamicWatcher can't be used since
+	// it utilizes watches. The namespace doesn't have any meaning but is required.
+	packageManifest, err := r.DynamicClient.Resource(packageManifestGVR).Namespace("default").Get(
+		ctx, subSpec.Package, metav1.GetOptions{},
+	)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return fmt.Errorf(
+				"%wthe subscription defaults could not be determined because the PackageManifest was not found",
```
JustinKuli, line 379: I don't know if I've seen wrapping a specific empty error like this. It's interesting. I'm still trying to find a way to do this kind of thing that I really like, did you find this somewhere else (just for reference)?

controllers/operatorpolicy_controller.go
```diff
@@ -329,7 +352,79 @@ func (r *OperatorPolicyReconciler) buildResources(policy *policyv1beta1.Operator
...
+
+	catalog, _, _ := unstructured.NestedString(packageManifest.Object, "status", "catalogSource")
+	catalogNamespace, _, _ := unstructured.NestedString(packageManifest.Object, "status", "catalogSourceNamespace")
+
+	if catalog == "" || catalogNamespace == "" {
+		return fmt.Errorf(
+			"%wthe subscription defaults could not be determined because the PackageManifest didn't specify a catalog",
+			ErrPackageManifest,
+		)
+	}
```
JustinKuli, line 400: I'm not sure that this one should wrap the error so it's retried - I think usually this would get fixed by the content in the catalog changing? What was your reasoning

controllers/operatorpolicy_controller.go
```diff
@@ -329,7 +352,79 @@ func (r *OperatorPolicyReconciler) buildResources(policy *policyv1beta1.Operator
...
+
+	if subSpec.Channel == "" {
+		defaultChannel, _, _ := unstructured.NestedString(packageManifest.Object, "status", "defaultChannel")
+		if defaultChannel == "" {
+			return fmt.Errorf(
+				"%wthe default channel could not be determined because the PackageManifest didn't specify one",
+				ErrPackageManifest,
+			)
+		}
```
JustinKuli, line 417: Same here, I'm not sure if this should be retried.

controllers/operatorpolicy_controller.go
mprahl: We do this in the propagator with `ErrRetryable` but that comes from me, so I don't know of prior art doing this but I think it does the trick. I think the more "right" thing would be to have a custom error type as opposed to wrapping a custom error, but I didn't think it was worth the extra code.

controllers/operatorpolicy_controller.go
mprahl: I don't know that this would ever occur, but I figured it was likely due to partial results being returned but it seems unlikely. I'll remove the retry.

JustinKuli (approved):
LGTM

PR #230, opened 2024-04-22
Improve OperatorGroup removal logic

JustinKuli:
In some clusters, a default OperatorGroup might be provided in a namespace for "global" operators. This OperatorGroup was not being correctly identified by the policy: in fact any pre-existing group that did not match the group specified by the policy, or the one that would be generated when the policy does not specify a group, was not reported correctly.

Now, pre-existing "special" operator groups should be reported and handled correctly. If they are owned by another resource, they will be considered as "used" for the purposes of the DeleteIfUnused setting.

Refs:
 - https://issues.redhat.com/browse/ACM-11022
 - https://issues.redhat.com/browse/ACM-11077

### DIFF
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -226,7 +226,12 @@ func (r *OperatorPolicyReconciler) handleResources(ctx context.Context, policy *
 		return earlyComplianceEvents, condChanged, err
 	}
 
-	earlyConds, changed, err := r.handleOpGroup(ctx, policy, desiredOG)
+	desiredSubName := ""
+	if desiredSub != nil {
+		desiredSubName = desiredSub.Name
+	}
+
+	earlyConds, changed, err := r.handleOpGroup(ctx, policy, desiredOG, desiredSubName)
 	earlyComplianceEvents = append(earlyComplianceEvents, earlyConds...)
 	condChanged = condChanged || changed
 
@@ -567,7 +572,10 @@ func buildOperatorGroup(
 }
 
 func (r *OperatorPolicyReconciler) handleOpGroup(
-	ctx context.Context, policy *policyv1beta1.OperatorPolicy, desiredOpGroup *operatorv1.OperatorGroup,
+	ctx context.Context,
+	policy *policyv1beta1.OperatorPolicy,
+	desiredOpGroup *operatorv1.OperatorGroup,
+	desiredSubName string,
 ) ([]metav1.Condition, bool, error) {
 	watcher := opPolIdentifier(policy.Namespace, policy.Name)
 
@@ -586,7 +594,7 @@ func (r *OperatorPolicyReconciler) handleOpGroup(
 		return r.musthaveOpGroup(ctx, policy, desiredOpGroup, foundOpGroups)
 	}
 
-	return r.mustnothaveOpGroup(ctx, policy, desiredOpGroup, foundOpGroups)
+	return r.mustnothaveOpGroup(ctx, policy, desiredOpGroup, foundOpGroups, desiredSubName)
 }
 
 func (r *OperatorPolicyReconciler) musthaveOpGroup(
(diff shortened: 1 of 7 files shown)


controllers/operatorpolicy_controller.go
```diff
@@ -725,63 +725,127 @@ func (r *OperatorPolicyReconciler) mustnothaveOpGroup(
...
-			foundOpGroupName = opGroup.GetName()
-			foundOpGroup = opGroup.DeepCopy()
+		foundSubscriptions, err := r.DynamicWatcher.List(
+			watcher, subscriptionGVK, desiredOpGroup.Namespace, labels.Everything())
+		if err != nil {
+			return nil, false, fmt.Errorf("error listing Subscriptions: %w", err)
+		}
 
-			break
+		if len(foundSubscriptions) != 0 {
```
mprahl, line 743: Since the subscription is handled after the operator group, I think this will be true and then the next reconcile this would be false. Perhaps filtering out the subscription if it should be deleted or rearrange the order in `handleResources` if it's must not have?

controllers/operatorpolicy_controller.go
```diff
@@ -790,16 +854,46 @@ func (r *OperatorPolicyReconciler) mustnothaveOpGroup(
...
 	}
 
-	err := r.Delete(ctx, desiredOpGroup)
-	if err != nil {
-		return earlyConds, changed, fmt.Errorf("error deleting the OperatorGroup: %w", err)
+	relatedOpGroups = leftovers // append the deleting and deleted ones to this
+	anyAlreadyDeleting := false
+	deletionErrs := make([]string, 0)
+
+	for _, opGroupToDelete := range toDelete {
```
mprahl, line 861: After a discussion, we agreed that we should only allow the removal behavior of `Keep` and `DeleteIfUnused` for operator group.

JustinKuli: And, if there are multiple operator groups, just report it as noncompliant and don't try to adjust things.

controllers/operatorpolicy_controller.go
JustinKuli: Looking at it more, I don't think this is just an ordering problem: if the policy was only being informed, right now it would report the OperatorGroup as compliant, because it's being used by the (NonCompliant) Subscription. So... something more to think about and fix

JustinKuli:
The force-push diff is probably very helpful for anyone that had already looked at the initial changed: https://github.com/open-cluster-management-io/config-policy-controller/compare/a90a129630b3ae4062e8127237ea38a8c5978354..06ba1e1020039bda973a98dd5a61714c4da52bb1

controllers/operatorpolicy_controller.go
```diff
@@ -716,44 +724,72 @@ func (r *OperatorPolicyReconciler) mustnothaveOpGroup(
...
+	if len(allFoundOpGroups) == 0 {
 		// Missing OperatorGroup: report Compliance
 		changed := updateStatus(policy, missingNotWantedCond("OperatorGroup"), missingNotWantedObj(desiredOpGroup))
 
 		return nil, changed, nil
 	}
 
-	var foundOpGroupName string
-	var foundOpGroup *unstructured.Unstructured
+	if len(allFoundOpGroups) > 1 {
```
mprahl, line 737: Just to confirm, OperatorPolicy would never add an additional operator group if one already existed so there is never a time where this would prevent undoing an OperatorPolicy mistake by the user right? This would only happen if there was already multiple operator groups when the OperatorPolicy was defined or the must have OperatorPolicy was defined and then an additional operator group was added afterwards.

mprahl (approved):
/hold in case you want other reviews

controllers/operatorpolicy_controller.go
JustinKuli: If the controller sees an OperatorGroup already in the namespace, it will not create another one, even if the policy specifies a different OperatorGroup than the one that was found.

But, there is always the possibility of a race condition where something else creates an OperatorGroup between when this controller determines one is not yet present and when it submits its request to create one. But, since this controller does not have concurrent reconciles, it at least can not race with itself.

PR #232, opened 2024-04-23
Update controller-runtime to 0.17.3

JeffeyL:
Changes required to update controller-runtime to 0.17.

Ref: https://issues.redhat.com/browse/ACM-10835


go.mod
```diff
@@ -80,40 +80,41 @@ require (
...
-	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20230501164219-8b0f38b5fd1f // Replaced so that 'go get -u' works. Remove/bump when upgrading.
-	k8s.io/kubectl => k8s.io/kubectl v0.26.10 // Replaced so that 'go get -u' works. Remove/bump when upgrading.
-	sigs.k8s.io/kustomize/api => sigs.k8s.io/kustomize/api v0.12.1 // Replaced so that 'go get -u' works. Remove/bump when upgrading.
-	sigs.k8s.io/kustomize/kyaml => sigs.k8s.io/kustomize/kyaml v0.13.9 // Replaced so that 'go get -u' works. Remove/bump when upgrading.
+	k8s.io/api => k8s.io/api v0.29.2 // Replaced so that 'go get -u' works. Remove/bump when upgrading.
+	k8s.io/client-go => k8s.io/client-go v0.29.2 // Replaced so that 'go get -u' works. Remove/bump when upgrading.
+	k8s.io/kubectl => k8s.io/kubectl v0.28.0 // Replaced so that 'go get -u' works. Remove/bump when upgrading.
+	sigs.k8s.io/kustomize/api => sigs.k8s.io/kustomize/api v0.15.0 // Replaced so that 'go get -u' works. Remove/bump when upgrading.
+	sigs.k8s.io/kustomize/kyaml => sigs.k8s.io/kustomize/kyaml v0.15.0 // Replaced so that 'go get -u' works. Remove/bump when upgrading.
```
JustinKuli, line 119: @dhaiducek if I remember right, you added these? I'd like to remove any of them if possible... is it as simple as removing it here, running a tidy, and then running some test somewhere involving `go test -u`? (that last bit is where I get lost)

go.mod
```diff
@@ -80,40 +80,41 @@ require (
...
 	sigs.k8s.io/kustomize/api v0.15.0 // indirect
 	sigs.k8s.io/kustomize/kyaml v0.15.0 // indirect
-	sigs.k8s.io/structured-merge-diff/v4 v4.3.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
 replace (
 	github.com/imdario/mergo => github.com/imdario/mergo v0.3.16 // Replaced so that 'go get -u' works. Remove/bump when upgrading.
-	golang.org/x/crypto => golang.org/x/crypto v0.14.0 // CVE-2021-43565
+	golang.org/x/crypto => golang.org/x/crypto v0.16.0 // CVE-2021-43565
```
JustinKuli, line 111: This might be a good time to clean up some of these `replace` directives for CVEs. I think that if you remove them here, and check the version it would choose (maybe in the go.sum after running a tidy), we can see if that would still be affected by this CVE. I expect some of them might have been updated in our direct dependencies, and so we don't need the `replace` anymore.

If this ends up being a big task, we can split it out into another issue.

JustinKuli (commented):
Seems good overall! Kudos for finding the right changes for these odd cache settings

main.go
```diff
@@ -216,35 +232,30 @@ func main() {
...
-				&corev1.Namespace{}: nsTransform,
-			},
-		}),
+		Cache: cache.Options{
+			ByObject: cacheByObject,
+		},
 		// Disable the cache for Secrets to avoid a watch getting created when the `policy-encryption-key`
-		// Secret is retrieved. Special cache handling is done by the controller.
-		ClientDisableCacheFor: []client.Object{&corev1.Secret{}},
+		// Secret is retrieved. Special cache handling is done by the controller.\
```
mprahl, line 253:
```suggestion
		// Secret is retrieved. Special cache handling is done by the controller.
```

mprahl (approved):

go.mod
```diff
@@ -80,40 +80,33 @@ require (
...
-	sigs.k8s.io/structured-merge-diff/v4 v4.3.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
 replace (
 	github.com/imdario/mergo => github.com/imdario/mergo v0.3.16 // Replaced so that 'go get -u' works. Remove/bump when upgrading.
-	golang.org/x/crypto => golang.org/x/crypto v0.14.0 // CVE-2021-43565
-	golang.org/x/net => golang.org/x/net v0.17.0 // CVE-2023-39325
-	golang.org/x/text => golang.org/x/text v0.13.0 // CVE-2022-32149
-	gopkg.in/yaml.v2 => gopkg.in/yaml.v2 v2.4.0 // CVE-2022-3064
```
dhaiducek, line 112: I have concerns that removing this block re-introduced affected versions of these packages in `go.sum`. Is updating the `indirects` sufficient to resolve the CVEs?

go.mod
dhaiducek: Yeah, they were set so that Go could programmatically find an unblocked upgrade path with `go get -u`. During an upgrade, they should be removed entirely and replacements added back on an as-needed basis with the highest possible version.

go.mod
JustinKuli: We might be ok - from what I'm reading in https://go.dev/ref/mod#go-sum-files, I think that the `.../go.mod` entries that are now in the go.sum file and refer to older (possibly affected) versions are not actually being used. But I'm going to research this some more because it would be really nice to get rid of some of these directives

dhaiducek: Nice--if that's the case, then we can stop using `replace` for CVEs!

PR #238, opened 2024-05-01
Stop modifying spec.versions directly in musthaveInstallPlan

mprahl:
This creates a separate slice rather than potentially mutating the underlying array of the policy.Spec.Versions slice.


zyjjay:
lgtm

JustinKuli (approved):
I don't think this is necessary, but I'm ok with being a little paranoid about it.

PR #240, opened 2024-05-02
[WIP] Fix consistency of relatedObjects in mustnothave mode

zyjjay:
Mostly for handling of deployments and catalogsources


zyjjay:
@JustinKuli This is pretty much what I had in mind for fixing consistency issues around the relatedObjects for deployments and catalogsources. I realized now that these changes might hurt the readability of the code in exchange for reporting finer details in the relatedObjects field. A simpler alternative is to just create a more "barebones" relatedObject that tells users the controller doesn't handle these objects in mustnothavemode, without caring too much about the granular details (like the name/namespace where the object can be found in). Let me know what you think!

controllers/operatorpolicy_status.go
```diff
@@ -856,19 +857,51 @@ func foundNotWantedObj(obj client.Object) policyv1.RelatedObject {
...
 			UID: string(obj.GetUID()),
 		},
 	}
 }
 
+// missingNotApplicableObj returns a conditionally Compliant RelatedObject with
+// reason = "Resource not found but should exist"/"Resource not found as expected"
+// based on the complianceType specified by the policy
+// Note: These relatedObjects are not applicable *only* in mustnothave mode
+func missingNotApplicableObj(name string, namespace string, complianceType policyv1.ComplianceType, gvk schema.GroupVersionKind) policyv1.RelatedObject {
```
JustinKuli, line 877: Can this be renamed to just `missingObj`? Since it doesn't use the NotApplicable reason, it seems odd.

controllers/operatorpolicy_controller.go
```diff
@@ -1623,9 +1629,26 @@ func (r *OperatorPolicyReconciler) handleCatalogSource(
...
+
+	if isMissing {
+		relObj = missingNotApplicableObj(catalogName, catalogNS, policyv1.MustNotHave, catalogSrcGVK)
+	} else {
+		relObj = foundNotApplicableObj(catalogSrc)
+	}
 
-	changed := updateStatus(policy, catalogSourceFindCond(isUnhealthy, isMissing, catalogName),
-		catalogSourceObj(catalogName, catalogNS, isUnhealthy, isMissing))
+	changed := updateStatus(policy, cond, relObj)
```
JustinKuli, line 1651: Since mustnothave mode does not need to check if the catalogsource is healthy, can it be handled earlier? I have a feeling it would clean up some of this. Right now, in particular, I needed to double-check that the `catalogSrc` here wasn't a nil pointer - but I think it could've just used `foundCatalogSrc`?

JustinKuli (commented):
I think this is on the right track, I think it is a better approach than just always wiping out the related object with a no-name not-applicable one.

zyjjay:
`handleCatalogSource` look a lot cleaner now IMO. Going to modify the tests now to account for the condition/related object changes.

JustinKuli (approved):
LGTM!

PR #242, opened 2024-05-07
Hosted mode Testing for OperatorPolicy

yiraeChristineKim:
The "hosting cluster" should be used for the Policy, OperatorPolicy, and compliance events, while the "target cluster" should be used for all the other resources like Subscription, InstallPlan, etc.
Ref: https://issues.redhat.com/browse/ACM-11255

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -29,6 +29,23 @@
                 "KUBECONFIG": "${workspaceFolder}/kubeconfig_managed_e2e",
             }
         },
+		// Set FDescribe or FIt on the test to debug. Then set the desired breakpoint.
+		{
+			"name": "Launch Hosted Test Function",
+			"type": "go",
+			"request": "launch",
+			"mode": "auto",
+			"program": "${workspaceFolder}/test/e2e/e2e_suite_test.go",
+			"args": [
+				"-ginkgo.debug",
+				"-ginkgo.v",
+				"-is_hosted=true"
+			],
+			"env": {
+				"KUBECONFIG": "${workspaceFolder}/kubeconfig_managed_e2e",
+				"TARGET_KUBECONFIG_PATH": "${workspaceFolder}/kubeconfig_managed2_e2e",
+			}
+		},
         // Set the correct path to the governance-policy-framework repo directory in the env section.
         {
             "name": "Launch Package (Framework E2E) (instructions in launch.json)",
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,7 @@ create-ns:
 # Run against the current locally configured Kubernetes cluster
 .PHONY: run
 run:
-	WATCH_NAMESPACE=$(WATCH_NAMESPACE) go run ./main.go controller --leader-elect=false
+	WATCH_NAMESPACE=$(WATCH_NAMESPACE) go run ./main.go controller --leader-elect=false --enable-operator-policy=true
 
 ############################################################
 # clean section
@@ -193,7 +193,7 @@ kind-create-cluster:
 .PHONY: kind-additional-cluster
 kind-additional-cluster: MANAGED_CLUSTER_SUFFIX = 2
 kind-additional-cluster: CLUSTER_NAME = $(MANAGED_CLUSTER_NAME)
-kind-additional-cluster: kind-create-cluster kind-controller-kubeconfig
+kind-additional-cluster: kind-create-cluster kind-controller-kubeconfig install-crds-hosted
 
 .PHONY: kind-delete-cluster
 kind-delete-cluster:
@@ -204,11 +204,10 @@ kind-delete-cluster:
 kind-tests: kind-delete-cluster kind-bootstrap-cluster-dev kind-deploy-controller-dev e2e-test
 
 OLM_VERSION := v0.26.0
+OLM_INSTALLER = $(LOCAL_BIN)/install.sh
 
 .PHONY: install-crds
-install-crds:
-	@echo installing olm
-	curl -L https://github.com/operator-framework/operator-lifecycle-manager/releases/download/$(OLM_VERSION)/install.sh -o $(LOCAL_BIN)/install.sh
+install-crds:  $(OLM_INSTALLER)
 	chmod +x $(LOCAL_BIN)/install.sh
 	$(LOCAL_BIN)/install.sh $(OLM_VERSION)
 	@echo installing crds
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -230,17 +241,22 @@ install-resources:
 	kubectl apply -k deploy/rbac
 	kubectl apply -f deploy/manager/service-account.yaml -n $(KIND_NAMESPACE)
 
+IS_HOSTED ?= false
+E2E_PROCS=20
+
 .PHONY: e2e-test
 e2e-test: e2e-dependencies
-	$(GINKGO) -v --procs=20 $(E2E_TEST_ARGS) test/e2e
+	$(GINKGO) -v --procs=$(E2E_PROCS) $(E2E_TEST_ARGS) test/e2e -- -is_hosted=$(IS_HOSTED)
```
yiraeChristineKim, line 249: Hosted mode test only 2. Therefore, setting up 20 procs takes more time unnecessary. Un

.vscode/launch.json
```diff
@@ -1,48 +1,89 @@
...
+        "--v=5",
+        "--enable-operator-policy=true"
+      ],
+      "env": {
+        "WATCH_NAMESPACE": "managed",
+        "KUBECONFIG": "${workspaceFolder}/kubeconfig_managed"
+      }
+    },
+    {
+        "name": "Launch hosted Package",
```
yiraeChristineKim, line 24: I added

.vscode/launch.json
```diff
@@ -1,48 +1,89 @@
...
+        ],
+        "env": {
+          "WATCH_NAMESPACE": "managed",
+          "KUBECONFIG": "${workspaceFolder}/kubeconfig_managed"
         }
-    ]
+      },
+    // Set FDescribe or FIt on the test to debug. Then set the desired breakpoint.
+    {
+      "name": "Launch Test Function (instructions in launch.json)",
```
yiraeChristineKim, line 44: I added

.vscode/launch.json
JustinKuli, line 1: It's clear you've been a JS dev, it looks like your editor prefers 2-space indenting, I think Matt's used 4-space :joy: 

The GH diff here is big if you don't turn off "show whitespace"... I think I prefer your indenting honestly, but we should probably just keep it consistent with the other `launch.json` files in the rest of the project.

Makefile
```diff
@@ -106,7 +106,7 @@ create-ns:
 # Run against the current locally configured Kubernetes cluster
 .PHONY: run
 run:
-	WATCH_NAMESPACE=$(WATCH_NAMESPACE) go run ./main.go controller --leader-elect=false
+	WATCH_NAMESPACE=$(WATCH_NAMESPACE) go run ./main.go controller --leader-elect=false --enable-operator-policy=true #--target-kubeconfig-path=./kubeconfig_managed2
```
JustinKuli, line 109: nit: remove commented part.

Adding `--enable-operator-policy=true` is a good idea though.

Makefile
```diff
@@ -221,6 +220,18 @@ install-crds:
 	kubectl apply -f deploy/crds/policy.open-cluster-management.io_configurationpolicies.yaml
 	kubectl apply -f deploy/crds/policy.open-cluster-management.io_operatorpolicies.yaml
 
+
+
+$(OLM_INSTALLER):
+	@echo installing olm
```
JustinKuli, line 226: nit:
```suggestion
	@echo getting OLM installer
```

Makefile
```diff
@@ -230,17 +241,22 @@ install-resources:
...
-	$(GINKGO) -v --procs=20 $(E2E_TEST_ARGS) test/e2e
+	$(GINKGO) -v --procs=$(E2E_PROCS) $(E2E_TEST_ARGS) test/e2e -- -is_hosted=$(IS_HOSTED)
 
 .PHONY: e2e-test-coverage
 e2e-test-coverage: E2E_TEST_ARGS = --json-report=report_e2e.json --label-filter='!hosted-mode && !running-in-cluster' --output-dir=.
 e2e-test-coverage: e2e-run-instrumented e2e-test e2e-stop-instrumented
 
 .PHONY: e2e-test-hosted-mode-coverage
-e2e-test-hosted-mode-coverage: E2E_TEST_ARGS = --json-report=report_e2e_hosted_mode.json --label-filter="hosted-mode && !running-in-cluster" --output-dir=.
+e2e-test-hosted-mode-coverage: E2E_TEST_ARGS = --json-report=report_e2e_hosted_mode.json --label-filter="hosted-mode || hosted-as-well && !running-in-cluster" --output-dir=.
```
JustinKuli, line 256: Why do we need a second label for hosted tests?

test/e2e/case38_install_operator_test.go
```diff
@@ -170,13 +170,32 @@ var _ = Describe("Testing OperatorPolicy", Ordered, func() {
...
+		if IsHosted {
+			KubectlTarget("create", "ns", opPolTestNS)
+			DeferCleanup(func() {
+				KubectlTarget("delete", "ns", opPolTestNS)
+			})
+		}
+		utils.Kubectl("create", "ns", opPolTestNS)
+		DeferCleanup(func() {
+			utils.Kubectl("delete", "ns", opPolTestNS)
+		})
```
JustinKuli, line 183: Should this be in an `else`?

test/e2e/case38_install_operator_test.go
```diff
@@ -170,13 +170,32 @@ var _ = Describe("Testing OperatorPolicy", Ordered, func() {
...
 			opPolName = "oppol-all-defaults"
 			subName   = "project-quay"
 		)
 		BeforeAll(func() {
+			if IsHosted {
+				KubectlTarget("create", "ns", opPolTestNS)
+				DeferCleanup(func() {
+					KubectlTarget("delete", "ns", opPolTestNS)
+				})
+			}
```
JustinKuli, line 198: Can it use `preFunc` here?

JustinKuli (commented):
I think it looks good overall! Just some small things I noticed

Makefile
dhaiducek: > ...Spec selection can be handled in `e2e_suite_test.go`.

Actually, it seems I might be wrong on that. But it seems the goal is to get `case38` to run? So maybe instead of the flag we might be able to just parse the labels in `case38`. Untested, but something like:
```
if Label("hosted-mode").MatchesLabelFilter(GinkgoLabelFilter()) {
```
ref: https://onsi.github.io/ginkgo/#spec-labels

It'd depend on how exactly Ginkgo handles setting up the nodes, though.

yiraeChristineKim: Operator Policy tests are tested in normal mode as well

test/e2e/case38_install_operator_test.go
yiraeChristineKim: I use this clean policy in manage cluster(not hosted cluater)

test/e2e/case38_install_operator_test.go
yiraeChristineKim: Thanks!!

test/e2e/case38_install_operator_test.go
```diff
@@ -23,7 +23,7 @@ import (
 	"open-cluster-management.io/config-policy-controller/test/utils"
 )
 
-var _ = Describe("Testing OperatorPolicy", Ordered, func() {
+var _ = Describe("Testing OperatorPolicy", Ordered, Label("hosted-as-well"), func() {
```
JustinKuli, line 26: I wonder if you can define the whole thing twice, once with the label, and once without... Or maybe Dale had a suggestion for something similar?

But we can probably just merge this as is, the additional label isn't too bad. Let me know if you're tired of trying random things from our comments :laughing: 

```suggestion
var _ = Describe("Testing OperatorPolicy", Ordered, Label("hosted-mode"), installOperatorTests)
var _ = Describe("Testing OperatorPolicy", Ordered, installOperatorTests)

func installOperatorTests() {
```

yiraeChristineKim: This is good but why are we afraid of adding a new label?

JustinKuli: Mainly I just wanted the test filters to be as simple as possible, it makes it more obvious which tests they run. But the new label is probably fine

mprahl (approved):

PR #246, opened 2024-05-17
WIP: Add support for recording the diff in the ConfigurationPolicy status

mprahl:
A new recordDiff option of "InStatus" allows the diff to be stored in
the object properties in the ConfigurationPolicy status.

The new default recordDiff value is "InStatus" unless sensitive data may
be in the diff. Then the user must explicitly set recordDiff.

Relates:
https://issues.redhat.com/browse/ACM-11421


api/v1/configurationpolicy_types_test.go
```diff
@@ -0,0 +1,95 @@
...
+			apiVersion: "oauth.openshift.io/v1",
+			kind:       "OAuthAuthorizeTokens",
+			expected:   RecordDiffSensored,
+		},
+		{
+			testName:   "Route-recordDiff-unset",
+			apiVersion: "route.openshift.io/v1",
+			kind:       "Route",
+			expected:   RecordDiffSensored,
+		},
```
yiraeChristineKim, line 71:
```suggestion
		},
		{
			testName:   "Route-recordDiff-unset",
			apiVersion: "route.openshift.io/v1",
			kind:       "Route",
			recordDiff: RecordDiffLog,
			expected:   RecordDiffSensored,
		},
```

Add this test too?

controllers/configurationpolicy_controller.go
```diff
@@ -1030,11 +1031,34 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
+							}
+						}
+					}
+
 					plc.Spec.ObjectTemplates = objTemps
 
 					break
 				}
 
+				if plc.Spec.ObjectTemplates[i].RecordDiff == "" && resolvedTemplate.HasSensitiveData {
```
yiraeChristineKim, line 1053: When a case of resolvedTemplate.HasSensitiveData falls here...  I think all HasSensitiveData cases falls to line 1034.

controllers/configurationpolicy_controller.go
```diff
@@ -2711,10 +2741,58 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
...
+
 		r.setEvaluatedObject(obj.policy, obj.existingObj, !throwSpecViolation)
 	}
 
-	return throwSpecViolation, "", updateNeeded, updateSucceeded
+	return throwSpecViolation, "", diff, updateNeeded, updateSucceeded
+}
+
+// handleDiff will generate the diff and then log it or return it based on the input recordDiff value. If recordDiff
+// is set to None or is set to InStatus with enforce, no diff is generated. When recordDiff is set to Sensored, a
```
yiraeChristineKim, line 2761: inform + inStatus creates diff?  why?

mprahl: If the policy is enforced, then the diff is no longer relevant after the update so it shouldn't get stored in the status.

api/v1/configurationpolicy_types_test.go
mprahl: If the user explicitly sets `recordDiff` we want to honor that. `RecordDiffSensored` is just to detect when sensitive data is present and we don't want to store the diff in the status by default.

controllers/configurationpolicy_controller.go
```diff
@@ -2711,10 +2741,58 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
...
+		return ""
+	}
+
+	if !isInform && recordDiff == policyv1.RecordDiffInStatus {
+		return ""
+	}
+
+	computedDiff, err := generateDiff(existingObject, mergedObject)
+	if err != nil {
+		log.Info("Failed to generate the diff: " + err.Error())
```
JustinKuli, line 2780: Is this log.Info because it's a user-error? It's not clear to me what could cause an error here

controllers/configurationpolicy_controller.go
```diff
@@ -2711,10 +2741,58 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
...
+		return ""
+	}
+
+	switch recordDiff { //nolint:exhaustive
+	case policyv1.RecordDiffLog:
+		log.Info("Logging the diff:\n" + computedDiff)
+	case policyv1.RecordDiffInStatus:
+		return computedDiff
+	case policyv1.RecordDiffSensored:
+		return `# This diff may contain sensitive data. The "recordDiff" field must be set to "InStatus" ` +
```
JustinKuli, line 2791: Why have `#` at the start?

controllers/configurationpolicy_controller.go
```diff
@@ -2711,10 +2741,58 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
...
+	}
+
+	computedDiff, err := generateDiff(existingObject, mergedObject)
+	if err != nil {
+		log.Info("Failed to generate the diff: " + err.Error())
+
+		return ""
+	}
+
+	switch recordDiff { //nolint:exhaustive
```
JustinKuli, line 2785: I would prefer to make the switch cases exhaustive, I think it makes things a tiny bit more clear.

controllers/configurationpolicy_controller.go
JustinKuli: I was confused about this as well, can the comment explain that detail?

deploy/crds/kustomize_operatorpolicy/policy.open-cluster-management.io_operatorpolicies.yaml
```diff
@@ -261,6 +261,8 @@ spec:
                           description: Whether the object was created by the parent
                             policy
                           type: boolean
+                        diff:
+                          type: string
```
JustinKuli, line 265: Can we remove this from OperatorPolicy? I think it should be easy with another patch in its kustomization.

JustinKuli (changes requested):
Looks good overall.

controllers/configurationpolicy_controller.go
JustinKuli: I was confused about this as well, can the comment explain that detail?

JustinKuli:
The Jira mentions "the new options of `InStatus` and `Automatic`" but I think this implementation doesn't have `Automatic` - it just has some behavior when the field isn't set. That's fine with me, just wanted to make sure that detail doesn't get lost.

api/v1/configurationpolicy_types.go
```diff
@@ -199,16 +200,47 @@ type ObjectTemplate struct {
...
+// +kubebuilder:validation:Enum=Log;InStatus;None
 type RecordDiff string
 
 const (
-	RecordDiffLog  RecordDiff = "Log"
-	RecordDiffNone RecordDiff = "None"
+	RecordDiffLog      RecordDiff = "Log"
+	RecordDiffInStatus RecordDiff = "InStatus"
+	RecordDiffNone     RecordDiff = "None"
+	RecordDiffSensored RecordDiff = "Sensored"
```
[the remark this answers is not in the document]
mprahl: I meant "Censored" but misspelled it lol. Thanks for catching that.

controllers/configurationpolicy_controller.go
mprahl: That was copied from the previous code but it seems to make sense to make it an error log.

controllers/configurationpolicy_controller.go
mprahl: I was on the fence about this but the thought was that `# ` would make it a valid comment when syntax highlighting the output.

controllers/configurationpolicy_controller.go
mprahl: I added a comment that should help clarify this.

controllers/configurationpolicy_controller.go
JustinKuli: I don't know if it will... but regardless I don't think it's a problem, and it could help the user identify these situations as 'comments' and not real diffs, so I think that's good.

JustinKuli (approved):
Looks good! Thanks for those updates.

PR #247, opened 2024-05-20
Check and report on overlapping subs

JustinKuli:
This adds two new status fields to OperatorPolicy, which are used to detect when two OperatorPolicies manage the same subscription. Overlapping OperatorPolicies will go into a non-compliant state, report the overlap, and will not compete with each other.

Refs:
 - https://issues.redhat.com/browse/ACM-11616


api/v1beta1/operatorpolicy_types.go
```diff
@@ -148,6 +148,13 @@ type OperatorPolicyStatus struct {
 	// List of resources processed by the policy
 	// +optional
 	RelatedObjects []policyv1.RelatedObject `json:"relatedObjects"`
+
+	// The resolved name.namespace of the subscription
+	ResolvedSubscriptionLabel string `json:"resolvedSubscriptionLabel,omitempty"`
+
+	// The list of overlapping OperatorPolicies (as name.namespace) which all watch the same
```
mprahl, line 155:
```suggestion
	// The list of overlapping OperatorPolicies (as name.namespace) which all manage the same
```

controllers/operatorpolicy_controller.go
```diff
@@ -114,12 +115,42 @@ func (r *OperatorPolicyReconciler) SetupWithManager(mgr ctrl.Manager, depEvents
...
 		Complete(r)
 }
 
+func overlapMapper(_ context.Context, obj client.Object) []reconcile.Request {
+	//nolint:forcetypeassert
+	pol := obj.(*policyv1beta1.OperatorPolicy)
+
+	var result []reconcile.Request
+
+	for _, overlap := range pol.Status.OverlappingPolicies {
```
mprahl, line 133: Would it be cleaner to have `OverlappingPolicies` be a struct with the name and namespace field separated?

mprahl (commented):
A few minor comments but this looks great. Nice solution!

controllers/operatorpolicy_controller.go
JustinKuli: Being able to use `slices.Sort`, `slices.Equal` and `strings.Join` on them as they are right now is kind of nice. It wouldn't be too difficult to get those same kinds of things on a new struct, but I don't think we would gain much overall.

mprahl (approved):

PR #249, opened 2024-05-21
Add upgradeApproval field to OperatorPolicy

JustinKuli:
Breaking change: setting `InstallPlanApproval` on the subscription is invalid, and will result in a noncompliant policy that does not perform any other actions. Previously, *not* setting that field was invalid.

Being able to set `InstallPlanApproval` in the policy was somewhat confusing, because the controller would override the supplied value in many cases, in order to have control over which upgrades would be approved. Additionally, there was a desire to separate the approval for the initial installation from the approval for later upgrades.

The new field addresses these concerns. Initial installs will be approved whenever the policy is enforced (as long as it matches the policy's specified allowed versions), and upgrades will only be approved when `upgradeApproval` is set to Automatic.

Refs:
 - https://issues.redhat.com/browse/ACM-11268

### DIFF
--- a/api/v1beta1/operatorpolicy_types.go
+++ b/api/v1beta1/operatorpolicy_types.go
@@ -135,6 +135,14 @@ type OperatorPolicySpec struct {
 	// When in inform mode, any resources that would be deleted if the policy was enforced will
 	// be causes for NonCompliance, but resources that would be kept will be considered Compliant.
 	RemovalBehavior RemovalBehavior `json:"removalBehavior,omitempty"`
+
+	//+kubebuilder:validation:Required
+	//+kubebuilder:validation:Enum=None;Automatic
+	// UpgradeApproval determines whether 'upgrade' InstallPlans for the operator will be approved
+	// by the controller when the policy is enforced and in 'musthave' mode. The initial InstallPlan
+	// approval is not affected by this setting. This setting has no effect when the policy is in
+	// 'mustnothave' mode. Allowed values are "None" or "Automatic".
+	UpgradeApproval string `json:"upgradeApproval"`
 }
 
 // OperatorPolicyStatus defines the observed state of OperatorPolicy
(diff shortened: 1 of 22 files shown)


api/v1beta1/operatorpolicy_types.go
```diff
@@ -135,6 +135,14 @@ type OperatorPolicySpec struct {
 	// When in inform mode, any resources that would be deleted if the policy was enforced will
 	// be causes for NonCompliance, but resources that would be kept will be considered Compliant.
 	RemovalBehavior RemovalBehavior `json:"removalBehavior,omitempty"`
+
+	//+kubebuilder:validation:Required
+	//+kubebuilder:validation:Enum=None;Automatic
+	// UpgradeApproval determines whether 'upgrade' InstallPlans for the operator will be approved
+	// by the controller when the policy is enforced and in 'musthave' mode. The initial InstallPlan
+	// approval is not affected by this setting. This setting has no effect in when the policy is in
```
gparvin, line 143:
```suggestion
	// approval is not affected by this setting. This setting has no effect when the policy is in
```

deploy/crds/policy.open-cluster-management.io_operatorpolicies.yaml
```diff
@@ -124,6 +124,16 @@ spec:
                   https://olm.operatorframework.io/docs/concepts/crds/subscription/
                 type: object
                 x-kubernetes-preserve-unknown-fields: true
+              upgradeApproval:
+                description: |-
+                  UpgradeApproval determines whether 'upgrade' InstallPlans for the operator will be approved
+                  by the controller when the policy is enforced and in 'musthave' mode. The initial InstallPlan
+                  approval is not affected by this setting. This setting has no effect in when the policy is in
```
gparvin, line 131:
```suggestion
                  approval is not affected by this setting. This setting has no effect when the policy is in
```

yiraeChristineKim (dismissed):
good to me

gparvin (approved):

PR #251, opened 2024-05-22
Delete the hosting test namespace before the hosted test namespace

mprahl:
This ensures the OperatorPolicy is deleted before the operator is deleted. Hopefully this will address some of the test flakes.

The first commit includes a launch.json configuration for hosted mode.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -14,6 +14,26 @@
                 "KUBECONFIG": "${workspaceFolder}/kubeconfig_managed",
             }
         },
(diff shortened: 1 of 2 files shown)


JustinKuli (approved):
Worth a try at least :laughing:

dhaiducek:
Things seem to be slower on older Kubernetes--the E2E tests seem to fail more frequently on the `minimum` version. I'm wondering: Should we extend the default timeout just for the older version? Or bump the `minimum` version?

PR #252, opened 2024-05-23
Implement new ComplianceConfig field

zyjjay:
Users can define how specific resource statuses affect the overall OperatorPolicy compliance events.

ref: https://issues.redhat.com/browse/ACM-11023


JustinKuli (commented):
Looks good overall to me! I think the one e2e test, combined with the unit tests is sufficient for verifying the behavior.

I think we're considering changing the default behaviors, which look easy to adjust here.

controllers/operatorpolicy_status.go
```diff
@@ -669,12 +675,16 @@ var installPlansNoApprovals = metav1.Condition{
...
 // and a message detailing which possible updates are available
-func installPlanUpgradeCond(versions []string, approvableIPs []unstructured.Unstructured) metav1.Condition {
+func installPlanUpgradeCond(
+	complianceConfig policyv1beta1.ComplianceConfigAction,
+	versions []string,
+	approvableIPs []unstructured.Unstructured,
+) metav1.Condition {
 	// FUTURE: check policy.spec.statusConfig.upgradesAvailable to determine `compliant`.
 	// For now this condition assumes it is set to 'NonCompliant'
+	canModifyComp := complianceConfig == "NonCompliant"
```
JustinKuli, line 685: I think I'd prefer using `complianceConfig == "NonCompliant"` directly in the condition below, it might make it easier in the future if we add other options. Also just a reminder to remove the `// FUTURE` comment.

controllers/operatorpolicy_status.go
```diff
@@ -775,6 +794,11 @@ func buildDeploymentCond(
 		message = fmt.Sprintf("the deployments %s do not have their minimum availability", names)
 	}
 
+	if !canModifyComp {
+		status = metav1.ConditionTrue
+		message += " - ignore, compliance overridden by complianceConfig"
```
JustinKuli, line 799: Can we add a conditional to only add this message if it is overriding the status? It looks to me like it would be adding this even if all of the deployments are available.

controllers/operatorpolicy_status.go
```diff
@@ -1044,7 +1079,7 @@ func noInstallPlansObj(namespace string) policyv1.RelatedObject {
 // existingInstallPlanObj returns a RelatedObject for the InstallPlan, with a reason
 // like 'The InstallPlan is ____' based on the phase. When the InstallPlan is in phase
 // 'Complete', the object will be Compliant, otherwise it will be NonCompliant.
-func existingInstallPlanObj(ip client.Object, phase string) policyv1.RelatedObject {
+func existingInstallPlanObj(ip client.Object, phase string, complianceConfig string) policyv1.RelatedObject {
```
JustinKuli, line 1082: I would prefer if this took type `policyv1beta1.ComplianceConfigAction`, there are two conversions to string that would be unnecessary.

JustinKuli (commented):
Just some early comments - they might be things you're already looking at changing.

mprahl:
Could you please squash the commits and update the PR title and commit title to mention `complianceConfig` instead of `statusConfig`?

api/v1beta1/operatorpolicy_types.go
```diff
@@ -12,19 +12,20 @@ import (
...
 
 // RemovalAction : Keep, Delete, or DeleteIfUnused
 type RemovalAction string
 
 const (
-	// StatusMessageOnly is a StatusConfigAction that only shows the status message
-	StatusMessageOnly StatusConfigAction = "StatusMessageOnly"
-	// NonCompliant is a StatusConfigAction that shows the status message and sets
+	// Compliant is a ComplianceConfigAction that only shows the status message and
+	// sets thje compliance to Compliant
```
mprahl, line 24:
```suggestion
	// Compliant is a ComplianceConfigAction that only shows the status message and
	// does not affect the overall compliance.
```

api/v1beta1/operatorpolicy_types.go
```diff
@@ -12,19 +12,20 @@ import (
...
+	Compliant ComplianceConfigAction = "Compliant"
+	// NonCompliant is a ComplianceConfigAction that shows the status message and sets
 	// the compliance to NonCompliant
```
mprahl, line 27:
```suggestion
	// NonCompliant is a ComplianceConfigAction that shows the status message and sets
	// the overall compliance to NonCompliant when the condition is met
```

api/v1beta1/operatorpolicy_types.go
```diff
@@ -98,12 +99,14 @@ func (rb RemovalBehavior) ApplyDefaults() RemovalBehavior {
 	return withDefaults
 }
 
-// StatusConfig defines how resource statuses affect the OperatorPolicy status and compliance
-type StatusConfig struct {
-	CatalogSourceUnhealthy StatusConfigAction `json:"catalogSourceUnhealthy,omitempty"`
-	DeploymentsUnavailable StatusConfigAction `json:"deploymentsUnavailable,omitempty"`
-	UpgradesAvailable      StatusConfigAction `json:"upgradesAvailable,omitempty"`
-	UpgradesProgressing    StatusConfigAction `json:"upgradesProgressing,omitempty"`
+// ComplianceConfig defines how resource statuses affect the OperatorPolicy status and compliance
```
mprahl, line 102:
```suggestion
// ComplianceConfig defines how resource statuses affect the OperatorPolicy compliance
```

api/v1beta1/operatorpolicy_types.go
```diff
@@ -98,12 +99,14 @@ func (rb RemovalBehavior) ApplyDefaults() RemovalBehavior {
...
+type ComplianceConfig struct {
+	// +kubebuilder:default=Compliant
```
mprahl, line 104: I like the idea of having a default, but this does mean the template sync `equivalentPolicies` will have to be modified to not think the OperatorPolicy on the cluster is different than the one specified in the `Policy` object if the defaults aren't specified in the `Policy` object. See how this is handled for `ConfigurationPolicy`:
https://github.com/open-cluster-management-io/governance-policy-framework-addon/blob/c34eec6923b9f4079977c116ecd513f1eb9f22b7/controllers/templatesync/template_sync.go#L958

api/v1beta1/operatorpolicy_types.go
```diff
@@ -130,7 +133,7 @@ type OperatorPolicySpec struct {
 	// in 'inform' mode, and which installPlans are approved when in 'enforce' mode
 	Versions []policyv1.NonEmptyString `json:"versions,omitempty"`
 
-	//+kubebuilder:default={}
+	// +kubebuilder:default={}
```
mprahl, line 136: Generally, comments related to autogenerating code don't have a space after `//` but I'm not strict about this. Some others on the team may be :wink:

api/v1beta1/operatorpolicy_types.go
```diff
@@ -143,6 +146,12 @@ type OperatorPolicySpec struct {
 	// approval is not affected by this setting. This setting has no effect when the policy is in
 	// 'mustnothave' mode. Allowed values are "None" or "Automatic".
 	UpgradeApproval string `json:"upgradeApproval"`
+
+	// +kubebuilder:default={}
+	// ComplianceConfig defines how resource statuses affect the OperatorPolicy status and compliance.
+	// Options include Compliant, which will essentially ignore NonCompliance for certain cases,
+	// and NonCompliant, which will report NonCompliance for those cases.
```
mprahl, line 153:
```suggestion
	// ComplianceConfig defines how resource statuses affect the OperatorPolicy compliance.
	// When set to Compliant, the condition does not impact the OperatorPolicy compliance. When
	// set to NonCompliant, the condition causes the OperatorPolicy to become NonComplaint.
```

controllers/operatorpolicy_controller.go
```diff
@@ -1356,7 +1356,9 @@ func (r *OperatorPolicyReconciler) musthaveInstallPlan(
 			}
 		}
 
-		relatedInstallPlans = append(relatedInstallPlans, existingInstallPlanObj(&ownedInstallPlans[i], phase))
+		complianceConfig := policy.Spec.ComplianceConfig.UpgradesAvailable
```
mprahl, line 1359: Could you set this once outside of the loop so it's not redefined every loop iteration and the code below can reference it as well?

controllers/operatorpolicy_status.go
```diff
@@ -1044,7 +1077,11 @@ func noInstallPlansObj(namespace string) policyv1.RelatedObject {
 // existingInstallPlanObj returns a RelatedObject for the InstallPlan, with a reason
 // like 'The InstallPlan is ____' based on the phase. When the InstallPlan is in phase
 // 'Complete', the object will be Compliant, otherwise it will be NonCompliant.
```
mprahl, line 1079: This docstring needs to be updated to mention the new argument.

controllers/operatorpolicy_status.go
```diff
@@ -1060,9 +1097,12 @@ func existingInstallPlanObj(ip client.Object, phase string) policyv1.RelatedObje
...
-		relObj.Compliant = string(policyv1.NonCompliant)
+		// Check policy.spec.statusConfig.upgradesAvailable to determine `compliant`.
+		if complianceConfig != "Compliant" {
+			relObj.Compliant = string(policyv1.NonCompliant)
+		} else {
+			relObj.Compliant = string(policyv1.Compliant)
+		}
 	case string(operatorv1alpha1.InstallPlanPhaseComplete):
 		relObj.Compliant = string(policyv1.Compliant)
 	default:
```
mprahl, line 1108: Does this default mean that an InstallPlan that is in the installing state is NonCompliant? I would not have expected this if `complianceConfig` is `Compliant`.

controllers/operatorpolicy_controller.go
```diff
@@ -1446,8 +1448,11 @@ func (r *OperatorPolicyReconciler) musthaveInstallPlan(
 	}
 
 	if len(approvableInstallPlans) != 1 {
-		changed := updateStatus(policy,
-			installPlanUpgradeCond(allUpgradeVersions, approvableInstallPlans), relatedInstallPlans...)
+		changed := updateStatus(
+			policy,
+			installPlanUpgradeCond(complianceConfig, allUpgradeVersions, approvableInstallPlans),
```
mprahl, line 1453: @JustinKuli this seems like an odd case where OperatorPolicy can't perform an upgrade but the user wouldn't be notified based on `complianceConfig`. I think this particular case should always be NonCompliant. What do you think?

controllers/operatorpolicy_controller.go
```diff
@@ -1680,15 +1685,19 @@ func (r *OperatorPolicyReconciler) handleDeployment(
 		if policy.Spec.ComplianceType.IsMustNotHave() {
 			relatedObjects = append(relatedObjects, foundNotApplicableObj(&dep))
 		} else {
-			relatedObjects = append(relatedObjects, existingDeploymentObj(&dep))
+			complianceConfig := policy.Spec.ComplianceConfig.DeploymentsUnavailable
```
mprahl, line 1688: Consider defining this once outside of the loop.

controllers/operatorpolicy_status.go
```diff
@@ -693,6 +700,13 @@ func installPlanUpgradeCond(versions []string, approvableIPs []unstructured.Unst
 		cond.Message += " but multiple of those match the versions specified in the policy"
 	}
 
+	if complianceConfig == "NonCompliant" {
```
mprahl, line 703: Optional: It doesn't really matter in this case because there are only two values, but I prefer checking specifically for "Compliant" and if it's not, assume it's "NonCompliant". If we add additional options in the future, this is clearer.

controllers/operatorpolicy_status.go
```diff
@@ -693,6 +700,13 @@ func installPlanUpgradeCond(versions []string, approvableIPs []unstructured.Unst
...
+		cond.Status = metav1.ConditionFalse
+	} else {
+		cond.Status = metav1.ConditionTrue
+		cond.Message += " - ignore, compliance overridden by complianceConfig"
```
mprahl, line 707: I don't think this additional message is required. Now that the default is "Compliant", I don't think overridden is the correct term.

controllers/operatorpolicy_status.go
```diff
@@ -750,6 +764,7 @@ var crdFoundCond = metav1.Condition{
 // minimum availability, the condition will be NonCompliant, and the message will
 // list the unavailable deployments.
 func buildDeploymentCond(
```
mprahl, line 766: The docstring needs to be updated with the new argument.

controllers/operatorpolicy_status.go
```diff
@@ -775,6 +790,13 @@ func buildDeploymentCond(
 		message = fmt.Sprintf("the deployments %s do not have their minimum availability", names)
 	}
 
+	// If the status changed, that means the condition should be NonCompliant
+	// Apply override if complianceConfig is set to Compliant
+	if status == metav1.ConditionFalse && complianceConfig == "Compliant" {
+		status = metav1.ConditionTrue
+		message += " - ignore, compliance overridden by complianceConfig"
```
mprahl, line 797:
```suggestion
		message += " (policy compliance is not impacted due to complianceConfig)"
```

controllers/operatorpolicy_status.go
```diff
@@ -794,7 +816,12 @@ var noDeploymentsCond = metav1.Condition{
 
 // catalogSourceFindCond is a conditionally compliant condition with reason
 // based on the `isUnhealthy` and `isMissing` parameters
```
mprahl, line 818: The docstring needs to be updated with the new argument.

controllers/operatorpolicy_status.go
```diff
@@ -811,6 +838,12 @@ func catalogSourceFindCond(isUnhealthy bool, isMissing bool, name string) metav1
 		message = "CatalogSource '" + name + "' was not found"
 	}
 
+	// Only override if condition evaluated to NonCompliant
+	if status == metav1.ConditionTrue && complianceConfig == "Compliant" {
+		status = metav1.ConditionFalse
+		message += " - ignore, compliance overridden by complianceConfig"
```
mprahl, line 844: I don't think this message is required since the default is "Compliant" now.

controllers/operatorpolicy_status.go
```diff
@@ -1155,13 +1195,18 @@ var noExistingCRDObj = policyv1.RelatedObject{
 
 // existingDeploymentObj returns a RelatedObject for a Deployment, which will
 // be Compliant if there are no unavailable replicas on the deployment.
```
mprahl, line 1197: The docstrin

controllers/operatorpolicy_status.go
```diff
@@ -1189,17 +1234,29 @@ var noExistingDeploymentObj = policyv1.RelatedObject{
 
 // catalogSourceObj returns a conditionally compliant RelatedObject with reason based on the
 // `isUnhealthy` and `isMissing` parameters
```
mprahl, line 1236: The docstring needs to be updated for the new argument.

api/v1beta1/operatorpolicy_types.go
```diff
@@ -98,12 +99,14 @@ func (rb RemovalBehavior) ApplyDefaults() RemovalBehavior {
...
-// StatusConfig defines how resource statuses affect the OperatorPolicy status and compliance
-type StatusConfig struct {
-	CatalogSourceUnhealthy StatusConfigAction `json:"catalogSourceUnhealthy,omitempty"`
-	DeploymentsUnavailable StatusConfigAction `json:"deploymentsUnavailable,omitempty"`
-	UpgradesAvailable      StatusConfigAction `json:"upgradesAvailable,omitempty"`
-	UpgradesProgressing    StatusConfigAction `json:"upgradesProgressing,omitempty"`
+// ComplianceConfig defines how resource statuses affect the OperatorPolicy status and compliance
+type ComplianceConfig struct {
+	// +kubebuilder:default=Compliant
+	CatalogSourceUnhealthy ComplianceConfigAction `json:"catalogSourceUnhealthy,omitempty"`
```
mprahl, line 105: Could you document these fields so they end up in the CRD?

controllers/operatorpolicy_status.go
zyjjay: I think because this condition is a NonCompliant condition, it would make sense to to say it was overridden. Especially since the condition message makes it to the final policy compliance event, it might be useful to see this detail.

JustinKuli: I think I somehow agree with both of you :laughing: - saying its overriden doesn't feel right, but I don't want to *not* mention that the possible upgrade is being ignored.

Maybe a rephrasing would help, something like " - not triggering noncompliance because of complianceConfig".

controllers/operatorpolicy_status.go
JustinKuli: That seems like a good phrasing.

controllers/operatorpolicy_status.go
```diff
@@ -211,8 +211,10 @@ func calculateComplianceCondition(policy *policyv1beta1.OperatorPolicy) metav1.C
 	}
 
 	idx, cond = policy.Status.GetCondition(installPlanConditionType)
+
```
JustinKuli, line 214: Nit: can you remove these extra newlines? I don't think they are needed by the linter; the other similar stanzas don't have them.

controllers/operatorpolicy_status.go
```diff
@@ -775,6 +794,13 @@ func buildDeploymentCond(
 		message = fmt.Sprintf("the deployments %s do not have their minimum availability", names)
 	}
 
+	// If the status changed, that means the condition should be NonCompliant
+	// Apply override if complianceConfig is set to Compliant
+	if status == metav1.ConditionFalse && complianceConfig == "Compliant" {
+		status = metav1.ConditionTrue
+		message += " (policy compliance is not impacted due to spec.complianceConfig.deploymentsUnavailable)"
+	}
```
JustinKuli, line 802: Nit: I think the `status == metav1.ConditionFalse` could be removed if this was moved into the `if len(unavailableDeps) != 0` section.

controllers/operatorpolicy_status.go
```diff
@@ -1059,10 +1104,20 @@ func existingInstallPlanObj(ip client.Object, phase string) policyv1.RelatedObje
...
 	case string(operatorv1alpha1.InstallPlanPhaseRequiresApproval):
-		// FUTURE: check policy.spec.statusConfig.upgradesAvailable to determine `compliant`.
-		// For now, assume it is set to 'NonCompliant'
-		relObj.Compliant = string(policyv1.NonCompliant)
+		// Check policy.spec.statusConfig.upgradesAvailable to determine `compliant`.
+		if complianceConfig != "Compliant" {
+			relObj.Compliant = string(policyv1.NonCompliant)
+		} else {
+			relObj.Compliant = string(policyv1.Compliant)
+		}
```
JustinKuli, line 1120: Should these also append to the reason, like we do for the condition messages?

JustinKuli (commented):
Looking good overall, I'd be ok merging this if Matt's discussions are all resolved... I do have some nits, and I'm wondering if we should add info to the related objects when their compliance is affected by the complianceConfig. But that's not strictly necessary to do now.

controllers/operatorpolicy_status.go
zyjjay: Yea this could go either way. I was debating on this but ended up putting it outside the `if len(unavailableDeps) != 0` to avoid nested if statements. I'm leaning towards the current implementation in case in the future we decide to add more cases which would make `status` false.

controllers/operatorpolicy_status.go
zyjjay: Correct me if I'm wrong but I think the resolution for this is similar to above conversations (for installPlanUpgradeCond)? Would it be fine if I modified message to say something like `(policy compliance is not impacted due to spec.complianceConfig.catalogSourceUnhealthy)`?

controllers/operatorpolicy_status.go
JeffeyL: I think this makes sense, I'll add it to the other relatedObj `reason` messages as well for consistency.

controllers/operatorpolicy_status.go
JustinKuli: That's a good point :+1:

JeffeyL:
Per Matt's previous mention, this PR (https://github.com/open-cluster-management-io/governance-policy-framework-addon/pull/145)  in the framework addon addresses the template sync for default values of complianceConfig

JustinKuli (approved):
/hold for @mprahl 

LGTM

api/v1beta1/operatorpolicy_types.go
```diff
@@ -143,6 +152,12 @@ type OperatorPolicySpec struct {
 	// approval is not affected by this setting. This setting has no effect when the policy is in
 	// 'mustnothave' mode. Allowed values are "None" or "Automatic".
 	UpgradeApproval string `json:"upgradeApproval"`
+
+	//+kubebuilder:default={}
+	// ComplianceConfig defines how resource statuses affect the OperatorPolicy status and compliance.
+	// When set to Compliant, the condition does not impact the OperatorPolicy compliance.
+	// When set to NonCompliant, the condition cause the OperatorPolicy to become NonCompliant.
```
mprahl, line 159:
```suggestion
	// When set to NonCompliant, the condition causew the OperatorPolicy to become NonCompliant.
```

controllers/operatorpolicy_status.go
```diff
@@ -1059,10 +1104,22 @@ func existingInstallPlanObj(ip client.Object, phase string) policyv1.RelatedObje
 	}
 
 	switch phase {
+	case string(operatorv1alpha1.InstallPlanPhaseInstalling):
+		// Check policy.spec.statusConfig.upgradesAvailable to determine `compliant`.
+		if complianceConfig != "Compliant" {
+			relObj.Compliant = string(policyv1.NonCompliant)
+		} else {
+			relObj.Compliant = string(policyv1.Compliant)
+			relObj.Reason += " (policy compliance is not impacted due to spec.complianceConfig.upgradesAvailable)"
```
mprahl, line 1113: I find this message to be quite noisy for a default behavior. My thought is that we only need to include a message such as this when the default is overridden.

mprahl: @JustinKuli what do you think?

controllers/operatorpolicy_status.go
```diff
@@ -1059,10 +1104,22 @@ func existingInstallPlanObj(ip client.Object, phase string) policyv1.RelatedObje
...
 	case string(operatorv1alpha1.InstallPlanPhaseRequiresApproval):
-		// FUTURE: check policy.spec.statusConfig.upgradesAvailable to determine `compliant`.
-		// For now, assume it is set to 'NonCompliant'
-		relObj.Compliant = string(policyv1.NonCompliant)
+		// Check policy.spec.statusConfig.upgradesAvailable to determine `compliant`.
+		if complianceConfig != "Compliant" {
+			relObj.Compliant = string(policyv1.NonCompliant)
+		} else {
+			relObj.Compliant = string(policyv1.Compliant)
+			relObj.Reason += " (policy compliance is not impacted due to spec.complianceConfig.upgradesAvailable)"
```
mprahl, line 1121: Same as above.

api/v1beta1/operatorpolicy_types.go
JeffeyL: Can I assume you meant "causes"?

mprahl: Yup, you caught it before the edit :laughing:

controllers/operatorpolicy_status.go
JustinKuli: I don't think it's prohibitively noisy, but I see what you mean with default behaviors. So then should we have a message on the NonCompliant case? Maybe "(in violation due to spec.complianceConfig.upgradesAvailable)"?

controllers/operatorpolicy_status.go
```diff
@@ -1187,20 +1252,38 @@ var noExistingDeploymentObj = policyv1.RelatedObject{
...
 	reason := reasonWantFoundExists
 
 	if isUnhealthy {
-		compliance = string(policyv1.NonCompliant)
 		reason = reasonWantFoundExists + " but is unhealthy"
+
+		if complianceConfig != "Compliant" {
+			compliance = string(policyv1.NonCompliant)
+		} else {
+			reason += " (policy compliance is not impacted due to spec.complianceConfig.catalogSourceUnhealthy)"
```
JustinKuli, line 1275: @mprahl how do you feel about this one. I think "___ exists but is unhealthy" on its own in a Compliant object could be confusing. It is the default behavior, but I feel like mentioning why "unhealthy" is ok in this case is good.

JustinKuli (approved):
/unhold

LGTM, I think you covered Matt's suggestions, thanks

PR #253, opened 2024-05-23
Add the recreateOption to the object template

mprahl:
When a user needs to update an object's immutable fields, the object must be replaced. The user may opt-in to setting recreateOption on an object template to "IfRequired" and "Always". When set to "IfRequired", normal updates proceed when possible.

Relates:
https://issues.redhat.com/browse/ACM-11846

### DIFF
--- a/api/v1/configurationpolicy_types.go
+++ b/api/v1/configurationpolicy_types.go
@@ -195,6 +195,15 @@ type ObjectTemplate struct {
 
 	MetadataComplianceType MetadataComplianceType `json:"metadataComplianceType,omitempty"`
 
+	// RecreateOption describes whether to delete and recreate an object when an update is required. `IfRequired`
+	// will recreate the object when updating an immutable field. `Always` will always recreate the object if a mismatch
+	// is detected. `RecreateOption` has no effect when the `remediationAction` is `inform`. `IfRequired` has no effect
+	// on clusters without dry run update support. The default value is `None`.
+	//
+	//+kubebuilder:validation:Enum=None;IfRequired;Always
+	//+kubebuilder:default=None
+	RecreateOption RecreateOption `json:"recreateOption,omitempty"`
+
 	// ObjectDefinition defines required fields for the object
 	// +kubebuilder:pruning:PreserveUnknownFields
 	ObjectDefinition runtime.RawExtension `json:"objectDefinition"`
@@ -339,6 +348,14 @@ func (c ComplianceType) IsMustNotHave() bool {
 // MetadataComplianceType describes how to check compliance for the labels/annotations of a given object
 // +kubebuilder:validation:Enum=MustHave;Musthave;musthave;MustOnlyHave;Mustonlyhave;mustonlyhave
 type MetadataComplianceType string
+
+type RecreateOption string
+
+const (
+	None       RecreateOption = "None"
+	IfRequired RecreateOption = "IfRequired"
+	Always     RecreateOption = "Always"
+)
 
 // RelatedObject is the list of objects matched by this Policy resource.
 type RelatedObject struct {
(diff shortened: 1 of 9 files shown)


api/v1/configurationpolicy_types.go
```diff
@@ -340,6 +342,19 @@ func (c ComplianceType) IsMustNotHave() bool {
 // +kubebuilder:validation:Enum=MustHave;Musthave;musthave;MustOnlyHave;Mustonlyhave;mustonlyhave
 type MetadataComplianceType string
 
+// RecreateOption describes the condition when to delete and recreate an object when an update is required. IfRequired
```
mprahl, line 345: @dockerymick could you please review this text?

yiraeChristineKim: Could you add default behaviour?

controllers/configurationpolicy_controller.go
```diff
@@ -2668,32 +2683,70 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
...
+						)
+					}
 
-			diff = handleDiff(log, recordDiff, isInform, existingObjectCopy, dryRunUpdatedObj)
+					return true, message, diff, false, nil
+				}
+			} else {
+				removeFieldsForComparison(dryRunUpdatedObj)
+
+				if reflect.DeepEqual(dryRunUpdatedObj.Object, existingObjectCopy.Object) {
```
yiraeChristineKim, line 2737: Why is this log mismatch message?

mprahl: What do you mean?

controllers/configurationpolicy_controller.go
```diff
@@ -2706,18 +2759,56 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
...
+
+			for {
+				updatedObj, err = res.Create(context.TODO(), &obj.desiredObj, metav1.CreateOptions{})
+				if !k8serrors.IsAlreadyExists(err) {
+					// If there is no error or the error is unexpected, break for the error handling below
+					break
+				}
+
+				if time.Since(start) > time.Second*10 {
+					message = fmt.Sprintf(
```
yiraeChristineKim, line 2790: In this for loop, it doesn't delete the obj. so the message should be changed?

yiraeChristineKim (approved):

controllers/configurationpolicy_controller.go
```diff
@@ -1791,15 +1806,18 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 			throwSpecViolation, msg, diff, triedUpdate, updatedObj = r.checkAndUpdateResource(
 				obj, objectT, remediation,
 			)
+
+			if updatedObj != nil && string(updatedObj.GetUID()) != uid {
+				uid = string(updatedObj.GetUID())
+				created = true
+			}
```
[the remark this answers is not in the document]
mprahl: Yes, exactly. So if the UID changes, the objectProperties will get the new UID.

api/v1/configurationpolicy_types.go
```diff
@@ -340,6 +342,19 @@ func (c ComplianceType) IsMustNotHave() bool {
...
+// will recreate the object when updating an immutable field. Always will always recreate the object if a mismatch is
+// detected. RecreateOption has no effect when the remediationAction is inform. IfRequired has no effect on clusters
+// without dry run update support. Default is None.
+// +kubebuilder:validation:Enum=None;IfRequired;Always
+type RecreateOption string
```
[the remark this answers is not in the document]
mprahl: This is the code that needs updating but it's not a big deal. I have the local changes queued up. Just need to run the tests once this merges:
https://github.com/open-cluster-management-io/governance-policy-framework-addon/blob/f299b7020823b3a09ab872ea4731080b42ff09d8/controllers/templatesync/template_sync.go#L959-L967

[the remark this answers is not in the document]
JustinKuli: Reading that makes me think that template-sync might be thinking it needs to do more updates to OperatorPolicy than it does. I thought I had checked and it wasn't stuck in a loop at least, but I'll need to investigate more

PR #260, opened 2024-06-06
Add support for approving InstallPlans with multiple CSVs

mprahl:
# First Commit

The operatorpolicy.policy.open-cluster-management.io/managed label is set to an empty value on the Subscription. The
operatorpolicy.policy.open-cluster-management.io/managed annotation is set to <policy name>.<policy namespace>.

The label doesn't set a value because the value could be too long.

This will allow querying for subscriptions managed by operator policies.

# Second Commit

When an InstallPlan contains multiple CSVs, it will be approved if all CSVs can be approved by the current operator policies.

Rather than query the operator policies directly, the subscriptions in the namespace that are managed by an operator policy are queried and then their associated operator policies are queried to avoid having to resolve templates.

Relates:
https://issues.redhat.com/browse/ACM-11981

### DIFF
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -12,6 +12,7 @@ import (
 	"reflect"
 	"regexp"
 	"slices"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -28,8 +29,10 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/dynamic"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -1443,10 +1446,19 @@ func (r *OperatorPolicyReconciler) handleInstallPlan(
 		return updateStatus(policy, invalidCausingUnknownCond("InstallPlan")), nil
 	}
 
-	watcher := opPolIdentifier(policy.Namespace, policy.Name)
-	selector := subLabelSelector(sub)
+	if sub.Status.CurrentCSV == "" && sub.Status.InstalledCSV == "" {
+		return updateStatus(policy, noInstallPlansCond, noInstallPlansObj(sub.Namespace)), nil
+	}
 
-	installPlans, err := r.DynamicWatcher.List(watcher, installPlanGVK, sub.Namespace, selector)
+	watcher := opPolIdentifier(policy.Namespace, policy.Name)
+
+	// An InstallPlan will contain all CSVs that can be installed/updated in the namespace at the time the InstallPlan
+	// is created. The latest is denoted by the highest `spec.generation`, though `creationTimestamp` would likely
+	// also work. If the catalog is updated so that `status.currentCSV` on the Subscription does not point to an
+	// existing CSV, `status.currentCSV` will get included in the new InstallPlan. A new Subscription will also trigger
+	// a new InstallPlan. This code will always pick the latest InstallPlan so as to avoid installing older and
+	// potentially vulnerable operators.
+	installPlans, err := r.DynamicWatcher.List(watcher, installPlanGVK, sub.Namespace, labels.Everything())
 	if err != nil {
 		return false, fmt.Errorf("error listing InstallPlans: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


controllers/operatorpolicy_controller.go
```diff
@@ -1426,187 +1470,346 @@ func (r *OperatorPolicyReconciler) handleInstallPlan(
...
+		if !matchFound {
+			continue
+		}
+
+		// If the InstallPlan contains the current CSV and the phase indicates it is installing or complete, always
+		// use that InstallPlan since race conditions in OLM can cause a newer InstallPlan to be created with this CSV.
+		if containsCurrentCSV {
+			phase, _, _ := unstructured.NestedString(installPlans[i].Object, "status", "phase")
+			ipInstalling := string(operatorv1alpha1.InstallPlanPhaseInstalling)
+			ipComplete := string(operatorv1alpha1.InstallPlanPhaseComplete)
```
JustinKuli, line 1512: I understand prioritizing the Installing plan, but why also Complete plans? It feels like there can be multiple old InstallPlans that have that one, if the operator is done installing?

mprahl: @JustinKuli if an InstallPlan is complete for `currentCSV` (not `installedCSV`) then that means this InstallPlan successfully installed the latest CSV in the channel for this subscription. Otherwise, due to OLM race conditions, it's possible that there are multiple InstallPlans for `currentCSV` where one is approved and completed and one is awaiting approval.

controllers/operatorpolicy_controller.go
```diff
@@ -1426,187 +1470,346 @@ func (r *OperatorPolicyReconciler) handleInstallPlan(
...
 	}
 
-	if err := r.TargetClient.Update(ctx, &approvableInstallPlans[0]); err != nil {
-		return false, fmt.Errorf("error updating approved InstallPlan: %w", err)
+	allowedCSVs := make([]policyv1.NonEmptyString, 0, len(policy.Spec.Versions)+1)
+	allowedCSVs = append(allowedCSVs, policy.Spec.Versions...)
+
+	if sub.Spec.StartingCSV != "" {
+		allowedCSVs = append(allowedCSVs, policyv1.NonEmptyString(sub.Spec.StartingCSV))
+	}
```
JustinKuli, line 1795: As a weird note, previously `allowedVersions` was the variable name.

In this review, it is difficult to tell what is new-new code and what is small changes to existing pieces. At least in GH, it's not very good at aligning the old and new... This is obviously not a major issue, but if there were some additional "anchor points" where there weren't changes (these six lines, for example), it might help `git blame` in the future.

controllers/operatorpolicy_controller.go
```diff
@@ -1426,187 +1470,346 @@ func (r *OperatorPolicyReconciler) handleInstallPlan(
...
+
+	// All versions for the operator are allowed if no version is specified
+	if len(policy.Spec.Versions) == 0 {
+		// currentCSV is the CSV related to the latest InstallPlan
+		if sub.Status.CurrentCSV != "" {
+			return sub.Status.CurrentCSV
+		}
```
JustinKuli, line 1802: Is there a guarantee that `sub.Status.CurrentCSV` is in `csvNames`?

JustinKuli: It feels like there could be items in `csvNames` that "should" be approved by a policy with no versions specified - I thought we were considering some sort of name matching in this case. Is that somewhere else? Or is it not necessary?

controllers/operatorpolicy_controller.go
```diff
@@ -1426,187 +1470,346 @@ func (r *OperatorPolicyReconciler) handleInstallPlan(
...
+		// an old installation. But if the current InstallPlan is failed, we should alert the user.
+		if len(ipCSVs) == 1 {
+			return updateStatus(
+				policy, installPlanFailed, existingInstallPlanObj(&latestInstallPlan, string(phase), complianceConfig),
+			), nil
+		}
+
+		// If there is more than one CSV in the InstallPlan, make sure this CSV failed before marking it as
+		// noncompliant.
+		for _, step := range latestInstallPlan.Status.Plan {
```
mprahl, line 1591: @JustinKuli I read the OLM code and it seems that the only successful final states are created (even if it updated) and present. Do you think it's best to do this filtering or should we just mark the InstallPlan as failed even if this operator was successful?

I don't really know how I can get test coverage on this because `clusterServiceVersionNames` order in the InstallPlan are not deterministic so even if I have an InstallPlan with a failing operator, that one may have been processed first and not move on to the successful operator. I suppose I could modify the InstallPlan status to order the expected successful one but that seems quite hacky.

JustinKuli: I didn't fully understand this part, so if you think it can work without it, that would be fine? I don't like the idea of modifying the InstallPlan status

controllers/operatorpolicy_controller.go
mprahl: Here are the breadcrumbs:
https://github.com/operator-framework/operator-lifecycle-manager/blob/889ae50a66fe36a46a3c756a0ad7b481b73672f1/pkg/controller/operators/catalog/operator.go#L1291
https://github.com/operator-framework/operator-lifecycle-manager/blob/889ae50a66fe36a46a3c756a0ad7b481b73672f1/pkg/controller/registry/resolver/step_resolver.go#L123-L127
https://github.com/operator-framework/operator-lifecycle-manager/blob/889ae50a66fe36a46a3c756a0ad7b481b73672f1/pkg/controller/registry/resolver/step_resolver.go#L132-L136
https://github.com/operator-framework/operator-lifecycle-manager/blob/889ae50a66fe36a46a3c756a0ad7b481b73672f1/pkg/controller/registry/resolver/step_resolver.go#L197-L204
https://github.com/operator-framework/operator-lifecycle-manager/blob/889ae50a66fe36a46a3c756a0ad7b481b73672f1/pkg/controller/operators/catalog/operator.go#L1428-L1433

controllers/operatorpolicy_controller.go
mprahl: @JustinKuli to clarify, modifying the InstallPlan status would be for test coverage.

So the question is, should an InstallPlan be marked as failed in the OperatorPolicy if the overall InstallPlan failed but the part for the operator being managed by OperatorPolicy succeeded?

controllers/operatorpolicy_controller.go
mprahl: The side by side diff is easier for this.

controllers/operatorpolicy_controller.go
mprahl: @JustinKuli I'll revise my statement of "latest in the channel" but rather the latest that replaces the current version.

JustinKuli (approved):
/hold

PR #266, opened 2024-06-10
Use suggested namespaces of packages

JustinKuli:
(Originally this PR had 2 commits - 1 has been removed to be handled later if necessary)

Previously, if the namespace was not set for a Subscription in an
OperatorPolicy, the default namespace on the controller (if set) would
be used in all cases. But many packages specify a suggested namespace in
their PackageManifests, which would be better to use, when provided.

Additionally, if the OperatorPolicy was already managing a subscription,
the namespace of that subscription will be used (if not set in the
policy). This may prevent some issues that could occur if the suggested
namespace for an operator changes at some point.

Refs:
 - https://issues.redhat.com/browse/ACM-12057

### DIFF
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -396,9 +396,9 @@ func (r *OperatorPolicyReconciler) buildResources(ctx context.Context, policy *p
 		opLog.V(1).Info("Templates disabled by annotation")
 	}
 
-	sub, subErr := buildSubscription(policy, r.DefaultNamespace, tmplResolver)
+	sub, subErr := buildSubscription(policy, tmplResolver)
 	if subErr == nil {
-		err := r.applySubscriptionDefaults(ctx, sub)
+		err := r.applySubscriptionDefaults(ctx, policy, sub)
 		if err != nil {
 			sub = nil
 
@@ -409,6 +409,14 @@ func (r *OperatorPolicyReconciler) buildResources(ctx context.Context, policy *p
 			}
 
 			validationErrors = append(validationErrors, err)
+		}
+
+		if sub != nil && sub.Namespace == "" {
+			if r.DefaultNamespace != "" {
+				sub.Namespace = r.DefaultNamespace
+			} else {
+				validationErrors = append(validationErrors, errors.New("namespace is required in spec.subscription"))
+			}
 		}
 	} else {
 		validationErrors = append(validationErrors, subErr)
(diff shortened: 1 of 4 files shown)


api/v1beta1/operatorpolicy_types.go
```diff
@@ -204,14 +204,19 @@ type OperatorPolicyStatus struct {
...
 	RelatedObjects []policyv1.RelatedObject `json:"relatedObjects"`
 
-	// The resolved name.namespace of the subscription
+	// The resolved name.namespace of the subscription.
 	ResolvedSubscriptionLabel string `json:"resolvedSubscriptionLabel,omitempty"`
 
+	// The value of the 'operatorpolicy.policy.open-cluster-management.io/namechange' annotation
+	// when the operator name and namespace were most recently changed. This field is used to
+	// prevent some unintentional changes, for example caused by templates.
+	ResolvedSubscriptionLock string `json:"resolvedSubscriptionLock,omitempty"`
```
JustinKuli, line 213: I'm not sure if "lock" or the specifics of this annotation are ideal, I'm open to different names or other suggestions!

controllers/operatorpolicy_controller.go
```diff
@@ -595,6 +644,28 @@ func (r *OperatorPolicyReconciler) applySubscriptionDefaults(
 		subSpec.CatalogSourceNamespace = catalogNamespace
 	}
 
+	if subscription.Namespace == "" {
```
mprahl, line 647: Do you think we could prefer the namespace of the existing subscription if the subscription already exists?

JustinKuli:
/hold

Another approach I'm thinking about now (an alternative to the admittedly odd annotation thing) is to only prevent the name/space change when the policy is musthave & enforce. Then to "force" the change, the user could change the policy to inform briefly, then back to enforce. This would also give them a chance to make the policy mustnothave if they did want to change the namespace of the operator, and clean up the old installation.

@mprahl what do you think? Or it sounded like you were thinking preventing this kind of change might not be necessary? The situation I'm trying to prevent is where the operator's suggested namespace changes at some point, and OperatorPolicy creates a duplicate operator installation in the new namespace without any indication of the potential issue.

controllers/operatorpolicy_controller.go
```diff
@@ -55,6 +55,8 @@ const (
 	ManagedByAnnotation    string = ManagedByLabel
 	ClusterNameLabel       string = "policy.open-cluster-management.io/cluster-name"
 	ClusterNamespaceLabel  string = "policy.open-cluster-management.io/cluster-namespace"
+	SubLockAnnotation      string = "operatorpolicy.policy.open-cluster-management.io/namechange"
```
mprahl, line 58: I think I'd prefer if the user got a noncompliant message on the subscription if the name or namespace changed and that the solution is to create a new or recreate the old OperatorPolicy or uninstall the existing operator.

Having this annotation override is more confusing.

controllers/operatorpolicy_controller.go
mprahl: Is it not in the relatedObjects on the status?

JustinKuli: Not if you create the policy after installing the operator.

mprahl: What I'm saying is that if the subscription already exists, use that namespace, otherwise look up the default.

This is to prevent switching namespaces in the event there is no suggested namespace but one is added later.

mprahl (approved):

PR #272, opened 2024-07-03
Ignore imagePullSecrets and secrets on ServiceAccounts

JeffeyL:
When policy is mustonlyhave for a ServiceAccount, infinite secrets are created due to consistent mismatch of imagePullSecrets and secrets fields. Thus, an exception is made to ignore those fields for ServiceAccounts.

Ref: https://issues.redhat.com/browse/ACM-12270

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -2982,6 +2982,12 @@ func handleKeys(
 				continue
 			}
 
+			// for ServiceAccounts, ignore "secrets" and "imagePullSecrets" fields, as these are managed by Kubernetes
+			if (existingObj.GetKind() == "ServiceAccount" && existingObj.GetAPIVersion() == "v1") &&
+				(key == "secrets" || key == "imagePullSecrets") {
+				continue
+			}
+
 			delete(existingObj.Object, key)
 
 			updateNeeded = true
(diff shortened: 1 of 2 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -2982,6 +2982,11 @@ func handleKeys(
 				continue
 			}
 
+			// for ServiceAccounts, ignore "secrets" and "imagePullSecrets" fields, as these are managed by Kubernetes
+			if existingObj.GetKind() == "ServiceAccount" && (key == "secrets" || key == "imagePullSecrets") {
```
dhaiducek, line 2986: (optional) There's a small chance there could be a `ServiceAccount` kind in a non-K8s core group, so we could add that check, but I don't think it's necessary.

dhaiducek (approved):
/hold For minor comment

LGTM! Thanks for the update!

controllers/configurationpolicy_controller.go
JeffeyL: Would a check for `apiVersion == "v1"`  be sufficient for this?

dhaiducek (approved):
LGTM! Thanks for the update!

mprahl (approved):

PR #273, opened 2024-07-08
Fix help messages in compliance messages getting removed on next eval

mprahl:
When a help message is provided such as suggesting to set recreateOption, it was getting overwritten during the next evaluation based on the cached evaluation result.

Relates:
https://issues.redhat.com/browse/ACM-12631


dhaiducek (approved):
LGTM!

PR #274, opened 2024-07-08
Make ConfigurationPolicy event driven by default

mprahl:
The user can still set `evaluationInterval` to go back to the legacy
polling behavior.

This also removes the code to remove the legacy finalizer on the Deployment and the  common_related_objects metric.

Relates:
https://issues.redhat.com/browse/ACM-11666


api/v1/configurationpolicy_types.go
```diff
@@ -84,29 +85,34 @@ func (t Target) String() string {
 }
 
 // EvaluationInterval configures the minimum elapsed time before a configuration policy is
-// reevaluated. If the policy spec is changed, or if the list of namespaces selected by the policy
+// reevaluated. This defaults to "watch" to leverage Kubernetes API watches instead of polling the Kubernetes API
```
yiraeChristineKim, line 88: I have a question. When it is watch, how policy detect and change the status when the resource which is created by configuration policy is changed, created or deleted (ex: change status from compliant -> noncompliant)

mprahl: Any query using the `DynamicWatcher` client or from a template function executed in `TemplateResolver` causes a watch for that `ConfigurationPolicy` and triggers a reconcile of the `ConfigurationPolicy` when any of those objects change. Those reconciles are triggered from the additional source added in the `SetupWithManager` method.

pkg/common/namespace_selection.go
```diff
@@ -289,26 +311,55 @@ func (r *NamespaceSelectorReconciler) Get(name string, t policyv1.Target) ([]str
...
-	delete(r.selections, name)
+	delete(r.selections, getKey(namespace, name))
+}
+
+func getKey(namespace string, name string) string {
+	return fmt.Sprintf("%s/%s", namespace, name)
 }
 
-func (r *NamespaceSelectorReconciler) update(name string, sel namespaceSelection) {
+func splitKey(key string) (string, string) {
```
yiraeChristineKim, line 333: Is this return (namespace, name).. ? Can we write return value in function description (comment)

controllers/configurationpolicy_controller.go
```diff
@@ -130,247 +182,188 @@ type ConfigurationPolicyReconciler struct {
...
-func (r *ConfigurationPolicyReconciler) handlePolicyWorker(
-	policyQueue <-chan *policyv1.ConfigurationPolicy, wg *sync.WaitGroup,
-) {
-	defer wg.Done()
+	var requeueAfter time.Duration
 
-	for policy := range policyQueue {
-		before := time.Now().UTC()
+	//nolint: ineffassign,wastedassign
+	err = nil
```
JustinKuli, line 310: It's not clear if/why the linter is wrong here. I might just prefer a new error variable to clear up the confusion - a descriptive name might be all I need to understand it.

controllers/configurationpolicy_controller.go
```diff
@@ -130,247 +182,188 @@ type ConfigurationPolicyReconciler struct {
...
-	}
+			return reconcile.Result{}, nil
+		}
 
-	if (resourceErr == nil && groupErr == nil) || (len(apiGroups) > len(r.discoveryInfo.apiGroups)) {
-		// update the list if it's complete, or if it's "better" than the old one
-		r.discoveryInfo.apiGroups = apiGroups
+		log.Error(
+			err,
+			"The policy has an invalid evaluation interval; defaulting to watch",
```
JustinKuli, line 352: But hasn't it already removed the watches? In this case does it constantly requeue immediately?

controllers/configurationpolicy_controller.go
```diff
@@ -425,138 +433,82 @@ func (r *ConfigurationPolicyReconciler) shouldEvaluatePolicy(
...
+	if usesSelector && r.SelectorReconciler.HasUpdate(policy.Namespace, policy.Name) {
+		log.V(1).Info("There was an update for this policy's namespaces. Will evaluate it now.")
+
+		return true, 0
 	}
 
 	var interval time.Duration
 
+	//nolint: ineffassign,wastedassign
+	err = nil
```
JustinKuli, line 452: Same as previous, I would prefer a new error variable with a descriptive name.

controllers/configurationpolicy_controller.go
```diff
@@ -606,18 +554,31 @@ func (r *ConfigurationPolicyReconciler) cleanUpChildObjects(plc policyv1.Configu
...
+		// Note that GetFromCache does not start a watch if a watch wasn't present, unlike "Get".
+		existing, err := r.DynamicWatcher.GetFromCache(
+			gvk,
 			object.Object.Metadata.Namespace,
 			object.Object.Metadata.Name,
-			mapping.Resource,
-			r.TargetK8sDynamicClient,
 		)
+		if err != nil {
+			log.V(2).Error(
```
JustinKuli, line 568: I'm not sure what `.V(2).Error` will do combined... do you have an example output?

controllers/configurationpolicy_controller.go
```diff
@@ -783,41 +748,69 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc policyv1.Confi
...
 		policyUserErrorsCounter.WithLabelValues(parent, plc.GetName(), "invalid-template").Add(1)
 
-		return
+		return fmt.Errorf("%w: %s", ErrPolicyInvalid, validationErr)
 	}
 
-	// object handling for when configurationPolicy is deleted
-	if plc.Spec.PruneObjectBehavior == "DeleteIfCreated" || plc.Spec.PruneObjectBehavior == "DeleteAll" {
-		var crdDeleting bool
+	//nolint: wastedassign
```
JustinKuli, line 754: Does using `var usingWatch bool` resolve the linter?

JustinKuli: Times like this I wish go had a ternary operator

JustinKuli (changes requested):
Mostly small-ish things, I think it's good overall (and of course the passing tests are a great sign).

This should be really nice for performance, likely reacting more quickly to changes in the cluster, while also doing fewer repetitive and wasted computations.

api/v1/configurationpolicy_types.go
```diff
@@ -282,7 +296,7 @@ type ConfigurationPolicySpec struct {
 
 // ComplianceState reports the observed status from the definitions of the policy.
 //
-// +kubebuilder:validation:Enum=Compliant;Pending;NonCompliant;Terminating
+// +kubebuilder:validation:Enum=Compliant;Pending;NonCompliant;Terminating;UnknownCompliancy
```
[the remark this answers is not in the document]
mprahl: @JustinKuli there was a bug in the code during development and the status update was failing because `UnknownCompliancy` was not a valid value. I just added it here since I thought it was an oversight since it is a value the status update code can set.

Is it your preference that I remove it? It should not be possible to get this value in ConfigurationPolicy but I wondered if having a status of UnknownCompliancy is better than the status update failing if a bug is introduced.

main.go
```diff
@@ -635,10 +673,10 @@ func parseOpts(flags *pflag.FlagSet, args []string) *ctrlOpts {
 	opts := &ctrlOpts{}
 
 	flags.UintVar(
-		&opts.frequency,
-		"update-frequency",
+		&opts.maxEvalFrequencySeconds,
+		"max-evaluation-frequency",
 		10,
-		"The status update frequency (in seconds) of a mutation policy",
+		"The number of seconds before a policy is eligible for reevaluation (throttles frequently evaluated policies)",
```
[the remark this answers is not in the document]
mprahl: After getting Dale's opinion, I'm changing this to `evaluation-backoff`.

controllers/configurationpolicy_controller.go
mprahl: Good point! Fixing.

controllers/configurationpolicy_controller.go
dhaiducek: I agree--this line isn't necessary because everywhere below this it's reassigned before use, so this assignment doesn't need to be here.

controllers/configurationpolicy_controller.go
mprahl: I thought it changed the log level but still display the stack trace but that's not documented behavior from what I can tell, so I'll change it to an info log with the error message as a key/value.

controllers/configurationpolicy_controller.go
mprahl: Setting this to `var usingWatch bool` seems to work. I tend to initialize to false based on previous PR review feedback as newcomers to Go don't know that `var usingWatch bool` is equivalent to a false boolean.

controllers/configurationpolicy_controller.go
```diff
@@ -81,12 +88,65 @@ const (
...
+
+					newTyped, ok := e.ObjectNew.(*policyv1.ConfigurationPolicy)
+					if !ok {
+						return false
+					}
+
+					// Handle the case where compliance was explicitly reset by the governance-policy-framework
+					return oldTyped.Status.ComplianceState != "" && newTyped.Status.ComplianceState == ""
+				},
+				DeleteFunc: func(event event.DeleteEvent) bool {
```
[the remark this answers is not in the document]
mprahl: If there is a finalizer, the initial deletion request causes `UpdateFunc` to be called. When the finalizer is removed and the object is deleted, `DeleteFunc` is called.

This leads me to believe that `UpdateFunc` should always return true if the deletion timestamp is set on the ConfigurationPolicy. Thanks for asking about this.

controllers/configurationpolicy_controller.go
dhaiducek: Right, and I agree that's a good practice, but I think it's complaining because there's a conditional block where no matter what happens, it's getting assigned, so it doesn't matter what it's initialized to.

main.go
```diff
@@ -387,20 +394,36 @@ func main() {
...
+		nsSelReconciler = common.NewNamespaceSelectorReconciler(nsSelMgr.GetClient(), nsSelUpdatesChan)
 		if err = nsSelReconciler.SetupWithManager(nsSelMgr); err != nil {
 			log.Error(err, "Unable to create controller", "controller", "NamespaceSelector")
 			os.Exit(1)
 		}
 
 		discoveryClient := discovery.NewDiscoveryClientForConfigOrDie(targetK8sConfig)
 
-		serverVersion, err := discoveryClient.ServerVersion()
+		var err error
```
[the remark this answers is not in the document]
mprahl: I like this since it doesn't overwrite the `err` value from the above scope.

api/v1/configurationpolicy_types.go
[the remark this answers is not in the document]
JustinKuli: We can keep it for completeness. Do you mean that it can be set by status-sync or something else in the framework?

If we were making it from scratch, my preference would just be to have an empty value imply an unknown compliance, as opposed to a special value. But if other things already use UnknownCompliancy, then it's not worth changing everywhere, and it should at least be mentioned here for user clarity.

main.go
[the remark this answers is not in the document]
JustinKuli: So it shadows the old `err`, instead of overwriting it? Maybe this is another case where it just needs a different name.

But this is pretty minor - we can keep it as-is if you really prefer.

controllers/configurationpolicy_controller.go
JustinKuli: Yeah the `var` syntax can be annoying... I just dislike `//nolint` even more (when avoidable).

JustinKuli: Yeah the `var` syntax can be annoying... I just dislike `//nolint` even more (when avoidable).

main.go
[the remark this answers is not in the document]
dhaiducek: Actually, if I'm looking at it right, I think this is correct. The old code was also scoping `serverVersion` such that it would not be populated for the `reconciler` variable farther down...

api/v1/configurationpolicy_types.go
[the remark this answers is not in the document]
mprahl: I'll just remove it.

dhaiducek (commented):
Beginnings of a review--I have some feedback on the CRD descriptions 🙂

mprahl:
Okay, the latest push fixes the CI failure of when an object deletion hangs from prune object behavior.

JustinKuli (approved):
/hold

I believe all of my requests were addressed and this is still looking good. Holding for any other reviews.

yiraeChristineKim (approved):

PR #275, opened 2024-07-17
Retry evaluating the policy if a mapping error occurs

mprahl:
Since there is no watch for the condition to resolve itself, the policy would never become compliant once the CRD is available.

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -306,6 +306,15 @@ func (r *ConfigurationPolicyReconciler) Reconcile(ctx context.Context, request c
 			return reconcile.Result{}, nil
 		}
 
+		// If a mapping error occurred, try again in 10 seconds to see if the CRD is available
+		if errors.Is(handleErr, depclient.ErrNoVersionedResource) &&
+			policy.Spec != nil &&
+			policy.Spec.EvaluationInterval.IsWatchForNonCompliant() {
+			log.Info("Requeuing the policy to be reevalauted in 10 seconds due to a mapping error")
+
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		}
+
 		return reconcile.Result{}, handleErr
 	}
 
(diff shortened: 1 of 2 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -88,7 +88,10 @@ const (
 	reasonFoundNotApplicable = "Resource found but will not be handled in mustnothave mode"
 )
 
-var ErrPolicyInvalid = errors.New("the Policy is invalid")
+var (
+	ErrPolicyInvalid = errors.New("the Policy is invalid")
+	ErrMappingError  = errors.New("a mapping occurred")
```
dhaiducek, line 93: Is there a word missing here?
```suggestion
	ErrMappingError  = errors.New("a mapping error occurred")
```

mprahl: Oops. Fixed!

controllers/configurationpolicy_controller.go
```diff
@@ -1172,6 +1186,10 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc *policyv1.Conf
 		// map raw object to a resource, generate a violation if resource cannot be found
 		if decodeErrResult == nil {
 			scopedGVR, mappingErrResult = r.getMapping(desiredObj.GroupVersionKind(), plc, indx)
+
+			if mappingErrResult != nil {
+				mappingError = ErrMappingError
+			}
```
dhaiducek, line 1192: It appears the first error case returned from `getMapping()` is a missing API version, which shouldn't be re-queued since it's an invalid Policy, so it'd be best to handle that one differently.

mprahl: Good catch! Fixed in the latest push.

yiraeChristineKim (approved):

mprahl:
@yiraeChristineKim and @dhaiducek could you please look again? I definitely rushed this PR but now it seems to be working well.

yiraeChristineKim (approved):

mprahl:
@yiraeChristineKim I made a small change to the test which should get the same result but make it 10 seconds faster.

dhaiducek:
> @yiraeChristineKim I made a small change to the test which should get the same result but make it 10 seconds faster.

@mprahl But that defeats the purpose of the test, which is looking for a retry (i.e. at least 2 total)

dhaiducek (approved):

PR #276, opened 2024-07-18
Refactor to split up handleObjectTemplates and help understandability

JustinKuli:
I've tried to make each of the commits understandable on their own, but the major reorganization in the big function was difficult to read no matter how I split it up.

There are two commits with changes beyond just refactoring:
- [Assume dry run is always supported](https://github.com/open-cluster-management-io/config-policy-controller/commit/3873da64679935346a1a6ca9dc25b438b0c5e9ab)
- [Cache empty selectors and errors in NS reconciler](https://github.com/open-cluster-management-io/config-policy-controller/commit/feec0d09c271a2450ad3be8b9513d20c7de83260)

I would be happy to walk through any changes if it would help.

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -1577,7 +1577,7 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 			scopedGVR,
 			namespace,
 			r.TargetK8sDynamicClient,
-			strings.ToLower(string(objectT.ComplianceType)),
+			objectT.ComplianceType,
 			// Dry run API requests aren't run on unnamed object templates for performance reasons, so be less
 			// conservative in the comparison algorithm.
 			true,
@@ -1951,7 +1951,7 @@ func (r *ConfigurationPolicyReconciler) getMapping(
 // buildNameList is a helper function to pull names of resources that match an objectTemplate from a list of resources
 func buildNameList(
 	desiredObj unstructured.Unstructured,
-	complianceType string,
+	complianceType policyv1.ComplianceType,
 	resList *unstructured.UnstructuredList,
 	zeroValueEqualsNil bool,
 ) (kindNameList []string) {
@@ -1989,7 +1989,7 @@ func (r *ConfigurationPolicyReconciler) getNamesOfKind(
 	scopedGVR depclient.ScopedGVR,
 	ns string,
 	dclient dynamic.Interface,
-	complianceType string,
+	complianceType policyv1.ComplianceType,
 	zeroValueEqualsNil bool,
 	useCache bool,
 ) (kindNameList []string, allResourceList []string) {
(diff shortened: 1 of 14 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -1469,11 +1469,23 @@ func addConditionToStatus(
...
+
+	if len(currentConds) == 0 {
+		plc.Status.CompliancyDetails[index].Conditions = []policyv1.Condition{*newCond}
 		updateNeeded = true
+	} else {
+		oldCond := currentConds[len(currentConds)-1]
+		newConditionIsSimilar := reflect.DeepEqual(oldCond.Status, newCond.Status) &&
+			reflect.DeepEqual(oldCond.Reason, newCond.Reason) &&
+			reflect.DeepEqual(oldCond.Message, newCond.Message) &&
+			reflect.DeepEqual(oldCond.Type, newCond.Type)
```
mprahl, line 1476: Why is reflect.DeepEqual used since these are strings?

mprahl: I see it was a cut and paste from another helper function. Not blocking then.

controllers/configurationpolicy_controller.go
```diff
@@ -791,13 +751,196 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc *policyv1.Conf
...
+
+	r.addForUpdate(plc, parentStatusUpdateNeeded)
+
+	return mappingErr
+}
+
+// validateConfigPolicy returns an error and increments the "invalid-template" error counter metric
+// if the configuration is invalid.
+func (r *ConfigurationPolicyReconciler) validateConfigPolicy(plc *policyv1.ConfigurationPolicy) error {
+	var invalidMessage string
```
mprahl, line 901:
```suggestion
	log := log.WithValues("policy", plc.GetName())

	var invalidMessage string
```

controllers/configurationpolicy_controller.go
```diff
@@ -940,378 +1106,250 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc *policyv1.Conf
...
-					addTemplateErrorViolation("", msg)
+	// process object templates for go template usage
+	for i, rawData := range rawDataList {
+		if !templates.HasTemplate(rawData, "", true) && !isRawObjTemplate {
+			continue
+		}
 
-					return returnedErr
-				}
+		if !templates.HasTemplate(rawData, "", true) {
```
mprahl, line 1138: I'd like to avoid calling `HasTemplate` a second time since this can be expensive as it scans the whole string.

controllers/configurationpolicy_controller.go
```diff
@@ -963,411 +1106,259 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc *policyv1.Conf
...
+			// Unmarshal raw template YAML into object if that has not already been done by the template
```
mprahl, line 1139:
```suggestion
			// Unmarshal raw template YAML into object
```

mprahl (approved):
Great job! This is much cleaner.

I left a few minor comments but they can be addressed separately to not hold up this PR.

PR #277, opened 2024-07-19
Fix order of deleting ConfigPolicies in test

dhaiducek:
This was causing some flaky test failures. The enforced policy was sometimes recreating the 'policy-pod-to-check' config policy after it was deleted.

Commit migrated from:
- #276


mprahl:
Ended up merging #276 so we can close this one. Thanks though!

PR #284, opened 2024-08-05
Use the controller-runtime cache to get the decryption key

mprahl:
The stale cached decryption key detection logic was not 100% accurate, so just use the controller-runtime cache every time.

Relates:
https://issues.redhat.com/browse/ACM-11497


JustinKuli (approved):
LGTM

PR #285, opened 2024-08-06
Ensure pod restart when target kubeconfig changes

zyjjay:
ref: https://issues.redhat.com/browse/ACM-12933

### DIFF
--- a/main.go
+++ b/main.go
@@ -39,6 +39,7 @@ import (
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 	"open-cluster-management.io/addon-framework/pkg/lease"
+	addonutils "open-cluster-management.io/addon-framework/pkg/utils"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -355,6 +356,7 @@ func main() {
 	var targetK8sConfig *rest.Config
 	var targetClient client.Client
 	var nsSelMgr manager.Manager // A separate controller-manager is needed in hosted mode
+	var configFiles []string
 
 	if opts.targetKubeConfig == "" {
 		targetK8sConfig = cfg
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -529,9 +530,36 @@ func main() {
...
+
+		if targetK8sConfig.TLSClientConfig.CertFile != "" {
+			configFiles = append(configFiles, targetK8sConfig.TLSClientConfig.CertFile)
+		}
+	}
+
+	hubCfg, err := clientcmd.BuildConfigFromFlags("", opts.hubConfigPath)
+	if err != nil {
+		log.Error(err, "Could not load hub config")
+	} else {
```
yiraeChristineKim, line 546: you can remove `else`

yiraeChristineKim:
Overal good!

main.go
zyjjay: I use the `hubCfg` in the else block so I'm making sure I'm not referencing a nil pointer there.

main.go
```diff
@@ -529,9 +530,36 @@ func main() {
...
+		configFiles = append(configFiles, opts.hubConfigPath)
+
+		if hubCfg.TLSClientConfig.CertFile != "" {
+			configFiles = append(configFiles, hubCfg.TLSClientConfig.CertFile)
+		}
+	}
+
+	configChecker, err := addonutils.NewConfigChecker("config-policy-controller", configFiles...)
```
mprahl, line 554: Could you please move this to below the `if opts.enableLease {` if statement so that we can reuse `hubCfg` that gets loaded when starting the lease?

main.go
```diff
@@ -575,6 +573,32 @@ func main() {
...
 
+	if hubCfg != nil {
+		configFiles = append(configFiles, opts.hubConfigPath)
+
+		if hubCfg.TLSClientConfig.CertFile != "" {
+			configFiles = append(configFiles, hubCfg.TLSClientConfig.CertFile)
+		}
+	}
+
+	configChecker, err := addonutils.NewConfigChecker("config-policy-controller", configFiles...)
```
mprahl, line 584: To be on the safe side, if `len(configFiles) == 0`, could you please fall back to using `healthz.Ping`?

main.go
```diff
@@ -575,6 +573,40 @@ func main() {
...
+	if err != nil {
+		log.Error(err, "unable to setup a configChecker")
+		os.Exit(1)
+	}
+
+	var healthzCheck healthz.Checker
+
+	if len(configFiles) == 0 {
+		healthzCheck = healthz.Ping
+	} else {
```
mprahl, line 594: Could you please put instantiating the `configChecker` in the `else` portion?

mprahl (approved):
Looks great!

PR #286, opened 2024-08-09
Add fake operator for testing Operator Policy

JeffeyL:
Not having control over operators used in testing makes tests inconsistent. This commit adds a fake operator for testing that is locally built, and uses a locally hosted registry in KinD. It lives in its own catalog.

Ref: https://issues.redhat.com/browse/ACM-11160

**Details:**

Majority of the code goes under the `fake-operator-test` folder, which I'll be explaining by subfolder below. I'll then outline things that have yet to be completed.

**`fake-catalog`**
This folder contains everything required to build a catalog, originally scaffolded using the `opm` tool. I have not modified anything after initial scaffolding.

**`fake-operator`**
This folder contains everything that goes into building the fake operator bundle. This was originally scaffolded using `operator-sdk` following [these](https://sdk.operatorframework.io/docs/building-operators/golang/) instructions, but I removed the majority of it, and changed the dockerfile to specify a container that sleeps indefinitely (using `tail -f /dev/null` as an entrypoint). Another important point is that the make target `make bundle` uses the `config` subfolder to generate the bundle. I modified the `rbac` folder where I commented out `service_account.yaml` in the kustomization file, since it was giving me weird serviceaccount errors on the deployments. I also modified `manager/manager.yaml` to not contain a command under the `containers ` part of the spec, since on deployment this would give an error where the command wouldn't run. Documentation [here](https://docs.openshift.com/container-platform/4.9/rest_api/operatorhub_apis/clusterserviceversion-operators-coreos-com-v1alpha1.html) (you might need to ctrl-f 'command') seems to claim that the dockerfile `ENTRYPOINT` is used instead.

**`scripts`**
This folder contains two shell files `local-registry.sh` and `create-catalog.sh`. `local-registry.sh` creates a locally hosted registry following [these](https://kind.sigs.k8s.io/docs/user/local-registry/) instructions hosted at `localhost:5001`. `create-catalog.sh` does everything else required to make the catalogsource available, from building the operator bundle, building the catalog bundle, and applying it. It also applies a serviceaccount that seems to be needed to install deployments properly (but doesn't seem to be applied automatically). I wasn't too sure about this, so it should probably be examined a bit more thoroughly. Also, the deployment seems to fail anyways.

**Files:**

**`fake-operator-template.yaml`**
This file is required for building the catalogsource.

**`kind-registry-config.yaml`**
Part of hosting the kind registry

**`kind-registry-configmap.yaml`**
Also part of hosting the kind registry

**TODOs:**
1. If you apply fake-operator-test/samples/sample-oppol.yaml, this will work (almost). The operator deployment pod fails due to reasons that I couldn't find out in the end. There doesn't seem to be an error message when checking events, and the pod just fails by timing out eventually.
2. After this, you could probably start using the fake operator in the tests normally. This might involve tweaking some Makefile stuff, as currently, the catalog is only setup on `make kind-bootstrap-cluster-dev`


yiraeChristineKim:
resolve by https://github.com/stolostron/grc-mock-operators

PR #287, opened 2024-08-12
Split kind.yaml workflow into parallel jobs

JustinKuli:
The hosted mode tests and the "regular" tests were running in series, and each took about 20 minutes. Now they run in parallel in separate jobs, with a new job to merge their coverage data. Some other "quick" checks (like formatting, linting) were also moved to a separate job, which allows the tests to run even if there is a linting issue.

Refs:
 - https://issues.redhat.com/browse/ACM-13032

### DIFF
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -15,7 +15,71 @@ defaults:
     shell: bash
 
 jobs:
+  preflight-tests:
+    name: Preflight Tests
+    runs-on: ubuntu-latest
+    env:
+      KIND_VERSION: latest
+      REGISTRY: localhost:5000
+    steps:
+      - name: Checkout Config Policy Controller
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Format, Lint, and Generate
+        run: |
+          go mod verify
+          make fmt
+          git diff --exit-code
+          make lint
+          make generate
+          make generate-operator-yaml
+          git diff --exit-code
+
+      - name: Unit Tests
+        run: |
+          make test-coverage
+
+      - name: Upload Unit Test Coverage
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage_unit
+          path: coverage_unit.out
+
+      - name: Bootstrap K8s KinD Cluster and Deploy Controller
+        run: |
+          make kind-bootstrap-cluster-dev
+          KUBECONFIG=${PWD}/kubeconfig_managed make kind-ensure-sa
+          make build-images
+          KUBECONFIG=${PWD}/kubeconfig_managed_e2e make kind-deploy-controller-dev
+
+      - name: E2E tests that require the controller running in a cluster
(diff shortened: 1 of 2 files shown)


.github/workflows/kind.yml
```diff
@@ -15,20 +15,13 @@ defaults:
     shell: bash
 
 jobs:
-  kind-tests:
+  quick-checks:
+    name: Quick Checks
     runs-on: ubuntu-latest
     env:
+      KIND_VERSION: latest
```
dhaiducek, line 22: I don't _think_ this is necessary since it's the default--is it just here for transparency?

.github/workflows/kind.yml
```diff
@@ -15,20 +15,13 @@ defaults:
...
-      fail-fast: false
-      matrix:
-        # Run tests on minimum and newest supported OCP Kubernetes
-        # The "minimum" tag is set in the Makefile
-        # KinD tags: https://hub.docker.com/r/kindest/node/tags
-        kind:
-          - "minimum"
-          - "latest"
-    name: KinD tests
+    # no strategy.matrix; for simplicity this only uses a 'latest' KinD cluster
```
dhaiducek, line 24: (Optional) This feels more like a comment for the PR rather than a comment that lives in the code?

.github/workflows/kind.yml
```diff
@@ -15,20 +15,13 @@ defaults:
     shell: bash
 
 jobs:
-  kind-tests:
+  quick-checks:
+    name: Quick Checks
```
dhaiducek, line 19: I'm not sure how I feel about "Quick Checks" since it seems too casual, though I don't have great alternatives since it is kind of a catch-all group.

Brainstorming (which seem about equivalent now that I'm trying to come up with some 😆 ):
- Lint, Unit, Deploy Tests
- Preflight Tests

dhaiducek (approved):
/hold if you want to address any of the comments

LGTM! This sure has come a long way from when it ran serially! I have a couple minor comments, but I think they're more musings than anything that should block the PR.

Don't forget to update the Prow branch protections also. 🙂

.github/workflows/kind.yml
JustinKuli: Yep, for transparency, especially if the "no strategy.matrix" comment is removed.

.github/workflows/kind.yml
JustinKuli: Yeah, it was more helpful for me when writing these, to make sure they all had the same fields in the same order just to help readability. I'll remove it, and it will live on in this PR thread :laughing:

.github/workflows/kind.yml
JustinKuli: I like the sound of Preflight Tests (even though technically it's not "pre" the other tests). Since it is basically a catch-all, I don't want it to be too specifically named, and have to be renamed to stay accurate.

dhaiducek (approved):
/unhold

JustinKuli:
The oddly failing test gives a chance to confirm if re-running will work with this artifacts upload/download stuff, so I guess that's nice :laughing:

PR #294, opened 2024-09-16
Add ocm-polices namespace

yiraeChristineKim:
Deploy policies to the same namespace on every managed cluster rather than dynamically determine the managed cluster namespace name for each cluster.(example: argoCD)
Ref: https://issues.redhat.com/browse/ACM-13609

### DIFF
--- a/test/e2e/case1_pod_handling_test.go
+++ b/test/e2e/case1_pod_handling_test.go
@@ -12,6 +12,8 @@ import (
 	"open-cluster-management.io/config-policy-controller/test/utils"
 )
 
+const ocmNs = "open-cluster-management-policies"
+
 var _ = Describe("Test pod obj template handling", Ordered, func() {
 	Describe("Test pod obj handling on managed cluster in ns:"+testNamespace, Ordered, func() {
 		const (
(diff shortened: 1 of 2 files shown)


JustinKuli (dismissed):
/hold

LGTM! Just holding in case you want to add the coverage to the overall calculation, but it's not necessary.

.github/workflows/kind.yml
```diff
@@ -130,6 +130,45 @@ jobs:
...
+        env:
+          KIND_VERSION: latest
+        run: |
+          make kind-bootstrap-cluster-dev
+          KUBECONFIG=${PWD}/kubeconfig_managed make kind-ensure-sa
+
+      - name: OCM-policy namespace Tests
+        run: |
+          export GOPATH=$(go env GOPATH)
+          KUBECONFIG=${PWD}/kubeconfig_managed make e2e-ocm-ns-test-coverage
```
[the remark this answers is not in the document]
yiraeChristineKim: great idea

Makefile
```diff
@@ -241,15 +241,21 @@ install-resources:
...
-	$(GINKGO) -v --procs=$(E2E_PROCS) $(E2E_TEST_ARGS) test/e2e -- -is_hosted=$(IS_HOSTED)
+	$(GINKGO) -v --procs=$(E2E_PROCS) $(E2E_TEST_ARGS) test/e2e -- -is_hosted=$(IS_HOSTED) -test_namespace=$(TEST_NAMESAPCE)
 
 .PHONY: e2e-test-coverage
 e2e-test-coverage: E2E_TEST_ARGS = --json-report=report_e2e.json --label-filter='!hosted-mode && !running-in-cluster' --output-dir=.
 e2e-test-coverage: e2e-run-instrumented e2e-test e2e-stop-instrumented
 
+.PHONY: e2e-ocm-ns-test-coverage
+e2e-ocm-ns-test-coverage: E2E_TEST_ARGS = --json-report=report_e2e.json --label-filter='ocm-ns' --output-dir=.
+e2e-ocm-ns-test-coverage: TEST_NAMESAPCE = open-cluster-management-policies
```
mprahl, line 256:
```suggestion
e2e-ocm-ns-test-coverage: TEST_NAMESPACE = open-cluster-management-policies
```

Makefile
```diff
@@ -241,15 +241,21 @@ install-resources:
...
 e2e-test: e2e-dependencies
-	$(GINKGO) -v --procs=$(E2E_PROCS) $(E2E_TEST_ARGS) test/e2e -- -is_hosted=$(IS_HOSTED)
+	$(GINKGO) -v --procs=$(E2E_PROCS) $(E2E_TEST_ARGS) test/e2e -- -is_hosted=$(IS_HOSTED) -test_namespace=$(TEST_NAMESAPCE)
 
 .PHONY: e2e-test-coverage
 e2e-test-coverage: E2E_TEST_ARGS = --json-report=report_e2e.json --label-filter='!hosted-mode && !running-in-cluster' --output-dir=.
 e2e-test-coverage: e2e-run-instrumented e2e-test e2e-stop-instrumented
 
+.PHONY: e2e-ocm-ns-test-coverage
+e2e-ocm-ns-test-coverage: E2E_TEST_ARGS = --json-report=report_e2e.json --label-filter='ocm-ns' --output-dir=.
```
mprahl, line 255: I don't think you need to separate the tests. Simplify having `TEST_NAMESPACE=managed` will cause it to watch the `managed` and `open-cluster-management-policies` namespaces which is what users will experience when it's installed as part of OCM and ACM.

The user could explicitly set the watch namespace to `open-cluster-management-policies` when the config-policy-controller is manually deployed, but we don't need to explicitly test that since that's not a different code path.

test/e2e/case1_pod_handling_test.go
```diff
@@ -293,4 +295,66 @@ var _ = Describe("Test pod obj template handling", Ordered, func() {
 			utils.KubectlDelete("event", "--field-selector=involvedObject.name="+configPolicyName, "-n", "managed")
 		})
 	})
+
+	Describe("Test pod obj in "+ocmNs+" namespace", Ordered, func() {
```
mprahl, line 299: Could you modify the existing testcase instead since if it runs at the same time as this test, the pod it creates will conflict.

yiraeChristineKim: This is Ordered so it won't conflict

yiraeChristineKim:
```
Describe("Test pod obj template handling", Ordered, func() {
```

test/e2e/case1_pod_handling_test.go
```diff
@@ -293,4 +295,66 @@ var _ = Describe("Test pod obj template handling", Ordered, func() {
...
+				podName, "default", true, defaultTimeoutSeconds)
+			Expect(pod).NotTo(BeNil())
+		})
+		AfterAll(func() {
+			utils.KubectlDelete("ns", ocmNs)
+			policies := []string{
+				configPolicyNameInform,
+				configPolicyNameEnforce,
+			}
+			deleteConfigPolicies(policies)
```
mprahl, line 352: Is this still needed since the namespace is deleted?

yiraeChristineKim: Oh no

mprahl (approved):

PR #298, opened 2024-09-23
Add a DryRun CLI

JustinKuli:
See each commit, or this story: https://issues.redhat.com/browse/ACM-14161


pkg/mappings/mappings.go
```diff
@@ -0,0 +1,173 @@
...
+
+	for _, endpoint := range []string{"/api", "/apis"} {
+		path := kubeConfig.Host + kubeConfig.APIPath + endpoint
+
+		req, err := http.NewRequestWithContext(cmd.Context(), http.MethodGet, path, bytes.NewReader([]byte{}))
+		if err != nil {
+			return err
+		}
+
+		req.Header.Set("Accept", "application/json;g=apidiscovery.k8s.io;v=v2beta1;as=APIGroupDiscoveryList")
```
mprahl, line 108: Why v2beta1 instead of v1?

pkg/mappings/mappings.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	cli, err := rest.HTTPClientFor(kubeConfig)
+	if err != nil {
+		return err
+	}
+
+	apiMappings := []APIMapping{}
+
+	for _, endpoint := range []string{"/api", "/apis"} {
```
mprahl, line 100: Is there a reason to do this instead of leveraging `GroupsAndMaybeResources` on the discovery client?

pkg/mappings/mappings.go
```diff
@@ -0,0 +1,173 @@
...
+func GenerateMappings(cmd *cobra.Command, _ []string) error {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, &clientcmd.ConfigOverrides{})
+
+	kubeConfig, err := clientConfig.ClientConfig()
+	if err != nil {
+		return err
+	}
+
+	cli, err := rest.HTTPClientFor(kubeConfig)
```
mprahl, line 93: Could you check the status code is what you expect? An error isn't returned for a status code indicating an error.

cmd/dryrun/dryrun.go
```diff
@@ -0,0 +1,537 @@
...
+		if !cfgPolFound {
+			return nil, fmt.Errorf("invalid input Policy: it must contain a ConfigurationPolicy")
+		}
+	default:
+		return nil, fmt.Errorf("unsupported input kind: %v, must be 'Policy' or 'ConfigurationPolicy'",
+			unstruct.GetKind())
+	}
+
+	cfgpol.Spec.RemediationAction = policyv1.Inform
+	cfgpol.Spec.EvaluationInterval = policyv1.EvaluationInterval{
```
mprahl, line 222: So this means the custom messages can't have access to the full object right? I suppose this may be necessary to not be "watch".

cmd/dryrun/dryrun.go
```diff
@@ -0,0 +1,537 @@
...
+	}}
+
+	return &cfgpol, nil
+}
+
+// readInputResources takes stdin and any paths given as "positional" arguments,
+// and decodes them from YAML into k8s resources. Directories can be passed in
+// arguments: in this case all files in that directory will be read and decoded
+// (but no "deeper" directories are considered).
+func (d *DryRunner) readInputResources(cmd *cobra.Command, args []string) (
```
mprahl, line 238: Not sure if this is useful, but this is how I pull up all nested YAML files from the current directory (could be modified for your case):

```go
func getYAMLFiles() ([]string, error) {
	workingDir, err := os.Getwd()
	if err != nil {
		return nil, err
	}

	yamlFiles := []string{}

	err = filepath.WalkDir(workingDir, func(path string, d fs.DirEntry, err error) error {
		if err != nil {
			return nil
		}

		if d.IsDir() {
			return nil
		}

		fileExt := filepath.Ext(d.Name())
		if fileExt == ".yaml" || fileExt == ".yml" {
			yamlFiles = append(yamlFiles, path)
		}

		return nil
	})
	if err != nil {
		return nil, err
	}

	return yamlFiles, nil
}
```

pkg/mappings/mappings.go
```diff
@@ -0,0 +1,173 @@
...
+
+//go:embed default-mappings.json
+var defaultMappingsJSON []byte
+
+// DefaultMapper returns a RESTMapper initialized on default mappings that were
+// embedded into this package. It should include built-in Kubernetes types.
+func DefaultMapper() (meta.RESTMapper, error) {
+	mappings := []APIMapping{}
+
+	if err := json.Unmarshal(defaultMappingsJSON, &mappings); err != nil {
```
mprahl, line 72: Does it make sense to use a YAML parser so that both JSON and YAML are supported? The Kubernets manifests are all treated as YAML, so it'd be nice to be consistent here.

cmd/dryrun/dryrun.go
```diff
@@ -0,0 +1,537 @@
...
+) error {
+	events := corev1.EventList{}
+
+	if err := rec.List(ctx, &events, client.InNamespace(ns)); err != nil {
+		return err
+	}
+
+	messages := []string{}
+
+	for _, ev := range events.Items {
```
mprahl, line 493: This is very clever!

cmd/dryrun/dryrun.go
```diff
@@ -0,0 +1,537 @@
...
+		}
+	}
+
+	if d.messagesPath != "" {
+		f, err := os.Create(d.messagesPath)
+		if err != nil {
+			return err
+		}
+
+		for _, msg := range messages {
```
mprahl, line 505: Is this already sorted by timestamp?

mprahl:
@JustinKuli this is how I got the template lookups to work:

```diff
diff --git a/cmd/dryrun/dryrun.go b/cmd/dryrun/dryrun.go
index e69978e..20e3723 100644
--- a/cmd/dryrun/dryrun.go
+++ b/cmd/dryrun/dryrun.go
@@ -447,14 +447,15 @@ func (d *DryRunner) setupReconciler(
 			return nil, err
 		}
 
-		rec.RestMapper = mappings.MapperFrom(apiMappings)
+		rec.RestMapper, clientset.Resources = mappings.MapperFrom(apiMappings)
 	} else {
-		mapper, err := mappings.DefaultMapper()
+		mapper, resourceList, err := mappings.DefaultMapper()
 		if err != nil {
 			return nil, err
 		}
 
 		rec.RestMapper = mapper
+		clientset.Resources = resourceList
 	}
 
 	// wait for dynamic watcher to have started
diff --git a/pkg/mappings/mappings.go b/pkg/mappings/mappings.go
index e4ea0f6..088b6a2 100644
--- a/pkg/mappings/mappings.go
+++ b/pkg/mappings/mappings.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -32,9 +33,11 @@ type APIMapping struct {
 // MapperFrom takes a list of APIMappings and uses them to populate a
 // DefaultRESTMapper which can be used locally without connecting to a Discovery
 // client.
-func MapperFrom(mappings []APIMapping) meta.RESTMapper {
+func MapperFrom(mappings []APIMapping) (meta.RESTMapper, []*metav1.APIResourceList) {
 	mapper := meta.NewDefaultRESTMapper(scheme.Scheme.PreferredVersionAllGroups())
 
+	resourceListMapping := map[schema.GroupVersion][]metav1.APIResource{}
+
 	for _, mapping := range mappings {
 		gv := schema.GroupVersion{
 			Group:   mapping.Group,
@@ -56,9 +59,25 @@ func MapperFrom(mappings []APIMapping) meta.RESTMapper {
 				namespaced,
 			)
 		}
+
+		resourceListMapping[gv] = append(resourceListMapping[gv], metav1.APIResource{
+			Kind:         mapping.Kind,
+			Name:         mapping.Plural,
+			SingularName: mapping.Singular,
+			Namespaced:   mapping.Scope == "Namespaced",
+		})
+	}
+
+	resourceLists := []*metav1.APIResourceList{}
+
+	for gv := range resourceListMapping {
+		resourceLists = append(resourceLists, &metav1.APIResourceList{
+			GroupVersion: gv.String(),
+			APIResources: resourceListMapping[gv],
+		})
 	}
 
-	return mapper
+	return mapper, resourceLists
 }
 
 //go:embed default-mappings.json
@@ -66,14 +85,16 @@ var defaultMappingsJSON []byte
 
 // DefaultMapper returns a RESTMapper initialized on default mappings that were
 // embedded into this package. It should include built-in Kubernetes types.
-func DefaultMapper() (meta.RESTMapper, error) {
+func DefaultMapper() (meta.RESTMapper, []*metav1.APIResourceList, error) {
 	mappings := []APIMapping{}
 
 	if err := json.Unmarshal(defaultMappingsJSON, &mappings); err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	return MapperFrom(mappings), nil
+	mapper, resourceList := MapperFrom(mappings)
+
+	return mapper, resourceList, nil
 }
 
 // GenerateMappings connects to a Kubernetes cluster and discovers the available

```

cmd/dryrun/dryrun.go
JustinKuli: Good point, I forgot about that when I was testing the custom messages and thought it was an issue with the discovery client. I think now that I added this to avoid the discovery client for a different reason, but maybe we can work around the issue in a different way.

cmd/dryrun/dryrun.go
JustinKuli: I believe the items in the event list are sorted lexically, but since a timestamp is in the name, effectively it is sorted by time.

pkg/mappings/mappings.go
JustinKuli: Can do.

pkg/mappings/mappings.go
JustinKuli: Down in the `Do`? Yeah I can check that, good point.

pkg/mappings/mappings.go
JustinKuli: Only because I wasn't familiar with that method, and I was somewhat clumsily recreating what I saw `kubectl` do :laughing:

pkg/mappings/mappings.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return err
+	}
+
+	apiMappings := []APIMapping{}
+
+	for gv, resList := range resources {
+		for _, res := range resList.APIResources {
+			if strings.Contains(res.Name, "/") {
+				continue // skip subresources
```
JustinKuli, line 122: @mprahl FYI, `GroupsAndMaybeResources` includes the subresources, which I was ignoring when parsing /api and /apis "by hand"... I didn't see a better way to determine if something was a subresource, but if I included them in the mappings / resources used by `dryrun`, tests would fail and it looked like it was incorrectly checking things in subresources (eg a noncompliant message indicated that a "pod/status did not match")

cmd/dryrun/dryrun.go
```diff
@@ -0,0 +1,535 @@
...
+	stdinInfo, err := os.Stdin.Stat()
+	if err != nil {
+		return nil, fmt.Errorf("could not read stdin: %w", err)
+	}
+
+	rawInputs := []*unstructured.Unstructured{}
+
+	filepaths := []string{}
+
+	if stdinInfo.Mode()&os.ModeCharDevice == 0 {
```
mprahl, line 250: Not entirely sure if this is the correct fix, but when running it through the debugger where it has no stdin, it would just hang forever for the YAML decoder to get something from stdin.
```suggestion
	if stdinInfo.Mode()&os.ModeCharDevice == 0 && stdinInfo.Size() != 0 {
```

cmd/dryrun/dryrun.go
mprahl: After investigating, it seems the fake client doesn't return a resource version from a list query, so the watch is unsupported.

mprahl:
@JustinKuli could we please remove the RestMapper in the controller? I think this works pretty well with this diff:

```diff
diff --git a/cmd/dryrun/dryrun.go b/cmd/dryrun/dryrun.go
index 662e716..18c72e5 100644
--- a/cmd/dryrun/dryrun.go
+++ b/cmd/dryrun/dryrun.go
@@ -22,9 +22,9 @@ import (
 	yaml "gopkg.in/yaml.v3"
 	corev1 "k8s.io/api/core/v1"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
 	dynfake "k8s.io/client-go/dynamic/fake"
@@ -71,17 +71,40 @@ func (d *DryRunner) dryRun(cmd *cobra.Command, args []string) error {
 	for _, obj := range inputObjects {
 		gvk := obj.GroupVersionKind()
 
-		mapping, err := rec.RestMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+		resourceList, err := rec.TargetK8sClient.Discovery().ServerResourcesForGroupVersion(gvk.GroupVersion().String())
 		if err != nil {
 			return err
 		}
 
+		var apiResource *metav1.APIResource
+
+		for _, resource := range resourceList.APIResources {
+			if resource.Kind == gvk.Kind {
+				resource := resource
+
+				apiResource = &resource
+
+				break
+			}
+		}
+
+		if apiResource == nil {
+			return fmt.Errorf("there is no mapping for %s; consider running the generate subcommand on a cluster "+
+				"with this resource", gvk.String())
+		}
+
+		gvr := schema.GroupVersionResource{
+			Group:    gvk.Group,
+			Version:  gvk.Version,
+			Resource: apiResource.Name,
+		}
+
 		var resInt dynamic.ResourceInterface
 
-		if mapping.Scope.Name() == meta.RESTScopeNameNamespace {
-			resInt = rec.TargetK8sDynamicClient.Resource(mapping.Resource).Namespace(obj.GetNamespace())
+		if apiResource.Namespaced {
+			resInt = rec.TargetK8sDynamicClient.Resource(gvr).Namespace(obj.GetNamespace())
 		} else {
-			resInt = rec.TargetK8sDynamicClient.Resource(mapping.Resource)
+			resInt = rec.TargetK8sDynamicClient.Resource(gvr)
 		}
 
 		if _, err := resInt.Create(ctx, obj, metav1.CreateOptions{}); err != nil {
@@ -247,7 +270,7 @@ func (d *DryRunner) readInputResources(cmd *cobra.Command, args []string) (
 
 	filepaths := []string{}
 
-	if stdinInfo.Mode()&os.ModeCharDevice == 0 {
+	if stdinInfo.Mode()&os.ModeCharDevice == 0 && stdinInfo.Size() != 0 {
 		filepaths = append(filepaths, "-")
 	}
 
@@ -447,9 +470,9 @@ func (d *DryRunner) setupReconciler(
 			return nil, err
 		}
 
-		rec.RestMapper, clientset.Resources = mappings.MapperFrom(apiMappings)
+		clientset.Resources = mappings.ResourceListFromMapping(apiMappings)
 	} else {
-		rec.RestMapper, clientset.Resources, err = mappings.DefaultMapper()
+		clientset.Resources, err = mappings.DefaultResourceList()
 		if err != nil {
 			return nil, err
 		}
diff --git a/controllers/configurationpolicy_controller.go b/controllers/configurationpolicy_controller.go
index 135fccf..007aafd 100644
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -27,7 +27,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	extensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	apimachinerymeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -203,7 +202,6 @@ type ConfigurationPolicyReconciler struct {
 	// This is a workaround to account for race conditions where the status is updated but the controller-runtime cache
 	// has not updated yet.
 	lastEvaluatedCache sync.Map
-	RestMapper         apimachinerymeta.RESTMapper
 }
 
 //+kubebuilder:rbac:groups=*,resources=*,verbs=*
@@ -1901,16 +1899,6 @@ func (r *ConfigurationPolicyReconciler) getMapping(
 		return depclient.ScopedGVR{}, err
 	}
 
-	if r.RestMapper != nil {
-		mapping, err := r.RestMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
-		if err == nil {
-			return depclient.ScopedGVR{
-				GroupVersionResource: mapping.Resource,
-				Namespaced:           mapping.Scope.Name() == apimachinerymeta.RESTScopeNameNamespace,
-			}, nil
-		}
-	}
-
 	scopedGVR, err := r.DynamicWatcher.GVKToGVR(gvk)
 	if err != nil {
 		if !errors.Is(err, depclient.ErrNoVersionedResource) {
diff --git a/pkg/mappings/mappings.go b/pkg/mappings/mappings.go
index 38fe86c..88998ba 100644
--- a/pkg/mappings/mappings.go
+++ b/pkg/mappings/mappings.go
@@ -13,7 +13,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/kubectl/pkg/scheme"
 	"sigs.k8s.io/yaml"
 )
 
@@ -32,12 +31,9 @@ func (a APIMapping) String() string {
 	return a.Group + "/" + a.Version + " " + a.Plural
 }
 
-// MapperFrom takes a list of APIMappings and uses them to populate a
-// DefaultRESTMapper which can be used locally without connecting to a Discovery
-// client.
-func MapperFrom(mappings []APIMapping) (meta.RESTMapper, []*metav1.APIResourceList) {
-	mapper := meta.NewDefaultRESTMapper(scheme.Scheme.PreferredVersionAllGroups())
-
+// ResourceListFromMapping takes a list of APIMappings and uses them to populate an
+// APIResourceList which can be used to populate the fake discovery client.
+func ResourceListFromMapping(mappings []APIMapping) []*metav1.APIResourceList {
 	resourceListMapping := map[schema.GroupVersion][]metav1.APIResource{}
 
 	for _, mapping := range mappings {
@@ -46,13 +42,6 @@ func MapperFrom(mappings []APIMapping) (meta.RESTMapper, []*metav1.APIResourceLi
 			Version: mapping.Version,
 		}
 
-		mapper.AddSpecific(
-			gv.WithKind(mapping.Kind),
-			gv.WithResource(mapping.Plural),
-			gv.WithResource(mapping.Singular),
-			mapping.Scope,
-		)
-
 		resourceListMapping[gv] = append(resourceListMapping[gv], metav1.APIResource{
 			Kind:         mapping.Kind,
 			Name:         mapping.Plural,
@@ -70,24 +59,24 @@ func MapperFrom(mappings []APIMapping) (meta.RESTMapper, []*metav1.APIResourceLi
 		})
 	}
 
-	return mapper, resourceLists
+	return resourceLists
 }
 
 //go:embed default-mappings.yaml
 var defaultMappings []byte
 
-// DefaultMapper returns a RESTMapper initialized on default mappings that were
-// embedded into this package. It should include built-in Kubernetes types.
-func DefaultMapper() (meta.RESTMapper, []*metav1.APIResourceList, error) {
+// ResourceListFromMapping uses the default list of APIMappings and uses them to populate an
+// APIResourceList which can be used to populate the fake discovery client.
+func DefaultResourceList() ([]*metav1.APIResourceList, error) {
 	mappings := []APIMapping{}
 
 	if err := yaml.Unmarshal(defaultMappings, &mappings); err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 
-	mapper, resList := MapperFrom(mappings)
+	resList := ResourceListFromMapping(mappings)
 
-	return mapper, resList, nil
+	return resList, nil
 }
 
 // GenerateMappings connects to a Kubernetes cluster and discovers the available
```

mprahl:
@JustinKuli could we please remove the RestMapper in the controller? I think this works pretty well:

```diff
diff --git a/cmd/dryrun/dryrun.go b/cmd/dryrun/dryrun.go
index 662e716..f93b601 100644
--- a/cmd/dryrun/dryrun.go
+++ b/cmd/dryrun/dryrun.go
@@ -22,7 +22,6 @@ import (
 	yaml "gopkg.in/yaml.v3"
 	corev1 "k8s.io/api/core/v1"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
@@ -71,17 +70,17 @@ func (d *DryRunner) dryRun(cmd *cobra.Command, args []string) error {
 	for _, obj := range inputObjects {
 		gvk := obj.GroupVersionKind()
 
-		mapping, err := rec.RestMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+		scopedGVR, err := rec.DynamicWatcher.GVKToGVR(gvk)
 		if err != nil {
 			return err
 		}
 
 		var resInt dynamic.ResourceInterface
 
-		if mapping.Scope.Name() == meta.RESTScopeNameNamespace {
-			resInt = rec.TargetK8sDynamicClient.Resource(mapping.Resource).Namespace(obj.GetNamespace())
+		if scopedGVR.Namespaced {
+			resInt = rec.TargetK8sDynamicClient.Resource(scopedGVR.GroupVersionResource).Namespace(obj.GetNamespace())
 		} else {
-			resInt = rec.TargetK8sDynamicClient.Resource(mapping.Resource)
+			resInt = rec.TargetK8sDynamicClient.Resource(scopedGVR.GroupVersionResource)
 		}
 
 		if _, err := resInt.Create(ctx, obj, metav1.CreateOptions{}); err != nil {
@@ -247,7 +246,7 @@ func (d *DryRunner) readInputResources(cmd *cobra.Command, args []string) (
 
 	filepaths := []string{}
 
-	if stdinInfo.Mode()&os.ModeCharDevice == 0 {
+	if stdinInfo.Mode()&os.ModeCharDevice == 0 && stdinInfo.Size() != 0 {
 		filepaths = append(filepaths, "-")
 	}
 
@@ -447,9 +446,9 @@ func (d *DryRunner) setupReconciler(
 			return nil, err
 		}
 
-		rec.RestMapper, clientset.Resources = mappings.MapperFrom(apiMappings)
+		clientset.Resources = mappings.ResourceListFromMapping(apiMappings)
 	} else {
-		rec.RestMapper, clientset.Resources, err = mappings.DefaultMapper()
+		clientset.Resources, err = mappings.DefaultResourceList()
 		if err != nil {
 			return nil, err
 		}
diff --git a/controllers/configurationpolicy_controller.go b/controllers/configurationpolicy_controller.go
index 135fccf..007aafd 100644
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -27,7 +27,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	extensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	apimachinerymeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -203,7 +202,6 @@ type ConfigurationPolicyReconciler struct {
 	// This is a workaround to account for race conditions where the status is updated but the controller-runtime cache
 	// has not updated yet.
 	lastEvaluatedCache sync.Map
-	RestMapper         apimachinerymeta.RESTMapper
 }
 
 //+kubebuilder:rbac:groups=*,resources=*,verbs=*
@@ -1901,16 +1899,6 @@ func (r *ConfigurationPolicyReconciler) getMapping(
 		return depclient.ScopedGVR{}, err
 	}
 
-	if r.RestMapper != nil {
-		mapping, err := r.RestMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
-		if err == nil {
-			return depclient.ScopedGVR{
-				GroupVersionResource: mapping.Resource,
-				Namespaced:           mapping.Scope.Name() == apimachinerymeta.RESTScopeNameNamespace,
-			}, nil
-		}
-	}
-
 	scopedGVR, err := r.DynamicWatcher.GVKToGVR(gvk)
 	if err != nil {
 		if !errors.Is(err, depclient.ErrNoVersionedResource) {
diff --git a/pkg/mappings/mappings.go b/pkg/mappings/mappings.go
index 38fe86c..88998ba 100644
--- a/pkg/mappings/mappings.go
+++ b/pkg/mappings/mappings.go
@@ -13,7 +13,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/kubectl/pkg/scheme"
 	"sigs.k8s.io/yaml"
 )
 
@@ -32,12 +31,9 @@ func (a APIMapping) String() string {
 	return a.Group + "/" + a.Version + " " + a.Plural
 }
 
-// MapperFrom takes a list of APIMappings and uses them to populate a
-// DefaultRESTMapper which can be used locally without connecting to a Discovery
-// client.
-func MapperFrom(mappings []APIMapping) (meta.RESTMapper, []*metav1.APIResourceList) {
-	mapper := meta.NewDefaultRESTMapper(scheme.Scheme.PreferredVersionAllGroups())
-
+// ResourceListFromMapping takes a list of APIMappings and uses them to populate an
+// APIResourceList which can be used to populate the fake discovery client.
+func ResourceListFromMapping(mappings []APIMapping) []*metav1.APIResourceList {
 	resourceListMapping := map[schema.GroupVersion][]metav1.APIResource{}
 
 	for _, mapping := range mappings {
@@ -46,13 +42,6 @@ func MapperFrom(mappings []APIMapping) (meta.RESTMapper, []*metav1.APIResourceLi
 			Version: mapping.Version,
 		}
 
-		mapper.AddSpecific(
-			gv.WithKind(mapping.Kind),
-			gv.WithResource(mapping.Plural),
-			gv.WithResource(mapping.Singular),
-			mapping.Scope,
-		)
-
 		resourceListMapping[gv] = append(resourceListMapping[gv], metav1.APIResource{
 			Kind:         mapping.Kind,
 			Name:         mapping.Plural,
@@ -70,24 +59,24 @@ func MapperFrom(mappings []APIMapping) (meta.RESTMapper, []*metav1.APIResourceLi
 		})
 	}
 
-	return mapper, resourceLists
+	return resourceLists
 }
 
 //go:embed default-mappings.yaml
 var defaultMappings []byte
 
-// DefaultMapper returns a RESTMapper initialized on default mappings that were
-// embedded into this package. It should include built-in Kubernetes types.
-func DefaultMapper() (meta.RESTMapper, []*metav1.APIResourceList, error) {
+// ResourceListFromMapping uses the default list of APIMappings and uses them to populate an
+// APIResourceList which can be used to populate the fake discovery client.
+func DefaultResourceList() ([]*metav1.APIResourceList, error) {
 	mappings := []APIMapping{}
 
 	if err := yaml.Unmarshal(defaultMappings, &mappings); err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 
-	mapper, resList := MapperFrom(mappings)
+	resList := ResourceListFromMapping(mappings)
 
-	return mapper, resList, nil
+	return resList, nil
 }
 
 // GenerateMappings connects to a Kubernetes cluster and discovers the available
```

cmd/dryrun/dryrun.go
JustinKuli: We'll need to look at this more. With `&& stdinInfo.Size() != 0`, I found that the cli doesn't successfully read inputs from stdin anymore when you run it directly. The tests still pass, I think because I kind of hacked the inputs for the tests...

So based on the `test_multiple_resources` case:
```
❯ cat input_stdin.yaml | ./dryrun-original -p policy.yaml input_3.yaml input_4.yaml 
# Diffs:
v1 Pod default/nginx-pod-e2e:

v1 Pod default/nginx-pod-e2e-2:

v1 Pod default/nginx-pod-e2e-3:

v1 Pod default/nginx-pod-e2e-4:

# Compliance messages:
Compliant; notification - pods [nginx-pod-e2e] found as specified in namespace default; notification - pods [nginx-pod-e2e-2] found as specified in namespace default; notification - pods [nginx-pod-e2e-3] found as specified in namespace default; notification - pods [nginx-pod-e2e-4] found as specified in namespace default

❯ cat input_stdin.yaml | ./dryrun-sizene0 -p policy.yaml input_3.yaml input_4.yaml                 
# Diffs:
v1 Pod default/nginx-pod-e2e-3:

v1 Pod default/nginx-pod-e2e-4:

# Compliance messages:
Compliant; notification - pods [nginx-pod-e2e-3] found as specified in namespace default; notification - pods [nginx-pod-e2e-4] found as specified in namespace default
```

JustinKuli: Is there a debugger setting where you can set stdin?

pkg/mappings/mappings.go
JustinKuli: (Unnecessary since change to `GroupsAndMaybeResources`)

cmd/dryrun/dryrun.go
mprahl: @JustinKuli I don't know of one but this is the launch.json section that I had with this issue:

```
        {
            "name": "Launch dryrun",
            "type": "go",
            "request": "launch",
            "mode": "auto",
            "program": "${workspaceFolder}/cmd/main.go",
            "args": [
                "-p",
                "${workspaceFolder}/cmd/dryrun/testdata/test_basic_compliant/policy.yaml",
                "${workspaceFolder}/cmd/dryrun/testdata/test_basic_compliant/input_1.yaml"
            ]
        }
```

mprahl: We can leave that out for now though!

mprahl (approved):

PR #302, opened 2024-10-03
More dryrun improvements

JustinKuli:
The "big" change here is the path of the cli, so that it can be better installed via `go install`. It will also now print the usage doc if the user inputs are incorrect.


dhaiducek (approved):
Nice! Thanks for the update!

PR #303, opened 2024-10-09
Uninstall scenario improvements

JustinKuli:
See each commit for details.

The previous test for the uninstall scenario was incorrectly passing, I think because there were leftover re-reconciles on the policies it was using. But generally those policies would not be reconciled when the uninstall starts.

### DIFF
--- a/main.go
+++ b/main.go
@@ -663,7 +663,7 @@ func main() {
 func handleTriggerUninstall() {
 	triggerUninstallFlagSet := pflag.NewFlagSet("trigger-uninstall", pflag.ExitOnError)
 
-	var deploymentName, deploymentNamespace, policyNamespace string
+	var deploymentName, deploymentNamespace, policyNamespace, additionalNamespace string
 	var timeoutSeconds uint32
 
 	triggerUninstallFlagSet.StringVar(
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -678,6 +678,9 @@ func handleTriggerUninstall() {
 	triggerUninstallFlagSet.StringVar(
 		&policyNamespace, "policy-namespace", "", "The namespace of where ConfigurationPolicy objects are stored",
 	)
+	triggerUninstallFlagSet.StringVar(
+		&additionalNamespace, "additional-namespace", "", "An additional namespace for ConfigurationPolicy objects",
```
yiraeChristineKim, line 682: Should config-controller watch this namespace too?

pkg/triggeruninstall/triggeruninstall.go
```diff
@@ -77,35 +81,47 @@ func TriggerUninstall(
...
+		cleanedUp := true
 
-				continue
+		for _, ns := range policyNamespaces {
+			policyClient := dynamicClient.Resource(configPolicyGVR).Namespace(ns)
+
+			configPolicies, err := policyClient.List(ctx, metav1.ListOptions{})
+			if err != nil {
+				if k8serrors.IsServerTimeout(err) || k8serrors.IsTimeout(err) {
+					klog.Infof("Retrying listing the ConfigurationPolicy objects due to error: %s", err)
```
yiraeChristineKim, line 92: Is this retrying? Since it uses continue, it skips the current namespace in this iteration."

main.go
JustinKuli: This `trigger-uninstall` flag is meant to match when the main controller is watching "open-cluster-management-policies": https://github.com/open-cluster-management-io/config-policy-controller/blob/main/main.go#L259

JustinKuli: I'll need to add it in the addon-controller

pkg/triggeruninstall/triggeruninstall.go
JustinKuli: Ah, good catch, I need to make it `continue` the bigger loop I think

JustinKuli: Or maybe just mark `cleanedUp = false`... but now that I'm looking at it closer, I don't quite follow why we're specially handling these specific errors. @mprahl do you remember?

yiraeChristineKim (approved):

PR #306, opened 2024-10-18
Send operator policy event if status is reset

JustinKuli:
ConfigurationPolicies were already handling this, and the predicate used there was modified for OperatorPolicy so that it is watching for the same events.

The new test for this behavior was added to the existing CRD status reporting test, because it benefits from being run after the full operator has been installed, and adding a separate test that does that would be an inefficient use of test time.

Refs:
 - https://issues.redhat.com/browse/ACM-15049


dhaiducek (approved):
/hold Is there an issue tied to this that should be tagged here?

LGTM!

JustinKuli:
Created and linked https://issues.redhat.com/browse/ACM-15049, thanks for the reminder @dhaiducek

PR #308, opened 2024-10-21
Add logs to debug possible lock issue

JustinKuli:
In one case, the operator-policy-controller seemed to completely stop, without emitting an error, after appearing to finish reconciling a policy that should have been requeued. These log messages may assist in identifying where it got stuck.

Refs:
 - https://issues.redhat.com/browse/ACM-14617

### DIFF
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -205,6 +205,15 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	policy := &policyv1beta1.OperatorPolicy{}
 	watcher := opPolIdentifier(req.Namespace, req.Name)
 
+	opLog.Info("Reconciling OperatorPolicy")
+
+	defer func() {
+		// ACM-14617: This log is to help identify if a deferred function call is getting stuck on
+		// a lock. This message must be at INFO level because the issue may not be reproduced if the
+		// log level is adjusted *after* the problem occurs.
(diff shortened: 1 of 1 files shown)


controllers/operatorpolicy_controller.go
```diff
@@ -205,6 +205,15 @@ func (r *OperatorPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	policy := &policyv1beta1.OperatorPolicy{}
 	watcher := opPolIdentifier(req.Namespace, req.Name)
 
+	opLog.Info("Reconciling OperatorPolicy")
+
+	defer func() {
+		// ACM-14617: This log is to help identify if a deferred function call is getting stuck on
+		// a lock. This message must be at INFO level because the issue may not be reproduced if the
+		// log level is adjusted *after* the problem occurs.
+		opLog.Info("All deferred functions for the reconcile are complete")
```
mprahl, line 214: Instead of this, could it be the reconcile complete log message? If you define the variables outside of this function scope just above, the defer function should pick up the latest values of `finalErr` and etc.

yiraeChristineKim (approved):

controllers/operatorpolicy_controller.go
JustinKuli: I originally tried something like that, but I realized it would be useful to have a "bottom of reconcile" kind of log message, in addition to this "everything actually complete" message - this would help us spot if the deferred cleanup steps are what is hanging. So it was unnecessary to pre-define any variables.

PR #309, opened 2024-10-23
Set `UnknownCompliancy` to an empty string

dhaiducek:
The framework-addon controller looks for an empty string. Plus, this aligns with the enum.

Signed-off-by: Dale Haiducek <19750917+dhaiducek@users.noreply.github.com>


dhaiducek:
> I kind of wish that the OperatorPolicy would label things as "Inapplicable" instead of relying on the UI to know about that kind of rule...

Like...this? 😄  https://github.com/open-cluster-management-io/config-policy-controller/pull/309/commits/79049bf3b4451746b8f775a8eaeac8198bdd98d7

JustinKuli (approved):
/lgtm

PR #312, opened 2024-10-31
Adopt existing sub when packagemanifest not found

JustinKuli:
Previously, if the pacakagemanifest was missing, perhaps due to the operator no longer being available in the catalog, the operator policy would be marked as invalid, and not report on the potentially helpful status of the subscription. Now the existing subscription can be used for the default values, and a more helpful status can be reported.

Refs:
 - https://issues.redhat.com/browse/ACM-14617

### DIFF
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -571,7 +571,8 @@ func (r *OperatorPolicyReconciler) checkSubOverlap(
 }
 
 // applySubscriptionDefaults will set the subscription channel, source, and sourceNamespace when they are unset by
-// utilizing the PackageManifest API.
+// utilizing the PackageManifest API. If the PackageManifest can not be found, and a subscription already exists for the
+// operator, then information from the found subscription will be used.
 func (r *OperatorPolicyReconciler) applySubscriptionDefaults(
 	ctx context.Context, policy *policyv1beta1.OperatorPolicy, subscription *operatorv1alpha1.Subscription,
 ) error {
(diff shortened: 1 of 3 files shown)


controllers/operatorpolicy_controller.go
```diff
@@ -679,6 +684,55 @@ func (r *OperatorPolicyReconciler) applySubscriptionDefaults(
...
+	policy *policyv1beta1.OperatorPolicy, subscription *operatorv1alpha1.Subscription,
+) bool {
+	if subscription.Namespace == "" {
+		// check for an already-known subscription to "adopt"
+		subs := policy.Status.RelatedObjsOfKind("Subscription")
+
+		if len(subs) == 1 {
+			// Note: RelatedObjsOfKind returns a map - in this case we just want the one object
+			for _, sub := range subs {
+				subscription.Namespace = sub.Object.Metadata.Namespace
```
yiraeChristineKim, line 700: Should break? if assigned?

controllers/operatorpolicy_controller.go
```diff
@@ -594,10 +595,14 @@ func (r *OperatorPolicyReconciler) applySubscriptionDefaults(
 	)
 	if err != nil {
 		if k8serrors.IsNotFound(err) {
-			return fmt.Errorf(
-				"%wthe subscription defaults could not be determined because the PackageManifest was not found",
-				ErrPackageManifest,
-			)
+			if !r.usingExistingSubIfFound(policy, subscription) {
```
yiraeChristineKim, line 598: We don't have any status change for usingExistingSubIfFound

JustinKuli: It doesn't make a special status directly, it just allows the policy to be considered valid so that the status of the subscription can be updated (an invalid policy doesn't update the other conditions). The subscription status will have more information in some cases, for example explaining that the operator is not available in a catalog.

yiraeChristineKim (approved):

PR #314, opened 2024-11-11
Provide the ObjectNamespace template variable

mprahl:
A user can now leverage the ObjectNamespace template variable to have a ConfigurationPolicy inform or enforce on a named object for all namespaces in the namespace selector.

As part of this, the code had to be refactored to resolve templates per object template rather than resolve all templates up front. This has the side effect of allowing some object templates to still be evaluated while another object template failed template resolution.

This made it so that the templating metrics no longer apply since there can be multiple calls to resolve templates per policy in different parts of the code.

Relates:
https://issues.redhat.com/browse/ACM-15392

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"math"
 	"reflect"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -70,6 +71,7 @@ const (
 	reasonWantNotFoundDNE    = "Resource not found as expected"
 	reasonCleanupError       = "Error cleaning up child objects"
 	reasonFoundNotApplicable = "Resource found but will not be handled in mustnothave mode"
+	reasonTemplateError      = "Error processing template"
 )
 
 var (
@@ -84,6 +86,8 @@ var (
 	// commonSprigFuncMap includes only the sprig functions that are available in the
 	// stolostron/go-template-utils library.
 	commonSprigFuncMap template.FuncMap
+
+	templateHasObjectNamespaceRegex = regexp.MustCompile(`(\.ObjectNamespace)`)
 )
 
 func init() {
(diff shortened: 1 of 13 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -766,17 +909,37 @@ func (r *ConfigurationPolicyReconciler) handleObjectTemplates(plc *policyv1.Conf
...
 
-		parentStatusUpdateNeeded = updateNeeded
+		if plc.Spec.ObjectTemplatesRaw != "" {
+			err := r.resolveObjectTemplatesRaw(plc, tmplResolver, resolveOptions)
+			if err != nil {
+				return err
+			}
+
+			// Templates are already handled so disable any further processing.
+			disableTemplates = true
```
dhaiducek, line 932: Nice--I was wondering about the use of the namespace template variable with `object-templates-raw`, but I think this makes sense to just shut it off after processing it.

controllers/configurationpolicy_controller.go
```diff
@@ -1045,277 +1225,250 @@ func (r *ConfigurationPolicyReconciler) handleDeletion(plc *policyv1.Configurati
...
 
-// handleTemplatization sets the `plc.Spec.ObjectTemplates` after resolving any templatized values.
-func (r *ConfigurationPolicyReconciler) handleTemplatization(
-	plc *policyv1.ConfigurationPolicy, usingWatch bool,
-) (parentStatusUpdateNeeded bool, err error) {
-	log := log.WithValues("policy", plc.GetName())
-
-	if r.EnableMetrics {
-		startTime := time.Now().UTC()
+type minimumMetadata struct {
```
dhaiducek, line 1228: Oh, nifty. I don't think I had ever considered marshalling directly to something smaller than the full object. 🙂

dhaiducek (commented):
Nice work! I have a couple minor comments, but I think this looks good!

I also see pre-existing instances of `context.TODO()` that I'm thinking could be swapped out for the relevant context, but that's more tech debt unrelated to this PR.

controllers/configurationpolicy_controller.go
mprahl: I'll admit, I saw something similar to parse the GVK from an object in a Kubernetes library, but it didn't include the name and namespace.

controllers/metric.go
```diff
@@ -25,22 +25,6 @@ var (
 		},
 		[]string{"name"},
 	)
-	plcTempsProcessSecondsCounter = prometheus.NewCounterVec(
```
JustinKuli, line 28: IIRC, these metrics were originally for our own investigations in performance, so I think we're totally fine removing them :+1:

test/e2e/case13_templatization_test.go
```diff
@@ -483,4 +461,90 @@ var _ = Describe("Test templatization", Ordered, func() {
...
+				managedPlc := utils.GetWithTimeout(
+					clientManagedDynamic, gvrConfigPolicy, policyName, testNamespace, true, defaultTimeoutSeconds,
+				)
+
+				utils.CheckComplianceStatus(g, managedPlc, "Compliant")
+
+				relatedObjects, _, _ := unstructured.NestedSlice(managedPlc.Object, "status", "relatedObjects")
+				g.Expect(relatedObjects).To(HaveLen(2))
+
+				realtedObject1, ok := relatedObjects[0].(map[string]interface{})
```
JustinKuli, line 491:
```suggestion
				relatedObject1, ok := relatedObjects[0].(map[string]interface{})
```

test/e2e/case13_templatization_test.go
```diff
@@ -483,4 +461,90 @@ var _ = Describe("Test templatization", Ordered, func() {
...
+					gvrConfigPolicy,
+					invalidPolicyName,
+					testNamespace,
+					true,
+					defaultTimeoutSeconds,
+				)
+
+				utils.CheckComplianceStatus(g, managedPlc, "NonCompliant")
+				g.Expect(utils.GetStatusMessage(managedPlc)).To(Equal(
+					"The object definition's namespace must match the selected namespace after template resolution",
```
JustinKuli, line 539: I think this is a nice approach - I haven't looked at the implementation yet, but is it still possible to "hard-code" a namespace into the object? Or do you have to use `{{.ObjectNamespace}}` when using a namespace selector now?

Using this test's scenario, would it be invalid to set the namespace on the object to be `default`? Because the namespace selector should select two namespaces?

controllers/configurationpolicy_controller.go
```diff
@@ -710,6 +714,145 @@ func currentlyUsingWatch(plc *policyv1.ConfigurationPolicy) bool {
...
+		if statusChanged {
+			r.recordInfoEvent(plc, true)
+		}
+
+		r.updatedRelatedObjects(plc, []policyv1.RelatedObject{})
+
+		// Note: don't clean up child objects when there is a template violation
+
+		r.addForUpdate(plc, statusChanged)
+	}()
```
JustinKuli, line 759: There are only two returns currently in this function where this defer is useful (`err != nil`). Are there plans to have more? I remember wanting to refactor something like this a while ago, there were quite a few more cases at that point though. For only two (and two returns where `err == nil`), I'm not sure if the complexity is worth it. But if it matches another function with more cases, or there is a good possibility of additional errors in this function to be handled in the future, then this is great.

JustinKuli: Seems like there are fewer instances of `addConditionToStatus(plc, -1,` than I remember. Maybe they got cleaned up some other way at some point.

I could go either way on this defer.

JustinKuli (approved):
I think it looks good overall!

test/e2e/case13_templatization_test.go
mprahl: @JustinKuli it's a bit confusing but if the user specifies a namespace explicitly, the namespace selector is ignored. If the user has an empty namespace or leverages the `{{ .ObjectNamespace }}` template variable, then the namespace selector is leveraged. If after template resolution the namespace on the object definition is not empty or the current namespace in the namespace selector, it is considered invalid.

This was brought up by @dhaiducek since without this restriction, you could create an object in another namespace than the one that was selected.

controllers/configurationpolicy_controller.go
mprahl: @JustinKuli I refactored this to remove the defer but still not duplicate the error handling. Let me know what you think!

controllers/configurationpolicy_controller.go
```diff
@@ -1045,277 +1216,255 @@ func (r *ConfigurationPolicyReconciler) handleDeletion(plc *policyv1.Configurati
...
-	if err != nil {
-		return parentStatusUpdateNeeded, err
+		if len(selectedNamespaces) != 0 {
+			relevantNamespaces = selectedNamespaces
+		}
+	} else if scopedGVR.Namespaced {
+		relevantNamespaces = []string{desiredNs}
+	} else {
+		// Cluster scoped
+		relevantNamespaces = []string{""}
```
dhaiducek, line 1348: With this conditional, I think there's a risk here that a cluster-scoped objectDefinition that specifies a namespace would work? If that's correct, we probably shouldn't cover up that human error.

```suggestion
	} else {
		relevantNamespaces = []string{desiredNs}
	}
```

controllers/configurationpolicy_controller.go
```diff
@@ -1045,277 +1216,255 @@ func (r *ConfigurationPolicyReconciler) handleDeletion(plc *policyv1.Configurati
...
+		desiredObj.SetNamespace(strings.TrimSpace(desiredObj.GetNamespace()))
 
-			// only report this error if there wasn't another yet
-			if errEvent == nil {
-				errEvent = &objectTmplEvalEvent{
-					compliant: false,
-					reason:    "namespaceSelector error",
-					message:   msg,
-				}
+		if desiredObj.GetNamespace() != desiredNs && desiredObj.GetNamespace() != ns {
```
dhaiducek, line 1451: Making sure I'm understanding this conditional correctly: this is ensuring that the resolved namespace is either an empty string (or a different marker if we so choose for `desiredNs`) or the namespace selected from the `namespaceSelector`?

EDIT: Actually, is there logic somewhere that allows skipping the empty stringed namespace?

dhaiducek: Oops--ignore that edit. Got my wires crossed with the ObjectName variable 🤦‍♂️

controllers/configurationpolicy_controller.go
mprahl: That is correct, but we already ignore the namespace if it's provided for a cluster scoped resource in `getObject` and in the kubernetes-dependency-watches library.

So it'd be an additional change if we started returning an error here.

dhaiducek: Oh, didn't know that. Well alrighty then!

test/e2e/case13_templatization_test.go
```diff
@@ -483,4 +461,90 @@ var _ = Describe("Test templatization", Ordered, func() {
...
+					clientManagedDynamic, gvrConfigPolicy, policyName, testNamespace, true, defaultTimeoutSeconds,
+				)
+
+				utils.CheckComplianceStatus(g, managedPlc, "Compliant")
+
+				relatedObjects, _, _ := unstructured.NestedSlice(managedPlc.Object, "status", "relatedObjects")
+				g.Expect(relatedObjects).To(HaveLen(2))
+
+				relatedObject1, ok := relatedObjects[0].(map[string]interface{})
+				g.Expect(ok).To(BeTrue(), "Related object was not a map")
```
dhaiducek, line 492: nit
```suggestion
				g.Expect(ok).To(BeTrue(), "Related object is not a map")
```

dhaiducek (approved):
Nice work! Couple nits/optional updates to the E2E tests, but I've approved it if you want to go ahead with the changes as-is.

dhaiducek (approved):

PR #315, opened 2024-11-13
Add templating to OperatorPolicy spec.versions

mprahl:
Relates:
https://issues.redhat.com/browse/ACM-15383


controllers/operatorpolicy_controller.go
```diff
@@ -733,6 +738,65 @@ func (r *OperatorPolicyReconciler) usingExistingSubIfFound(
...
+	for _, version := range policy.Spec.Versions {
+		if templates.HasTemplate([]byte(version), "", false) {
+			hasTemplate = true
+
+			break
+		}
+	}
+
+	if !hasTemplate {
+		return nil
```
JustinKuli, line 761: Should we remove the requirement that items in the `versions` list must be non-empty? Since we'll allow them to be empty after template resolution, it might be nice to consistently allow them without templates.

The question is whether these should all be allowed and behave the same:
- `[]`
- `[""]`
- `["{{ /* comment */ }}"]`
- `["{{ if false }}fake{{ end }}"]`

test/resources/case38_operator_install/operator-policy-no-group-one-version.yaml
```diff
@@ -22,5 +22,8 @@ spec:
     sourceNamespace: olm
     startingCSV: quay-operator.v3.10.0
   versions:
-    - quay-operator.v3.10.0
+    - '{{ if false }}some value to test empty string is discarded{{ end }}'
+    - '{{ if true }}quay-operator.v3.10.0{{ end }}'
+    - '{{ if false }}some other value to test empty string is discarded{{ end }}'
+    - '{{ if false }}some other value to test empty string is discarded{{ end }}'
```
yiraeChristineKim, line 28: These two are the same, Is this intended?

controllers/operatorpolicy_controller.go
mprahl: That's a good point. I'll adjust it.

test/resources/case38_operator_install/operator-policy-no-group-one-version.yaml
mprahl: It was just to further test the code of removing empty strings.

JustinKuli (dismissed):
Looks good to me! Thanks for updating the CRD field description.

deploy/crds/policy.open-cluster-management.io_operatorpolicies.yaml
```diff
@@ -206,10 +206,10 @@ spec:
                 type: string
               versions:
                 description: |-
-                  Versions is a list of non-empty strings that specifies which installed versions are compliant
-                  when in `inform` mode and which `InstallPlans` are approved when in `enforce` mode.
+                  Versions is a list of templatable strings that specifies which installed ClusterServiceVersion names are
+                  compliant when in `inform` mode and which `InstallPlans` are approved when in `enforce` mode. Empty or whitespace
+                  only strings are ignored. An empty list approves all ClusterServiceVersion names.
```
yiraeChristineKim, line 211: The default is empty right?

yiraeChristineKim:
love it Thanks!

PR #317, opened 2024-11-19
Add an `objectSelector` to ConfigurationPolicy

dhaiducek:
The `objectSelector` is implemented as a pure `LabelSelector` so it can be passed directly to a `List` operation in the dynamic watcher.

An unaddressed bug here is that `mustnothave` behaves in such a way that it doesn't check fields--it only checks the name and assumes it shouldn't exist. This PR is already so large, I'm thinking I'll create a followup issue to address it in a future PR.

ref: https://issues.redhat.com/browse/ACM-15389

- https://github.com/open-cluster-management-io/enhancements/issues/128


controllers/configurationpolicy_controller.go
```diff
@@ -1371,97 +1397,171 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
...
+
+			return nil, &scopedGVR, errEvent, err
+		}
+
+		filteredObjects, err := r.DynamicWatcher.List(plc.ObjectIdentifier(), objGVK, "", selector)
+		if err != nil {
+			log.Error(err, "Failed to fetch the resources",
+				"objectSelector", fmt.Sprint(objectT.ObjectSelector.String()))
+
+			msg := fmt.Sprintf("Error listing resources with provided objectSelector: %v", err)
```
mprahl, line 1427: Could you please include the object-template index in the error message?

controllers/configurationpolicy_controller.go
```diff
@@ -1371,97 +1397,171 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
...
 	desiredObjects := []*unstructured.Unstructured{}
+
+	// The object either has no name specified or it is templated in the object definition
+	if desiredName == "" && objectT.ObjectSelector != nil {
+		selector, err := metav1.LabelSelectorAsSelector(objectSelector)
+		if err != nil {
+			log.Error(err, "Failed to select the resources",
+				"objectSelector", fmt.Sprint(objectT.ObjectSelector.String()))
+
+			msg := fmt.Sprintf("Error parsing provided objectSelector: %v", err)
```
mprahl, line 1410: Could you please include the object-template index in the error message?

controllers/configurationpolicy_controller.go
```diff
@@ -1371,97 +1397,171 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
...
+			errEvent := &objectTmplEvalEvent{
+				compliant: false,
+				reason:    "objectSelector error",
+				message:   msg,
+			}
+
+			return nil, &scopedGVR, errEvent, err
+		}
+
+		filteredObjects, err := r.DynamicWatcher.List(plc.ObjectIdentifier(), objGVK, "", selector)
```
mprahl, line 1422: I'd rather perform a list query per selected namespace rather than perform client-side filtering. This should lead to less network traffic and memory usage.

controllers/configurationpolicy_controller.go
```diff
@@ -1371,97 +1397,171 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
...
 
-		// Process templating only on the first loop if the ObjectNamespace template variable isn't used
-		if tmplResolver != nil && hasTemplate && (desiredObj == nil || needsPerNamespaceTemplating) {
-			r.processedPolicyCache.Delete(plc.GetUID())
+			// Process templating only on the first loop if the ObjectNamespace template variable isn't used
+			if tmplResolver != nil && hasTemplate && (desiredObj == nil || needsPerNamespaceTemplating) {
+				r.processedPolicyCache.Delete(plc.GetUID())
 
-			templateContext := struct{ ObjectNamespace string }{ObjectNamespace: ns}
+				templateContext := struct{ ObjectNamespace string }{ObjectNamespace: ns}
```
mprahl, line 1461:
```suggestion
				templateContext := struct{ ObjectNamespace string; ObjectName string }{ObjectNamespace: ns, ObjectName: name}
```

controllers/configurationpolicy_controller.go
```diff
@@ -1371,97 +1397,171 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
...
+			resourceNs := res.GetNamespace()
+			// Add the object if the object is in a namespace returned by the namespaceSelector
+			if _, ok := relevantNsNames[resourceNs]; ok {
+				relevantNsNames[resourceNs] = append(relevantNsNames[resourceNs], res.GetName())
+			}
+		}
+	}
+
 	hasTemplate := templates.HasTemplate(objectT.ObjectDefinition.Raw, "", true)
 	needsPerNamespaceTemplating := templateHasObjectNamespaceRegex.Match(objectT.ObjectDefinition.Raw)
```
mprahl, line 1449: Once you add the `ObjectName` template variable, you'll need to update this regex or add a separate regex.

controllers/configurationpolicy_controller.go
```diff
@@ -1371,97 +1397,171 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
...
 
 	var desiredObj *unstructured.Unstructured
 
-	for _, ns := range relevantNamespaces {
-		var rawDesiredObject []byte
+	for ns, names := range relevantNsNames {
+		for _, name := range names {
```
mprahl, line 1454: An optimization can be to detect if `ObjectName` is available. If it's not, the template resolution can be skipped after the first name since the template result can't change.

controllers/configurationpolicy_controller.go
mprahl: If you choose to not implement the `ObjectName` template variable in this PR, then the template resolution should not be per object as the resolution at the namespace level can be reused for each object.

test/resources/case42_resource_selector/case42_results_policy.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: policy.open-cluster-management.io/v1
+kind: ConfigurationPolicy
+metadata:
+  name: case42-selector-results-e2e
+spec:
+  namespaceSelector:
+    include:
+      - case42-e2e-1
```
mprahl, line 8: Could you please include an additional included namespace? I want to see the compliance message in the test of what that looks like.

mprahl: Never mind, I see an example with an empty namespace selector which I forgot selects everything.

test/e2e/case42_resource_selection_test.go
```diff
@@ -0,0 +1,211 @@
...
+		}, defaultTimeoutSeconds, 1).Should(Equal(fmt.Sprintf(
+			"fakeapis [case42-1-e2e] found as specified in namespace %[1]s; "+
+				"fakeapis [case42-2-e2e] found as specified in namespace %[1]s; "+
+				"fakeapis [case42-3-e2e] found as specified in namespace %[1]s; "+
+				"fakeapis [case42-4-e2e] found as specified in namespace %[1]s; "+
+				"fakeapis [case42-5-e2e] found as specified in namespace %[1]s", targetNs)))
+	},
+		Entry("Empty selector", `{}`),
+		Entry("Empty selector", `{"matchLabels":{}}`),
+		Entry("Empty selector", `{"matchExpressions":[}`),
```
mprahl, line 90: It looks like there is a missing closing bracket. Do you know why it didn't fail?
```suggestion
		Entry("Empty selector", `{"matchExpressions":[]}`),
```

dhaiducek: Oh goodness. Because I never actually finished writing the test! 🤦‍♂️  The input args are never used, so it's re-running the same passing steps. 😅 I'll fix it.

controllers/configurationpolicy_controller.go
dhaiducek: I think I'd prefer to tackle the ObjectName in a subsequent PR since it'd have its own tests (and there's already enough in this PR). I've set a conditional to only run on the first loop for now, and I've created an issue to track the subsequent work: https://issues.redhat.com/browse/ACM-15809

controllers/configurationpolicy_controller.go
dhaiducek: Good catch on adding the index to the messages! 😄

controllers/configurationpolicy_controller.go
JustinKuli: Client-side filtering (currently in the PR):
- higher controller memory usage
- one big watch on the API server
- reconciles potentially triggered when objects change that would be excluded by the namespace selector

List per relevant namespace (proposal):
- lowest controller memory usage
- multiple watches on the API server *(I think this is the one downside)*
- fewer reconciles triggered

And I think we have all of the namespace information at this point in the flow to make those requests pretty easily, is that right @dhaiducek?

dhaiducek: Ah, I wasn't thinking about the namespaceSelector exclusions. Yes, it ought to be straightforward to perform watches by namespace. I'll update it.

test/e2e/case42_resource_selection_test.go
```diff
@@ -0,0 +1,211 @@
...
+		By("Verifying policy is noncompliant and returns no objects")
+		utils.Kubectl("patch", "--namespace=managed", "configurationpolicy", policyName, "--type=json",
+			fmt.Sprintf(objectSelectorPatchFmt, `{"matchLabels":{"selects":"nothing"}}`),
+		)
+		Eventually(func() interface{} {
+			managedPlc := utils.GetWithTimeout(clientManagedDynamic, gvrConfigPolicy,
+				policyName, testNamespace, true, defaultTimeoutSeconds)
+
+			return utils.GetStatusMessage(managedPlc)
+		}, defaultTimeoutSeconds, 1).Should(Equal(noMatchesMsg))
```
yiraeChristineKim, line 70: Why this procedure is included in this DescribeTable?

test/e2e/case42_resource_selection_test.go
```diff
@@ -0,0 +1,211 @@
...
+		}, defaultTimeoutSeconds, 1).ShouldNot(ContainSubstring(
+			fmt.Sprintf("fakeapis [case42-3-e2e] found but not as specified in namespace %s", targetNs)))
+	})
+
+	It("should become compliant when enforced", func() {
+		utils.Kubectl("patch", "--namespace=managed", "configurationpolicy", policyName, "--type=json",
+			`--patch=[{"op":"replace","path":"/spec/remediationAction","value":"enforce"}]`,
+		)
+	})
+})
```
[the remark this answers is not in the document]
dhaiducek: No, because `mustnothave` doesn't behave correctly. It will be addressed in https://issues.redhat.com/browse/ACM-15772.

policy-security.yml
JustinKuli, line 1: What's this file?

JustinKuli (commented):
Looks good overall to me, thanks for updating the List watches to be namespaced.

policy-security.yml
dhaiducek: It was added in a previous PR--looks to be some sort of test file that was included accidentally.

test/e2e/case42_resource_selection_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+			return utils.GetStatusMessage(managedPlc)
+		}, defaultTimeoutSeconds, 1).ShouldNot(ContainSubstring(
+			fmt.Sprintf("fakeapis [case42-3-e2e] found but not as specified in namespace %s", targetNs)))
+	})
+
+	It("should become compliant when enforced", func() {
+		utils.Kubectl("patch", "--namespace=managed", "configurationpolicy", policyName, "--type=json",
+			`--patch=[{"op":"replace","path":"/spec/remediationAction","value":"enforce"}]`,
+		)
```
yiraeChristineKim, line 209: Isn't it better we add message check?  To check that only 1, 2,4 are applied 
```
 message: fakeapis [case42-1-e2e] found as specified in namespace case42-e2e-2;
        fakeapis [case42-2-e2e] found as specified in namespace case42-e2e-2; fakeapis
        [case42-4-e2e] found as specified in namespace case42-e2e-2
      reason: K8s `must have` object already exists
```

yiraeChristineKim: Oh also another question, enforce with objectSelector will never create new resources right?

dhaiducek: > Isn't it better we add message check? To check that only 1, 2,4 are applied

That's fair since there isn't really another enforce test. I'll add it.

> enforce with objectSelector will never create new resources

Right, it can only discover and modify existing resources.

pkg/dryrun/policy.open-cluster-management.io_configurationpolicies.yaml
```diff
@@ -213,6 +213,54 @@ spec:
                         compared with objects on the cluster.
                       type: object
                       x-kubernetes-preserve-unknown-fields: true
+                    objectSelector:
+                      description: |-
+                        ObjectSelector defines the label selector for objects defined in the `objectDefinition`. If
+                        there is an object name defined in the `objectDefinition`, the `objectSelector` is ignored.
```
yiraeChristineKim, line 219: Can we add this as a test?

test/e2e/case42_resource_selection_test.go
dhaiducek: Oh goodness! There's not even any assertion here! 😂 I'll _definitely_ add it--thanks for catching it!

pkg/dryrun/policy.open-cluster-management.io_configurationpolicies.yaml
yiraeChristineKim, line 219: Should we add this as a  test?

yiraeChristineKim: "there is an object name defined in the `objectDefinition`, the `objectSelector` is ignored."

dhaiducek: Good call! I've added a test (and a code fix to address its failure! :sweat_smile:)

controllers/configurationpolicy_controller.go
```diff
@@ -1307,17 +1314,37 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
...
+	// If the name is templated, consider it not explicitly set
+	if templates.HasTemplate([]byte(desiredName), "", true) {
+		desiredName = ""
+	}
+
+	// Set up default name array to be used for each namespace. If no
+	// objectSelector is provided, add the desired name as the default.
+	defaultNamesPerNs := []string{}
+	objectSelector := objectT.ObjectSelector
+	fetchNames := desiredName == "" && objectSelector != nil
```
yiraeChristineKim, line 1339: Is this just boolean? Can we rename this? fetchNamea look like string array

dhaiducek: Good point. I was originally thinking I'd need it more than one place, but it looks like I don't. I'll just move it into the conditional directly.

controllers/configurationpolicy_controller.go
```diff
@@ -1371,97 +1399,183 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
...
+	// Detect templates
+	hasTemplate := templates.HasTemplate(objectT.ObjectDefinition.Raw, "", true)
+	needsPerNamespaceTemplating := false
+
+	// Detect .ObjectNamespace template context variables
+	if hasTemplate {
+		needsPerNamespaceTemplating = templateHasObjectNamespaceRegex.Match(objectT.ObjectDefinition.Raw)
+	}
+
+	var desiredObj *unstructured.Unstructured
```
yiraeChristineKim, line 1464: Can we add comment why this diff from desiredObjects

test/e2e/case42_resource_selection_test.go
```diff
@@ -0,0 +1,246 @@
...
+
+		Eventually(func() interface{} {
+			managedPlc := utils.GetWithTimeout(clientManagedDynamic, gvrConfigPolicy,
+				policyName, testNamespace, true, defaultTimeoutSeconds)
+
+			return utils.GetStatusMessage(managedPlc)
+		}, defaultTimeoutSeconds, 1).Should(Equal(fmt.Sprintf(
+			"fakeapis [case42-3-e2e] found as specified in namespace %s", targetNs)))
+	})
+})
```
yiraeChristineKim, line 246: Do we have templated ns case?

dhaiducek: Good question--that was covered in:
- https://github.com/open-cluster-management-io/config-policy-controller/pull/314

And I'll probably also reuse some of those test cases for the `.ObjectName` variable.

controllers/configurationpolicy_controller.go
dhaiducek: I've added some comments--let me know if they make sense!

yiraeChristineKim (approved):

PR #318, opened 2024-11-22
Fix alreadyEvaluated check for dryrun

JustinKuli:
It appears that the fake API server used by dryrun does not supply resource versions, so when multiple object templates were supplied in a ConfigurationPolicy to dryrun, it would incorrectly think that the later templates had already been evaluated and didn't need to be updated.

Now if the resource version is empty, it will not consider the object as already evaluated.

Refs:
 - https://issues.redhat.com/browse/ACM-15899


dhaiducek (approved):
LGTM!

yiraeChristineKim (approved):

PR #319, opened 2024-11-25
Add `.ObjectName` template context variable

dhaiducek:
Allows leveraging the results from the `objectSelector`. 

ref: https://issues.redhat.com/browse/ACM-15809


controllers/configurationpolicy_controller.go
```diff
@@ -1560,6 +1570,30 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
 				return nil, &scopedGVR, errEvent, nil
 			}
 
+			// Skip adding objects named $SKIP -- This allows users to leverage
+			// templates to further filter results from the objectSelector
+			if desiredObj.GetName() == "$SKIP" {
```
mprahl, line 1575: I think instead of the `$SKIP`, we wanted a separate template function that could be called at any point to skip the object. This would be a better experience since you could call it when nested down in conditionals unrelated to the name.

dhaiducek: Ah, indeed. That makes sense--I forgot about that detail. I've removed the `$SKIP` logic.

controllers/configurationpolicy_controller.go
```diff
@@ -1457,9 +1458,11 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
 	// Detect templates
 	hasTemplate := templates.HasTemplate(objectT.ObjectDefinition.Raw, "", true)
 	needsPerNamespaceTemplating := false
+	needsPerNameTemplating := false
 
-	// Detect .ObjectNamespace template context variables
+	// Detect .objectNamespace and .ObjectName template context variables
```
mprahl, line 1463:
```suggestion
	// Detect .ObjectNamespace and .ObjectName template context variables
```

mprahl (changes requested):
This looks great! You just need to fix the one typo in the comment and then I'll approve the PR.

PR #321, opened 2024-12-05
Fix when context variable isn't used for metadata

dhaiducek:
When the `.ObjectName` template variable was used outside `.metadata.name`, the controller complained that the name didn't match.

This also removes an errant E2E test that was testing that a template that resolved to an empty string when the `objectSelector` was in play was NonCompliant, but really an empty string would be expected there regardless so that the controller could fill in the value.

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -1584,43 +1584,43 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
 				return nil, &scopedGVR, errEvent, err
 			}
 
+			// Populate the namespace and name if applicable
+			if desiredObj.GetName() == "" && name != "" {
+				desiredObj.SetName(strings.TrimSpace(name))
+			}
+
+			if desiredObj.GetNamespace() == "" && ns != "" {
+				desiredObj.SetNamespace(strings.TrimSpace(ns))
+			}
+
 			// strings.TrimSpace() is needed here because a multi-line value will have
 			// '\n' in it. This is kept for backwards compatibility.
 			desiredObj.SetName(strings.TrimSpace(desiredObj.GetName()))
 			desiredObj.SetKind(strings.TrimSpace(desiredObj.GetKind()))
 			desiredObj.SetNamespace(strings.TrimSpace(desiredObj.GetNamespace()))
 
-			// If the namespace doesn't match the original, return an error
-			if needsPerNamespaceTemplating && desiredObj.GetNamespace() != ns {
+			// Error if the namespace doesn't match the parsed namespace from the namespaceSelector
+			if !plc.Spec.NamespaceSelector.IsEmpty() && desiredObj.GetNamespace() != ns {
 				errEvent := &objectTmplEvalEvent{
 					compliant: false,
 					reason:    reasonTemplateError,
-					message: "The object definition's namespace must match the selected namespace after template " +
-						"resolution",
+					message: "The object definition's namespace must match the result " +
+						"from the namespace selector after template resolution",
 				}
 
 				return nil, &scopedGVR, errEvent, nil
 			}
 
-			// If the name doesn't match the original, return an error
-			if needsPerNameTemplating && desiredObj.GetName() != name {
+			// Error if the name doesn't match the parsed name from the objectSelector
+			if objectSelector != nil && desiredObj.GetName() != name {
 				errEvent := &objectTmplEvalEvent{
 					compliant: false,
 					reason:    reasonTemplateError,
-					message: "The object definition's name must match the selected name after template " +
-						"resolution",
+					message: "The object definition's name must match the result " +
+						"from the object selector after template resolution",
 				}
 
 				return nil, &scopedGVR, errEvent, nil
-			}
-
-			// Populate the namespace and name if they're not empty strings
-			if name != "" {
-				desiredObj.SetName(strings.TrimSpace(name))
-			}
-
-			if ns != "" {
-				desiredObj.SetNamespace(strings.TrimSpace(ns))
 			}
 
 			desiredObjects = append(desiredObjects, desiredObj)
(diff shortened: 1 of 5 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -1602,8 +1611,8 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
 				return nil, &scopedGVR, errEvent, nil
 			}
 
-			// If the name doesn't match the original, return an error
-			if needsPerNameTemplating && desiredObj.GetName() != name {
+			// Error if the name doesn't match the parsed name from the objectSelector
+			if objectSelector != nil && desiredObj.GetName() != name {
```
dhaiducek, line 1615: I reconsidered this block _a lot_ and even considered removing it, not wanting to unnecessarily throttle the capabilities of policy users. I think the logic here makes sense now. My initial concern was the ability to set the name or namespace to something other than what was determined by the `namespaceSelector` or `objectSelector`, which seemed like somewhat of an antipattern. But I could be swayed if the logic seems off or there are cases I'm not considering where setting the name or namespace to something else could be beneficial somehow.

JustinKuli: I'm concerned that if I create a policy that uses object selector, and at first it only selects one object, I can use that name in the policy. But if later another object is created that matches the selector, suddenly by policy will have this error? Is that correct?

If that's the current behavior (I could've misunderstood), I think it might be better if we *always* report an error if the policy defines both the name and the object selector. Does that make sense?

I was also worried a little bit about the error message, since it doesn't give the details of what the "selected name" is - if the error was just something like `name may not be set when object selector is used`, it might be more clear to the user how to fix it.

yiraeChristineKim (dismissed):

yiraeChristineKim:
LGTM but hold for Justin 
/hold

controllers/configurationpolicy_controller.go
yiraeChristineKim: @JustinKuli GOOD Point!!

dhaiducek: @JustinKuli Hmmm. The differing error behavior is a good point. We _could_ forbid users from using templated strings for the name or namespace if the `objectSelector` or `namespaceSelector` are in use, since that's really the complexity here. (Providing a non-templated string overrides the selector as it does with the `namespaceSelector`.) Does that align with your point? It'd seem to align with your point about the error message also.

JustinKuli: > Providing a non-templated string overrides the selector as it does with the namespaceSelector.

I hadn't remembered that detail. It makes sense for namespace, since I think I've been guilty of just copying a default namespaceSelector into new policies, and still setting the namespace in a template to whatever I want. And then for consistency, it makes sense to do the same thing for object selector and names... I have a feeling if we were building this from scratch I might try to make slightly different decisions, but it's too late now.

I don't think I'd want us to forbid templated strings entirely here, since `name: {{.ObjectName}}` "feels right". But `name: {{.ObjectName}}-new` definitely feels wrong. So the conditional statement that is currently there seems right. And now, after talking about it for a while, I finally understand what the error message is trying to say. But I don't have a great way to reword it to be more clear immediately.

JustinKuli: Maybe something to the effect "templates may not change the name of the object when an object selector is used"

yiraeChristineKim: "templates may not change the name of the object when an object selector is used" is the only case for "enforce". sound like enforce log to me

dhaiducek: @yiraeChristineKim It's an interesting point, since if users do crazy things with templating while in `inform` it doesn't _really_ affect anything, but to @JustinKuli's first point it might be a better experience to have a consistent experience where if the policy is flipped to `enforce` they'd get the same message.

> ...if we were building this from scratch I might try to make slightly different decisions...

@JustinKuli It hasn't been delivered anywhere yet--unless you're referring to existing behaviors, I don't think it's too late to reconsider logic around the context variables and `objectSelector`.

JustinKuli (approved):
The new message looks better, thank you!

controllers/configurationpolicy_controller.go
yiraeChristineKim: good point!!  @JustinKuli  finding edge cases!!!

dhaiducek:
Unrelated E2E failure:
```
• [FAILED] [65.762 seconds]
Testing OperatorPolicy Test reporting of unapproved version after installation [It] Should report a violation after the versions list is patched to exclude the current version [supports-hosted]
/home/runner/work/config-policy-controller/config-policy-controller/test/e2e/case38_install_operator_test.go:3697

  Timeline >>
  STEP: Creating the parent object @ 12/09/24 20:49:32.834
  STEP: Creating the child object with the owner reference @ 12/09/24 20:49:32.96
  STEP: Verifying the child object exists @ 12/09/24 20:49:32.965
  STEP: Patching the versions field to exclude the installed version @ 12/09/24 20:49:32.967
  [FAILED] in [It] - /home/runner/work/config-policy-controller/config-policy-controller/test/e2e/case38_install_operator_test.go:3702 @ 12/09/24 20:50:33.028

  [FAILED] Timed out after 60.000s.
  The function passed to Eventually failed at /home/runner/work/config-policy-controller/config-policy-controller/test/e2e/case38_install_operator_test.go:153 with:
  Should have related object ClusterServiceVersion with name 'quay-operator.v3.10.6'
  Expected
      <bool>: false
  to be true
  In [It] at: /home/runner/work/config-policy-controller/config-policy-controller/test/e2e/case38_install_operator_test.go:3702 @ 12/09/24 20:50:33.028
```

ref: https://github.com/open-cluster-management-io/config-policy-controller/actions/runs/12243701334/job/34153830369?pr=321#step:5:28208
```

dhaiducek:
/unhold

CI is passing--ready for another review!

JustinKuli (approved):

PR #323, opened 2024-12-06
Consolidate compliance messages

yiraeChristineKim:
Currently the compliance message is listed out for each object from the object selector. Ideally these messages would be consolidated for better consumption and readability.

Definition of Done for Engineering Story Owner (Checklist)
Status messages from the objectSelector are consolidated rather than listed out individually for each name.

Only combine messages for the same object-template (when it uses a resource selector, there can be multiple objects)

example of change:
`configmaps [rex, woody] found but not as specified in namespace toy-story1; configmaps [buzz, potato] found but not as specified in namespace toy-story2`
Ref: https://issues.redhat.com/browse/ACM-15773

### DIFF
--- a/controllers/configurationpolicy_controller_test.go
+++ b/controllers/configurationpolicy_controller_test.go
@@ -790,8 +790,8 @@ func TestCreateStatus(t *testing.T) {
 			},
 			true,
 			"K8s `must have` object already exists",
-			"secrets [bo-peep] found as specified in namespace toy-story4; secrets [buzz] found as specified in " +
-				"namespace toy-story",
+			"secrets [buzz] found as specified in namespace toy-story; " +
+				"secrets [bo-peep] found as specified in namespace toy-story4",
 		},
 		{
 			"must have single object created",
@@ -958,6 +958,34 @@ func TestCreateStatus(t *testing.T) {
 			"K8s missing namespace",
 			"namespaced object of kind ConfigMap has no namespace specified from the policy namespaceSelector " +
 				"nor the object metadata",
+		},
+		{
+			"unnamed object multiple error",
+			"configmaps",
+			map[string]*objectTmplEvalResultWithEvent{
+				"toy-story1": {
+					result: objectTmplEvalResult{
+						objectNames: []string{"rex", "woody"},
+					},
+					event: objectTmplEvalEvent{
+						compliant: false,
+						reason:    reasonWantFoundNoMatch,
+					},
+				},
+				"toy-story2": {
+					result: objectTmplEvalResult{
+						objectNames: []string{"buzz", "potato"},
+					},
+					event: objectTmplEvalEvent{
+						compliant: false,
+						reason:    reasonWantFoundNoMatch,
+					},
+				},
+			},
+			false,
+			"K8s does not have a `must have` object",
+			"configmaps [rex, woody] found but not as specified in namespace toy-story1; " +
+				"configmaps [buzz, potato] found but not as specified in namespace toy-story2",
 		},
 		{
 			"multiple errors",
--- a/controllers/configurationpolicy_utils.go
+++ b/controllers/configurationpolicy_utils.go
@@ -565,7 +565,7 @@ func createStatus(
 			_, namesStr, _ := strings.Cut(nsName, "/")
 
 			if len(nsNameToEvent[nsName].result.objectNames) > 0 {
-				namesStr = " [" + strings.Join(nsNameToEvent[nsName].result.objectNames, ", ") + "]"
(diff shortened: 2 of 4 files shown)


controllers/configurationpolicy_utils.go
```diff
@@ -577,98 +577,161 @@ func createStatus(
...
 
-				compliancyDetailsReason += generatedReason
+				compliancyDetailsReason += reason
 			}
 
-			compliancyDetailsMsg += generatedMsg
+			for j, nsName := range objectNameStrsToNsNames[namesStr] {
+				if j != 0 {
+					compliancyDetailsMsg = setCompliancyDetailsMsgEnd(compliancyDetailsMsg)
+				}
```
JustinKuli, line 687: If `j == 0`, then the current message would be empty, right? So then `setCompliancyDetailsMsgEnd` would not add a semicolon?

I think checking the index and the message in `setCompliancyDetailsMsgEnd` is redundant. I'd prefer to just use the index, and add the semicolon directly:

```suggestion
			for j, nsName := range objectNameStrsToNsNames[namesStr] {
				if j != 0 {
					compliancyDetailsMsg += "; "
				}
```

controllers/configurationpolicy_utils.go
```diff
@@ -577,98 +577,161 @@ func createStatus(
...
+
+		compliancyDetailsMsg = setCompliancyDetailsMsgEnd(compliancyDetailsMsg)
+
+		joinedNames := strings.Join(names, ", ")
+		if joinedNames != "" {
+			joinedNames = fmt.Sprintf(" [%s]", joinedNames)
+		}
+
+		compliancyDetailsMsg += fmt.Sprintf(msgTemplate, resourceName, joinedNames)
+	}
```
JustinKuli, line 623: Assuming the index approach works, I think that would be more clear here as well:
```suggestion
	for j, msgTemplate := range msgTemplates {
		names := msgMap[msgTemplate]
		slices.Sort(names) // Sort names for consistent output

		if j != 0 {
			compliancyDetailsMsg += "; "
		}
		
		joinedNames := strings.Join(names, ", ")
		if joinedNames != "" {
			joinedNames = fmt.Sprintf(" [%s]", joinedNames)
		}

		compliancyDetailsMsg += fmt.Sprintf(msgTemplate, resourceName, joinedNames)
	}
```

controllers/configurationpolicy_utils.go
```diff
@@ -577,98 +577,161 @@ func createStatus(
...
+
+	return compliancyDetailsMsg
+}
 
-			// This prevents repeating the same reason for each unique object name list.
+// Process the object name strings in order to ensure a deterministic reason and message.
```
JustinKuli, line 628: Nit: since this function returns a map (and lookups from maps in go can be in an inconsistent order), it is the `getCombinedCompliancyDetailsMsg` and `createStatus` functions which ensures they are in a deterministic order.

Also, I just want to thank you for bringing this into a separate function! It really helps readability of `createStatus` :1st_place_medal:

policy.yaml
```diff
@@ -0,0 +1,36 @@
...
+    #     matchExpressions:
+    #       - key: yikim
+    #         operator: Exists
+    #   objectDefinition:
+    #     apiVersion: config-policy-controller.io/v1
+    #     kind: FakeAPI
+    #     metadata:
+    #       labels:
+    #         selected: "true"
+    #       namespace: case42-e2e-2
```
JustinKuli, line 36: Are these commented bits necessary?

controllers/configurationpolicy_controller_test.go
```diff
@@ -959,6 +959,34 @@ func TestCreateStatus(t *testing.T) {
...
+						compliant: false,
+						reason:    reasonWantFoundNoMatch,
+					},
+				},
+			},
+			false,
+			"K8s does not have a `must have` object",
+			"configmaps [rex, woody] found but not as specified in namespace toy-story1; " +
+				"configmaps [buzz, potato] found but not as specified in namespace toy-story2",
+		},
```
JustinKuli, line 989: I love when the GH review begins with an easy-to-read test! This looks great

JustinKuli (changes requested):
Just some small things, but overall this is working great

policy.yaml
yiraeChristineKim: OH my god this should be deleted

controllers/configurationpolicy_utils.go
yiraeChristineKim: You are right, how about move "// Process the object name strings in order to ensure a deterministic reason and message." above "getCombinedCompliancyDetailsMsg" as function comment

JustinKuli (approved):

PR #324, opened 2024-12-13
Update OperatorPolicy `operatorGroup` description

dhaiducek:
Add note that the `AllNamespaces` type is the default.


api/v1beta1/operatorpolicy_types.go
```diff
@@ -148,17 +148,22 @@ type OperatorPolicySpec struct {
 	// +kubebuilder:validation:Enum=musthave;mustnothave
 	ComplianceType policyv1.ComplianceType `json:"complianceType"`
 
-	// OperatorGroup specifies which `OperatorGroup` to inspect. Include the name, namespace, and any
-	// `spec` fields for the operator group. For more info, see `kubectl explain operatorgroups.spec`
-	// or view https://olm.operatorframework.io/docs/concepts/crds/operatorgroup/.
+	// OperatorGroup specifies which `OperatorGroup` to inspect. This resource is generated by the
+	// operator policy controller. Include the name, namespace, and any `spec` fields for the operator
+	// group. By default, if the `operatorGroup` field is not specified, the controller generates an
+	// `AllNamespaces` type `OperatorGroup` in the same namespace as the subscription, if supported.
```
JustinKuli, line 154: Small update to clarify when it would create the resource.
```suggestion
	// group. By default, if the `operatorGroup` field is not specified, and no OperatorGroup already exists in the namespace, then the controller generates an
	// `AllNamespaces` type `OperatorGroup` in the same namespace as the subscription, if supported.
```

JustinKuli (changes requested):

api/v1beta1/operatorpolicy_types.go
dhaiducek: Good addition! We'll have to update the docs 🙂 

https://github.com/stolostron/rhacm-docs/blob/db98a8d85a8bd1f24146af3ab99475424481ad59/governance/operator_pol_ctrl.adoc?plain=1#L43

yiraeChristineKim (approved):

JustinKuli (approved):

PR #325, opened 2024-12-17
Fix objectSelector polling

dhaiducek:
The objectSelector was always running in watch mode regardless of the evaluationInterval.

ref: https://issues.redhat.com/browse/ACM-15952


JustinKuli (approved):
Nice tests!

PR #327, opened 2025-01-22
Add deprecation status in operatorpolicy

yiraeChristineKim:
OperatorPolicy is defined for an operator which is either running a deprecated version, or a channel for updates which is deprecated, the policy should report this in its status 

Additionally, resolve the flaky OperatorPolicy issue by switching from the Quay operator to the GRC fake operator.

ref: https://issues.redhat.com/browse/ACM-16120

### DIFF
--- a/api/v1beta1/operatorpolicy_types.go
+++ b/api/v1beta1/operatorpolicy_types.go
@@ -134,6 +134,13 @@ type ComplianceConfig struct {
 	//
 	//+kubebuilder:default=Compliant
 	UpgradesAvailable ComplianceConfigAction `json:"upgradesAvailable,omitempty"`
+	// DeprecationsPresent specifies how the overall policy compliance is affected by deprecations.
+	// The default value is `Compliant`. If any deprecations are detected
+	// while DeprecationsPresent = NonCompliant,
+	// then the policy compliance will be set to `NonCompliant`.
+	//
+	// +kubebuilder:default=Compliant
+	DeprecationsPresent ComplianceConfigAction `json:"deprecationsPresent,omitempty"`
 }
 
 // OperatorPolicySpec defines the desired state of a particular operator on the cluster.
--- a/controllers/operatorpolicy_status.go
+++ b/controllers/operatorpolicy_status.go
@@ -283,7 +283,8 @@ func calculateComplianceCondition(policy *policyv1beta1.OperatorPolicy) metav1.C
 	}
 
 	idx, cond = policy.Status.GetCondition(deprecationType)
-	if !policy.Spec.ComplianceType.IsMustNotHave() {
+	if !policy.Spec.ComplianceType.IsMustNotHave() &&
+		policy.Spec.ComplianceConfig.DeprecationsPresent == "NonCompliant" {
 		if idx == -1 {
 			messages = append(messages, "The deprecation status is unknown")
 			foundNonCompliant = true
--- a/deploy/crds/kustomize_operatorpolicy/policy.open-cluster-management.io_operatorpolicies.yaml
+++ b/deploy/crds/kustomize_operatorpolicy/policy.open-cluster-management.io_operatorpolicies.yaml
@@ -68,6 +68,17 @@ spec:
                     - Compliant
                     - NonCompliant
                     type: string
+                  deprecationsPresent:
+                    default: Compliant
+                    description: |-
+                      DeprecationsPresent specifies how the overall policy compliance is affected by deprecations.
+                      The default value is `Compliant`. If any deprecations are detected
+                      while DeprecationsPresent = NonCompliant,
+                      then the policy compliance will be set to `NonCompliant`.
+                    enum:
+                    - Compliant
+                    - NonCompliant
+                    type: string
                   upgradesAvailable:
                     default: Compliant
                     description: |-
--- a/deploy/crds/policy.open-cluster-management.io_operatorpolicies.yaml
+++ b/deploy/crds/policy.open-cluster-management.io_operatorpolicies.yaml
@@ -70,6 +70,17 @@ spec:
                     - Compliant
                     - NonCompliant
                     type: string
+                  deprecationsPresent:
+                    default: Compliant
+                    description: |-
+                      DeprecationsPresent specifies how the overall policy compliance is affected by deprecations.
+                      The default value is `Compliant`. If any deprecations are detected
+                      while DeprecationsPresent = NonCompliant,
+                      then the policy compliance will be set to `NonCompliant`.
+                    enum:
+                    - Compliant
+                    - NonCompliant
+                    type: string
                   upgradesAvailable:
                     default: Compliant
                     description: |-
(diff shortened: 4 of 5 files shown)


controllers/operatorpolicy_controller.go
```diff
@@ -1440,6 +1445,128 @@ func (r *OperatorPolicyReconciler) musthaveSubscription(
...
+					message, _, _ := unstructured.NestedString(bundleDeprecation, "message")
+					updateStatus(policy, deprecationCond(entryName, Bundle, message))
+
+					return nil
+				}
+			}
+		}
+	}
+
+	return nil
```
JustinKuli, line 1567: I think it should update the deprecation condition here to reflect that nothing deprecated was found. Otherwise, if I have a deprecated version, then I update to something supported, I think the OperatorPolicy will continue to report the (stale) deprecation.

JustinKuli (changes requested):
This is looking great overall! Awesome to see the fake operator for this.

In addition to the "nothing deprecated" condition I mentioned, we should conditionally report the deprecation in the overall status: https://github.com/open-cluster-management-io/config-policy-controller/blob/1915491b8c03384b34ce2985538adc81441b197d/controllers/operatorpolicy_status.go#L178

If you'd prefer, I'd be ok with that addition in a separate PR.

yiraeChristineKim:
> This is looking great overall! Awesome to see the fake operator for this.
> 
> In addition to the "nothing deprecated" condition I mentioned, we should conditionally report the deprecation in the overall status:
> 
> https://github.com/open-cluster-management-io/config-policy-controller/blob/1915491b8c03384b34ce2985538adc81441b197d/controllers/operatorpolicy_status.go#L178
> 
> If you'd prefer, I'd be ok with that addition in a separate PR.

so my understanding what you suggest, should deprecation status affect compliance(NoCompliant or Compliant)

controllers/operatorpolicy_status.go
```diff
@@ -367,6 +379,7 @@ const (
 	deploymentConditionType  = "DeploymentCompliant"
 	catalogSrcConditionType  = "CatalogSourcesUnhealthy"
 	installPlanConditionType = "InstallPlanCompliant"
+	deprecationType          = "DeprecationWarning"
```
JustinKuli, line 382: The other ConditionTypes we have use "positive polarity" - when the condition is true, that part of the policy is compliant. It would be nice to keep that convention here, so this could be renamed to something like `NoDeprecations`

(The one exception to this is the CatalogSource; we copy the OLM status which uses a negative polarity)

controllers/operatorpolicy_controller.go
```diff
@@ -1440,6 +1445,135 @@ func (r *OperatorPolicyReconciler) musthaveSubscription(
...
+				}
+			}
+		}
+	}
+
+	updateStatus(policy, metav1.Condition{
+		Type:    "DeprecationWarning",
+		Status:  metav1.ConditionTrue,
+		Reason:  "Recommended",
+		Message: "The requested package, channel, and bundle are all at the recommended versions",
```
JustinKuli, line 1571: This should use the constant defined in `_status.go`.

controllers/operatorpolicy_status.go
```diff
@@ -552,6 +565,30 @@ func updatedCond(kind string) metav1.Condition {
...
+)
+
+func (s Schema) toString() string {
+	return string(s)
+}
+
+// deprecationCond returns a Compliant condition (without affecting the condition itself),
+// and a message in the format: 'The ____ (Package, Channel, Bundle) was deprecated.'
+// + additional details from packageManifest.
+func deprecationCond(depSchemaName string, depSchema Schema, message string) metav1.Condition {
```
JustinKuli, line 583: If we don't want this to affect the final compliance, we should ensure that this returns a `Status` that matches whatever is Compliant. Right now it doesn't.

To allow users to determine if they want the deprecation status to affect the overall compliance, we'll need to add something to the `ComplianceConfig`, and pass that into this function. For example, `buildDeploymentCond` uses that pattern.

controllers/operatorpolicy_controller.go
yiraeChristineKim: Type or all of this

controllers/operatorpolicy_status.go
```diff
@@ -552,6 +565,30 @@ func updatedCond(kind string) metav1.Condition {
...
+	return metav1.Condition{
+		Type:    deprecationType,
+		Status:  metav1.ConditionFalse,
+		Reason:  depSchema.toString() + "Deprecated",
+		Message: "the requested " + depSchemaName + " " + depSchema.toString() + " was deprecated. " + message,
```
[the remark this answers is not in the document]
yiraeChristineKim: When a package is deprecated, deprecationType -> NoDeprecation, status is false, Reason is PackageDeprecated Is this correct?

controllers/operatorpolicy_controller.go
JustinKuli: Just `Type`

controllers/operatorpolicy_status.go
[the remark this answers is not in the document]
JustinKuli: Correct

api/v1beta1/operatorpolicy_types.go
```diff
@@ -134,6 +134,13 @@ type ComplianceConfig struct {
 	//
 	//+kubebuilder:default=Compliant
 	UpgradesAvailable ComplianceConfigAction `json:"upgradesAvailable,omitempty"`
+	// DeprecationAvailable specifies how the NoDeprecations condition impacts overall
+	// policy compliance. The default value is `Compliant`. If any deprecations are detected
+	// in the package, channel, or bundle with DeprecationAvailable = NonCompliant,
+	// then the policy compliance will be set to `Compliant`.
+	//
+	// +kubebuilder:default=Compliant
+	DeprecationAvailable ComplianceConfigAction `json:"deprecationAvailable,omitempty"`
```
JustinKuli, line 143: Maybe `DeprecationsPresent` would make this more clear?

```suggestion
	// DeprecationsPresent specifies how the overall policy compliance is affected by deprecations.
	// The default value is `Compliant`. If any deprecations are detected
	// while DeprecationsPresent = NonCompliant,
	// then the policy compliance will be set to `NonCompliant`.
	//
	// +kubebuilder:default=Compliant
	DeprecationsPresent ComplianceConfigAction `json:"deprecationsPresent,omitempty"`
```

controllers/operatorpolicy_status.go
```diff
@@ -282,6 +282,21 @@ func calculateComplianceCondition(policy *policyv1beta1.OperatorPolicy) metav1.C
 		}
 	}
 
+	idx, cond = policy.Status.GetCondition(deprecationType)
+	if !policy.Spec.ComplianceType.IsMustNotHave() &&
+		policy.Spec.ComplianceConfig.DeprecationAvailable == "Compliant" {
```
JustinKuli, line 287: Other conditions change the `Status` of the condition based on the complianceConfig, but I feel like your approach here is better.

If you follow my other suggestion, I think this part would need to be updated like this:
```suggestion
		policy.Spec.ComplianceConfig.DeprecationsPresent == "NonCompliant" {
```

api/v1beta1/operatorpolicy_types.go
yiraeChristineKim: genius

JustinKuli (approved):
LGTM!

PR #328, opened 2025-01-31
Account for nil annotations in uninstall

JustinKuli:
`.GetAnnotations()` can return a `nil` map, which can not be written to without panicking. These cases are now addressed in the triggeruninstall code.

Refs:
 - https://issues.redhat.com/browse/ACM-17562


dhaiducek (approved):
LGTM!

yiraeChristineKim (approved):

PR #329, opened 2025-02-04
Add DryCLI status comparison

yiraeChristineKim:
Screenshot 2025-02-04 at 11 51 51 AM
Ref: https://issues.redhat.com/browse/ACM-17517


pkg/dryrun/dryrun_utils.go
```diff
@@ -0,0 +1,264 @@
...
+func compareStatus(cmd *cobra.Command,
+	inputStatus map[string]interface{},
+	resultStatus map[string]interface{},
+) {
+	cmd.Println("# Status compare:")
+
+	isStatusMatch := compareStatusObj(cmd, true, "", inputStatus, resultStatus)
+
+	if isStatusMatch {
+		cmd.Println(successColor("\033[1m Expected status matches the actual status \033[0m"))
```
yiraeChristineKim, line 32: bold

JustinKuli: If we'll have terminal escapes for bold, I wonder if we should just "manually" do the terminal escapes for the colors ourselves, as opposed to using that library. How much trouble do you think that would it be?

pkg/dryrun/dryrun_utils.go
```diff
@@ -0,0 +1,264 @@
...
+
+				continue
+			}
+
+			if len(typedV) == 0 {
+				continue
+			}
+
+			switch typedV[0].(type) {
+			case map[interface{}]interface{}, map[string]interface{}:
```
JustinKuli, line 97: Isn't `map[string]interface{}` a sub-case of `map[interface{}]interface{}`? Is it necessary to make it a specific case?

pkg/dryrun/dryrun_utils.go
```diff
@@ -0,0 +1,264 @@
...
+				notArrayPrint(cmd, path, key)
+
+				isStatusMatch = false
+
+				continue
+			}
+
+			if len(typedV) == 0 {
+				continue
+			}
```
JustinKuli, line 94: Maybe I need to look at some specific examples, but I think it's possible a user would put in an empty list somewhere and want the result to also have an empty list there.

pkg/dryrun/dryrun_utils.go
JustinKuli: It seems like other pieces here want to only deal with `map[string]interface`. So I think that could be made a `case` on its own, and then if the next case was `map[interface{}]interface{}`, that would catch the errors all together. Might be cleaner (assuming it actually works like that).

pkg/dryrun/dryrun_utils.go
```diff
@@ -0,0 +1,264 @@
...
+
+			if v != av {
+				errorMatchPrint(cmd, path, v, av)
+
+				isStatusMatch = false
+			} else {
+				successMatchPrint(cmd, path, v, av)
+			}
+
+		case []interface{}:
```
JustinKuli, line 81: For readability, is it at all possible to split this case into another function? Or maybe just split the map case in the switch to another function?

pkg/dryrun/dryrun_utils.go
```diff
@@ -0,0 +1,264 @@
...
+				failedParsePrint(cmd, path)
+
+				isStatusMatch = false
+			}
+		case interface{}:
+			typeOfTypeV := reflect.TypeOf(typedV).String()
+			typeOfActual := reflect.TypeOf(actualValue).String()
+
+			if typeOfTypeV == "map[interface {}]interface {}" &&
+				typeOfActual == "map[interface {}]interface {}" {
```
JustinKuli, line 171: Is this not covered by the `map[interface{}]interface{}` case above? Is there a test case that actually reaches this point?

pkg/dryrun/dryrun_utils.go
```diff
@@ -0,0 +1,264 @@
...
+					continue
+				}
+
+				isStatusMatch = compareStatusObj(cmd, isStatusMatch, path, mapObj, mapActualObj)
+
+				continue
+			}
+
+			if typeOfTypeV != typeOfActual {
+				cmd.Println(errorColor(fmt.Sprintf("The path '%s' does not match at all", path)))
```
JustinKuli, line 192: We might be able to say something like "The input status field '`%s`' is not the right type - expected '`<go type>`'"

Because the actual result should be statically typed, we can think of its fields as always correct, I bet.

pkg/dryrun/dryrun_utils.go
```diff
@@ -0,0 +1,264 @@
...
+
+				continue
+			}
+
+			cmd.Println(errorColor(fmt.Sprintf("Unsupported type %T at path '%s' for key '%s'", typedV, path, key)))
+		default:
+			cmd.Println(errorColor(fmt.Sprintf("Unsupported type %T at path '%s' for key '%s'", typedV, path, key)))
```
JustinKuli, line 199: Is this ever reached in tests? I'd expect `case interface{}` to grab everything, and maybe we should just make it the `default` case.

pkg/dryrun/dryrun_utils.go
```diff
@@ -0,0 +1,264 @@
...
+			switch typedV[0].(type) {
+			case map[interface{}]interface{}, map[string]interface{}:
+				allMatch := true
+				// Check all input status are match
+				for i, obj := range typedV {
+					elementMatch := false
+
+					mapObj, ok := convertMapStringKey(obj).(map[string]interface{})
+					if !ok {
+						failedParsePrint(cmd, path)
```
JustinKuli, line 105: I think we can consider it not a match if we fail to parse something. Warnings are too likely to be ignored in my opinion.

pkg/dryrun/dryrun_utils.go
```diff
@@ -0,0 +1,264 @@
...
+func failedParsePrint(cmd *cobra.Command, path string) {
+	cmd.Println(warningColor("failed to parse, skip " + path))
+}
+
+func notArrayPrint(cmd *cobra.Command, from, path string) {
+	cmd.Println(warningColor(fmt.Sprintf("skip %s, %s is not array,", path, from)))
+}
+
+func errorMatchPrint(cmd *cobra.Command, path string, inputVal, actualVal interface{}) {
+	cmd.Println(errorColor(fmt.Sprintf("%s: %v does not match %v", path, inputVal, actualVal)))
```
JustinKuli, line 215:
```suggestion
	cmd.Println(errorColor(fmt.Sprintf("%s: '%v' does not match '%v'", path, inputVal, actualVal)))
```

pkg/dryrun/dryrun_utils.go
```diff
@@ -0,0 +1,264 @@
...
+func errorMatchAnyArrayPrint(cmd *cobra.Command, path string) {
+	cmd.Println(errorColor(fmt.Sprintf("%s does not match any elements", path)))
+}
+
+func errorMatchArrayPrint(cmd *cobra.Command, path string) {
+	cmd.Println(errorColor(fmt.Sprintf("%s does not match the elements", path)))
+}
+
+func successMatchArrayPrint(cmd *cobra.Command, path string) {
+	cmd.Println(successColor(fmt.Sprintf("%s is founded", path)))
```
JustinKuli, line 227:
```suggestion
	cmd.Println(successColor(fmt.Sprintf("%s matches", path)))
```

I'm worried this message is very repetitive. Maybe it shouldn't be printed at all?

pkg/dryrun/testdata/test_config_status/output.txt
```diff
@@ -0,0 +1,20 @@
...
+.compliancyDetails[0] does not match any elements
+.compliancyDetails does not match the elements
+.compliant: Compliant does not match NonCompliant
+.relatedObjects[0].compliant: NonCompliant does match NonCompliant
+.relatedObjects[0].object.apiVersion: v1 does match v1
+.relatedObjects[0].object.kind: Pod does match Pod
+.relatedObjects[0] is founded
+.relatedObjects is founded
+[1m Expected status does not match the actual status [0m
+
```
JustinKuli, line 15: Why does only the bold part have terminal escape codes? Shouldn't every line that is colored have some?

JustinKuli (changes requested):
Sorry for the many small comments. Overall this looks really great.

I personally like the colors, but I feel like it's common to give a `--no-color` option. Especially if this is a tool used in someone's CI, it can be a headache to look at the output with a bunch of terminal escape codes.

I think the exit code of the CLI is still just whether the result is compliant or not... I think we'd want to adjust that so that if the user provided a desired status, the exit code is zero if the status matches.

pkg/dryrun/dryrun_utils.go
yiraeChristineKim: Let me try I don't think it is big deal

pkg/dryrun/cmd.go
```diff
@@ -60,6 +61,14 @@ func (d *DryRunner) GetCmd() *cobra.Command {
 		"Set to false to omit any diffs generated by the policy.",
 	)
 
+	cmd.Flags().StringVarP(
+		&d.configStatus,
+		"config-status",
```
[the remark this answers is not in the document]
yiraeChristineKim: lovely

pkg/dryrun/dryrun_utils.go
yiraeChristineKim: Sorry, I don’t understand why these are separated—the code would be duplicated. Could you suggest an alternative implementation?

pkg/dryrun/dryrun_utils.go
yiraeChristineKim: you right, len==0 also should be compared

pkg/dryrun/dryrun_utils.go
yiraeChristineKim: Fixed Could you check

pkg/dryrun/dryrun_utils.go
```diff
@@ -0,0 +1,273 @@
...
+	"fmt"
+	"reflect"
+	"sort"
+
+	"github.com/spf13/cobra"
+
+	policyv1 "open-cluster-management.io/config-policy-controller/api/v1"
+)
+
+var ResetColor = "\033[0m"
```
JustinKuli, line 16: nitpick (if this works)
```suggestion
const resetColor = "\033[0m"
```

pkg/dryrun/dryrun_utils.go
JustinKuli: I think that's looking good. What do you think about a `--no-colors` option that would turn off those escapes?

pkg/dryrun/dryrun_utils.go
```diff
@@ -0,0 +1,273 @@
...
+			if !ok {
+				failedTypeMissMatch(cmd, path, reflect.TypeOf(actualValue).String())
+
+				isStatusMatch = false
+
+				continue
+			}
+
+			// Recursive call for nested maps
+			isStatusMatch = compareStatusObj(cmd, isStatusMatch, path, mapObj, mapActualObj)
```
JustinKuli, line 124: Previous discussion: https://github.com/open-cluster-management-io/config-policy-controller/pull/329#discussion_r1941892136

```suggestion
		case map[string]interface{}:
			mapObj := typedV.(map[string]interface{}) // guaranteed? Or maybe unnecessary?

			mapActualObj, ok := convertMapStringKey(actualValue).(map[string]interface{})
			if !ok {
				failedTypeMissMatch(cmd, path, reflect.TypeOf(actualValue).String())

				isStatusMatch = false

				continue
			}

			// Recursive call for nested maps
			isStatusMatch = compareStatusObj(cmd, isStatusMatch, path, mapObj, mapActualObj)

		case map[interface{}]interface{}:
			failedParsePrint(cmd, path)
```

And actually, that new `case map[interface{}]interface{}:` might not be necessary at all, if it is handled well enough in the existing default case.

yiraeChristineKim:
```
case map[string]interface{}:
			mapObj := typedV.(map[string]interface{}) // guaranteed? Or maybe unnecessary?
```

guaranteed so unnecessary

yiraeChristineKim: why unnecessary case is failedParsePrint? The reason  I addded map[interface{}]interface{} case because in testcase, parsing yaml turn to `map[interface{}]interface{}`. so I had to change to `map[string]interface{}` manually

yiraeChristineKim: oh "yaml.UnmarshalStrict()" maybe resolve! the issue?

yiraeChristineKim: Oh my god fixed Thanks Justin I use yaml.v3 then resolve everything!!

PR #330, opened 2025-02-04
Make `inform` default for `remediationAction`

dhaiducek:


JustinKuli (approved):
LGTM! Interesting that it was already required in ConfigurationPolicy

PR #332, opened 2025-02-07
Implement standalone-hub-templating

JustinKuli:
When ConfigurationPolicy or OperatorPolicy is deployed directly to the managed cluster, for example via GitOps, it skips the step in the policy framework that would normally resolve hub templates. This feature allows the config-policy-controller to be configured with a kubeconfig in order to access the hub cluster and resolve those templates. When the new argument to enable this is not set, unresolved hub templates will result in a specific compliance message to report this situation.

Limitations on what the hub templates can access are managed only through the permissions that the given hub kubeconfig has.

To handle situations where the hub cluster can not be accessed (especially temporarily), the policy state after resolving hub templates will be cached on the managed cluster in a secret.

Refs:
 - https://issues.redhat.com/browse/ACM-16091
 - https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-policy/134-standalone-hub-templates


JustinKuli:
/hold

The hosted mode framework test should be fixed by https://github.com/open-cluster-management-io/governance-policy-addon-controller/pull/196

controllers/hubtemplates.go
```diff
@@ -0,0 +1,322 @@
...
+
+		return err
+	}
+
+	resolveOptions := &templates.ResolveOptions{
+		InputIsYAML: true,
+		Watcher:     &objID,
+	}
+
+	if strings.Contains(string(yamlBytes), "ManagedClusterLabels") {
```
yiraeChristineKim, line 87: .ManagedClusterLabels more accurate. There is small possibility that user use this variable such as name

controllers/hubtemplates.go
```diff
@@ -0,0 +1,322 @@
...
+
+		if cacheGetErr != nil && !k8serrors.IsNotFound(cacheGetErr) {
+			// can't clear the cache right now, but this should be retried anyway
+			return err
+		}
+
+		if !k8serrors.IsNotFound(cacheGetErr) {
+			// ignore error; if the delete fails, it will be retried anyway
+			_ = r.getClient().Delete(ctx, cachedResultObject)
+		}
```
yiraeChristineKim, line 135:
```suggestion
if cacheGetErr != nil {
	if !k8serrors.IsNotFound(cacheGetErr) {
		// Cannot clear cache now, but retrying is expected
		return cacheGetErr
	}
} 

_ = r.getClient().Delete(ctx, cachedResultObject)
```

yiraeChristineKim: if error is nil, then execute "_ = r.getClient().Delete(ctx, cachedResultObject)" ??

test/e2e/case43_standalone_hub_templates_test.go
```diff
@@ -0,0 +1,306 @@
...
+			}, defaultTimeoutSeconds, 1).ShouldNot(BeEmpty())
+		})
+
+		It("should have created the cache-secret to cache the hub result", func() {
+			configPolicy := utils.GetWithTimeout(
+				clientManagedDynamic, gvrConfigPolicy, policyName, testNamespace, true, defaultTimeoutSeconds,
+			)
+
+			secretName := string(configPolicy.GetUID()) + "-last-resolved"
+			cacheSecret := utils.GetWithTimeout(
```
yiraeChristineKim, line 154: line 165 in hubtemplates.go
```
if err := r.getClient().Create(ctx, &resolutionToCache); err != nil {
			log.Error(err, "failure to create secret")

			return err
		}
```

I thought we save secrets in hub

JustinKuli: This is saved on the managed cluster, because it's used when the hub cluster can't be accessed.

yiraeChristineKim: ooooh ohoho getclient and hubClient lol Thanks @JustinKuli

yiraeChristineKim (approved):

PR #334, opened 2025-02-13
Fix bug where deprecationsPresent config changes were not reflected

yiraeChristineKim:
Ref: https://issues.redhat.com/browse/ACM-17661
Operator policy should show noncompliant in template details page if there are any deprecation messages


controllers/operatorpolicy_status.go
```diff
@@ -156,6 +138,29 @@ func updateStatus(
 	return condChanged || relObjsChanged || genUpdated
 }
 
+// Update ComplianceCondition and Sort conditions
+func updatedComplianceCondition(policy *policyv1beta1.OperatorPolicy) {
```
yiraeChristineKim, line 142: just refector

JustinKuli, line 142:
```suggestion
func updateComplianceCondition(policy *policyv1beta1.OperatorPolicy) {
```

JustinKuli (changes requested):

JustinKuli (approved):
Looks good. I think when I was first looking through this I was missing a detail and thought that it might be missing a situation, but this looks complete to me now.

(Details on what I was missing:) I see that the situation where the `complianceConfig` is updated is "caught" by the `genUpdated` condition at the end of `updateStatus`. So it will now update the compliance condition, even though `condChanged` is still false, and the controller will know to send the update to the API server because this function returns true.

PR #339, opened 2025-02-20
Trim dryrun status checking output

JustinKuli:
Previously, while looking through the entire status and comparing it to the desired status, the dryrun tool would print whether each individual piece matched or not. In simple cases, this can be helpful, but in more complex cases it becomes hard to read. In some cases when the desired status had two related objects, the output might have a "red" line, but ultimately find that the status matched what was desired, which could be very confusing.

Now, fewer lines are printed in these cases. Note that all the same things are still being checked.


pkg/dryrun/utils_testdata/test_multiple_relobj_compliant/output.txt
JustinKuli, line 1: Under the previous behavior, this test would have a "red" line:
Screenshot from 2025-02-19 21-41-03

Now it's all green (because many of those lines are not printed).

yiraeChristineKim (approved):

PR #340, opened 2025-02-20
Fix mustnothave on unnamed templates with ObjectSelector

JustinKuli:
When a mustnothave policy specifies a named object, then the policy will
always want that resource to be missing from the cluster, regardless of
the other fields in the object template or the actual resource. However,
for unnamed templates, the policy will check to see if other fields
match.

Previously, the way ObjectSelector sets names on templates while
processing them internally was causing mustnothave policies to mark
*all* objects matching the selector as undesired, disregarding any other
fields.

Now, the ObjectSelector and the other fields defined in the template are
both considered when marking objects as undesired by a mustnothave
policy.

Refs:
 - https://issues.redhat.com/browse/ACM-15772

Also includes a small refactor which disambiguates a function.

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -2111,7 +2111,7 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 		// it is a musthave and it does not exist, so it must be created
 		if remediation.IsEnforce() {
 			var uid string
-			completed, reason, msg, uid, err := r.enforceByCreatingOrDeleting(obj)
+			completed, reason, msg, uid, err := r.enforceByCreating(obj)
 
 			hasStatus := false
 			if tmplObj, err := unmarshalFromJSON(objectT.ObjectDefinition.Raw); err == nil {
@@ -2145,7 +2145,7 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 	if exists && !obj.shouldExist {
 		// it is a mustnothave but it exist, so it must be deleted
 		if remediation.IsEnforce() {
-			completed, reason, msg, _, err := r.enforceByCreatingOrDeleting(obj)
+			completed, reason, msg, err := r.enforceByDeleting(obj)
 			if err != nil {
 				objLog.Error(err, "Could not handle existing mustnothave object")
 				result.enforcementErr = err
@@ -2371,12 +2371,10 @@ func (r *ConfigurationPolicyReconciler) getNamesOfKind(
 	return buildNameList(desiredObj, complianceType, resList), allResourceList
 }
 
-// enforceByCreatingOrDeleting can handle the situation where a musthave or mustonlyhave object is
-// completely missing (as opposed to existing, but not matching the desired state), or where a
-// mustnothave object does exist. Eg, it does not handle the case where a targeted update would need
-// to be made to an object.
-func (r *ConfigurationPolicyReconciler) enforceByCreatingOrDeleting(obj singleObject) (
-	result bool, reason string, msg string, uid string, erro error,
+// enforceByCreating handles the situation where a musthave or mustonlyhave object is
+// completely missing (as opposed to existing, but not matching the desired state)
+func (r *ConfigurationPolicyReconciler) enforceByCreating(obj singleObject) (
+	completed bool, reason string, msg string, uid string, err error,
 ) {
 	log := log.WithValues(
 		"object", obj.name,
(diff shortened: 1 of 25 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -2333,27 +2349,31 @@ func (r *ConfigurationPolicyReconciler) getNamesOfKind(
...
+	objectT *policyv1.ObjectTemplate,
 ) (kindNameList []string, allResourceList []string) {
 	var resList *unstructured.UnstructuredList
-	var err error
 
 	ns := desiredObj.GetNamespace()
 
+	sel, err := metav1.LabelSelectorAsSelector(objectT.ObjectSelector)
+	if err != nil {
+		sel = labels.Everything() // in case of an error, just use a selector that matches everything
```
dhaiducek, line 2360: In the unlikely even the objectSelector is broken, should we match nothing? Or am I not looking at this correctly...

controllers/configurationpolicy_controller.go
```diff
@@ -1912,6 +1913,21 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
 	desiredObjName := desiredObj.GetName()
 	desiredObjKind := desiredObj.GetKind()
 
+	// Set an empty name to fix the behavior for an unnamed template in a
+	// mustnothave policy that uses ObjectSelector, so the behavior matches
+	// what happens when ObjectSelector is not used.
```
dhaiducek, line 1918: I probably need to look more broadly at the code, but since `desiredName` is getting wiped:
- Is this running multiple times as it iterates over the list from the objectSelector from the determineDesiredObjects() function? 
- Is it compensating for when `skipObject` is called?
- As a result it seems this is no longer enforceable?

dhaiducek (commented):
The refactor is a nice, small change that definitely clarifies each purpose! I have a couple questions about it.

controllers/configurationpolicy_controller.go
JustinKuli: It might be running multiple times, but from the status and compliance message, everything seemed to be combined correctly, so I don't think there is duplicated information that would confuse the user. I don't think it's so inefficient that it needs to be addressed immediately.

`skipObject` is handled by `determineDesiredObjects`, before `handleObjects` where these changes are, so I think it's handled correctly. What did you mean by "compensating"?

That is true: a `mustnothave` policy targeting multiple objects via objectSelector and not specifying a name in the template will not delete matches objects when enforced. This matches the existing behavior for mustnothave policies without a name specified, which are just marked as noncompliant and do not delete any resources.

controllers/configurationpolicy_controller.go
```diff
@@ -1912,6 +1913,21 @@ func (r *ConfigurationPolicyReconciler) handleObjects(
...
+	if objectT.ComplianceType.IsMustNotHave() && objectT.ObjectSelector != nil {
+		parsedMinMetadata := minimumMetadata{}
+
+		// This error should be impossible because of CRD validation, but just
+		// skip possibly changing the name if an error occurs.
+		if err := json.Unmarshal(objectT.ObjectDefinition.Raw, &parsedMinMetadata); err == nil {
```
yiraeChristineKim, line 1924: When error != nil, handle it and follow the comment: 'Skip possibly changing the name if an error occurs.'

controllers/configurationpolicy_controller.go
yiraeChristineKim: I think matching nothing makes sense too me.  Can you elaborate more this?

controllers/configurationpolicy_controller.go
```diff
@@ -2393,62 +2411,79 @@ func (r *ConfigurationPolicyReconciler) enforceByCreatingOrDeleting(obj singleOb
...
+	var res dynamic.ResourceInterface
+	if obj.scopedGVR.Namespaced {
+		res = r.TargetK8sDynamicClient.Resource(obj.scopedGVR.GroupVersionResource).Namespace(obj.namespace)
+	} else {
+		res = r.TargetK8sDynamicClient.Resource(obj.scopedGVR.GroupVersionResource)
+	}
+
+	log.Info("Enforcing the policy by deleting the object")
+
+	if completed, err = deleteObject(res, obj.name, obj.namespace); !completed {
```
yiraeChristineKim, line 2479:
```suggestion
	if completed, err = deleteObject(res, obj.name, obj.namespace); !completed || err != nil {
```

yiraeChristineKim:
Good Good! dryrun test is very easy to understand as a reviewer and intuitive

controllers/configurationpolicy_controller.go
JustinKuli: `err` might be non-nil if the object was already deleted. But in this case, we can just consider it a successful deletion. Other than that scenario, `completed` is equivalent to `err == nil`

controllers/configurationpolicy_controller.go
JustinKuli: That's a good point. Actually, if this selector is broken, it would already be handled in `determineDesiredObjects` with an error, and it shouldn't get to this section in `handleObjects`. I'll make this select nothing, and adjust the comment to reflect this.

controllers/configurationpolicy_controller.go
JustinKuli: This error should already have been handled in `determineDesiredObjects`, and wouldn't get here in `handleObjects`, so I'll update the comment here.

controllers/configurationpolicy_controller.go
yiraeChristineKim: `mustnothave` + `enforce` + `no name` + `objectSelector`. doesn't delete any resources?

JustinKuli: Correct. This matches previous behavior where `mustnothave` + `enforce` + `no name` doesn't delete any resources.

dhaiducek: >This matches the existing behavior for mustnothave

So, this is the bug. Where, when specifying the `objectSelector`, if an `inform` policy is NonCompliant, when enforced it should be deleted. The `objectSelector` should still allow enforcement, but the bug is that it should be consistent with the `inform` status (i.e. Compliant --> Don't delete; NonCompliant --> Delete).

JustinKuli: > The objectSelector should still allow enforcement

I think this was a requirement I had missed, so I was just trying to make the behavior here consistent with `mustnothave` + `enforce` + `no name`.

I think I can do what @dhaiducek is talking about... It will take some time to adjust some things (and write an `enforce` test)

JustinKuli: Ok, updated! I'm going to resolve this conversation because it got long, but we can start a new one on the updated changes.

controllers/configurationpolicy_controller.go
```diff
@@ -1709,6 +1710,29 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
...
+		targetedObjects := make([]*unstructured.Unstructured, 0, len(desiredObjects))
+
+		for _, d := range desiredObjects {
+			unnamedObj := d.DeepCopy()
+			unnamedObj.SetName("")
+
+			matchingNames, _ := r.getMatchingNames(plc, unnamedObj, scopedGVR, objectT)
+
+			for _, n := range matchingNames {
+				if n == d.GetName() {
```
dhaiducek, line 1725: (Optional) Minor optimization: `d.GetName()` could be declared as a variable outside this loop. (Though it'd also make it busier, so maybe it's better leave it as-is. 🙂 )

controllers/configurationpolicy_controller.go
```diff
@@ -1709,6 +1710,29 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
 		return nil, &scopedGVR, event, nil
 	}
 
+	// For mustnothave with no name and with an object selector, filter the "desired" objects
+	// so that only ones matching the other details in the template are included.
+	if objectT.ComplianceType.IsMustNotHave() && parsedMinMetadata.Metadata.Name == "" && objectSelector != nil {
```
dhaiducek, line 1715: Oh, interesting. So this is just removing the object-templates that don't match. Very creative! 🤔 💡 

Some thoughts:
1. This comment _might_ be an edge case/technical debt that I noticed in `buildNameList()`:
   ```
   // Dry run API requests aren't run on unnamed object templates for performance reasons, so be less
   // conservative in the comparison algorithm.
   ```

2. Using `parsedMinMetadata.Metadata.Name` is tricky because it _could_ have a templated value, even as simple as `{{ .ObjectName }}`, in which case I think it wouldn't proceed through this block? Would it work to do just this?
   ```
   if objectT.ComplianceType.IsMustNotHave() {
   ```
   Then the behavior would be consistent for all cases, regardless of the contents of `metadata.name`? Either that or I guess check for empty or templated?

dhaiducek (commented):
Left a comment on the conditional about some edge cases, but I think it's looking good!

controllers/configurationpolicy_controller.go
JustinKuli: What would be the possible problem with (1)?

I was interpreting any templated value in the name of a mustnothave policy as meaning "please remove the objects named this". So a mustnothave policy with an object selector and `name: {{ .ObjectName }}` would try to delete all resources matching the selector.

If you use objectSelector *and* set a name in the template, the selector is just ignored, right? Unless the name is templated to use `.ObjectName`

JustinKuli: Do you mean that if a mustnothave policy uses a named template, but some of the spec in the policy doesn't match the actual object on the cluster, then the object should *not* be deleted?

An example might help clarify. Say I have this policy:
```yaml
apiVersion: policy.open-cluster-management.io/v1
kind: ConfigurationPolicy
metadata:
  name: test-policy-1
spec:
  remediationAction: enforce
  object-templates:
    - complianceType: mustnothave
      objectDefinition:
        apiVersion: networking.k8s.io/v1
        kind: Ingress
        metadata:
          namespace: default
          name: wrong-3-ingress
        spec:
          ingressClassName: unknown
```

And this Ingress on the cluster:
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  namespace: default
  name: wrong-3-ingress
spec:
  ingressClassName: wrong-name
```

The existing behavior has been that this Ingress would be deleted. If the new logic applied to all mustnothave policies (specifically those with named object templates), then this behavior would change: it would not delete that Ingress.

dhaiducek: Yes, that's what I'm saying. I personally think that makes more sense and would provide a more nuanced deletion. If you want the old behavior, just include metadata and nothing else.

dhaiducek: With `objectSelector`, we're introducing an intersection between the named `objectDefinition` and an unnamed one. I think the behavior should be the same.

JustinKuli: I agree that the new behavior would make more sense in some scenarios. But it is a change in existing behavior, which I would generally like to avoid.

Also, I can imagine scenarios where users are relying on the current behavior. Maybe they use an enforced musthave policy to set some objects up initially on a cluster. Then they change the policy to inform, and make some updates manually to that object. Eventually they want to clean up these resources, so they change the policy to mustnothave, and enforce it. (This approach has some benefits over `pruneObjectBehavior`: it doesn't use finalizers, and the user can watch the policy become Compliant to be sure that it has deleted the objects.) Under the existing behavior, the policy doesn't care if some fields have been changed: the template has a name, so that object will be deleted.

We've had e2e tests cleaning up things like this before, so it wouldn't be a huge surprise if any users followed that pattern. Changing the objects away from what was originally defined in the policy is a little bit odd, but it seems reasonable to me.

dhaiducek: Hmmm. Those are interesting points. So the `objectSelector` provides a consistent experience with unnamed objects, but also provides a new, more nuanced way of deleting objects in addition to the current way of obliterating objects directly by name. And if you are using the object selector and want the old way, just put a name/template in there.

In that case we'll need to clearly document this, but I think it makes sense. (Like if you want the new behaviors, don't put `skipObject` logic in the `metadata.name` field.)

For the sake of this discussion, could you also add a test case that has a templated name?

JustinKuli: Test added: https://github.com/open-cluster-management-io/config-policy-controller/compare/25c86052046b5ed404df5cc1d7d76a553dab260b..ace11e9e708571fa3a197ee2ce8a3e00cf9b4079

dhaiducek (approved):
/unhold

LGTM! Thanks for the discussion around it! Had to take a minute to wrap my mind around all those test cases too--those are thorough! 😄

PR #342, opened 2025-02-26
Fix dryrun default namespaceSelector issue

yiraeChristineKim:
- Add the "default" namespace if the input resources do not have a namespace.
- Add "default" namespace in dryrun 
Ref: https://issues.redhat.com/browse/ACM-18064


JustinKuli (approved):
LGTM! I was worried for a moment when I saw the default namespace was removed from several tests, but it looks like it's not a problem if users do supply the default namespace in their input resources.

yiraeChristineKim:
> LGTM! I was worried for a moment when I saw the default namespace was removed from several tests, but it looks like it's not a problem if users do supply the default namespace in their input resources.

Yes, If they add default namespace (already have), I added skip "alreadyExist error" so Should be fine! Thanks @JustinKuli  for double check

PR #346, opened 2025-03-10
Use lowercase APIMapping

dhaiducek:
Proposed followup to:
- #345


pkg/mappings/mappings.go
```diff
@@ -140,7 +140,7 @@ func GenerateMappings(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	cmd.Println(string(out))
```
yiraeChristineKim, line 143: Why this changed?

yiraeChristineKim:
Let me create a Jira ticket to add a unit test for 'generate'.

JustinKuli:
I'm hesitant to change the mapping file format, because I feel like it did work at some point? I remember manually "hacking" in some simple fake kind at some point, and I think Matt did something similar... The lack of a test makes it hard to tell if/when things broke.

But this is still a small piece of preview feature. If changing it brings other benefits, we should do it.

What does this give us over the current fix?

pkg/mappings/mappings.go
dhaiducek: It was adding a newline where one already existed.

dhaiducek:
> What does this give us over the current fix?

Just tried it out. It seems if lowercase keys are used, that works before the fix. So it's an inflection point. Do we go with what `dryrun generate` has always output, or do we go with what was working before this?

yiraeChristineKim:
Should we add unit test with capital?

dhaiducek:
> Should we add unit test with capital?

There is one--your test is unchanged 🙂

JustinKuli (approved):

yiraeChristineKim (approved):

PR #350, opened 2025-03-24
Watch correct openshift templates

JustinKuli:
Chiefly, this updates the kubernetes-dependency-watches library to use verbs allowed on the API resources to only choose ones that can be watched - which fixes the ambiguous case around Openshift Templates, which was causing the config-policy-controller to use the wrong endpoint.

This also updates the mappings file format to include verbs, so that dryrun can match this behavior.

https://issues.redhat.com/browse/ACM-18827

### DIFF
--- a/pkg/mappings/mappings.go
+++ b/pkg/mappings/mappings.go
@@ -19,12 +19,18 @@ import (
 // APIMapping stores information required for mapping between GroupVersionKinds
 // and GroupVersionResources, as well as whether the API is namespaced.
 type APIMapping struct {
-	Group    string `json:"group"`
-	Version  string `json:"version"`
-	Kind     string `json:"kind"`
-	Singular string `json:"singular"`
-	Plural   string `json:"plural"`
-	Scope    Scope  `json:"scope"`
+	Group    string   `json:"group"`
+	Version  string   `json:"version"`
+	Kind     string   `json:"kind"`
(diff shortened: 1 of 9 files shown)


JustinKuli:
This initial draft uses a manual replacement to my forked/fixed version of the library.

pkg/mappings/mappings.go
```diff
@@ -116,13 +132,22 @@ func GenerateMappings(cmd *cobra.Command, _ []string) error {
 				scoping = meta.RESTScopeNameNamespace
 			}
 
+			slices.Sort(res.Verbs)
+
+			verbs := res.Verbs
+
+			if res.Verbs.String() == fmt.Sprintf("%v", DefaultVerbs) {
```
dhaiducek, line 139: Parsing to a string isn't necessary since they're both sorted.

```suggestion
			if slices.Equal(res.Verbs, defaultVerbs) {
```

dhaiducek (approved):
/hold

My conversation there with myself is nitpicky, so approving and holding in case you want to revisit it. (Also, FYI: the E2E here failed.)

pkg/mappings/mappings.go
JustinKuli: Good point! This was a remnant from before I realized sorting it would be very useful.

dhaiducek (approved):
LGTM!

PR #355, opened 2025-04-07
Refactor `skipObject` to report arg types

dhaiducek:
This also shortens the message because it appears the templating library already wraps our error with "error calling skipObject", so saying "skipObject" again in the message isn't necessary.

Followup to:
- #354


JustinKuli (approved):
LGTM

PR #357, opened 2025-04-11
Colored diffs are displayed as output.

yiraeChristineKim:
During a dry run test, colored diffs are displayed as output.
Ref: https://issues.redhat.com/browse/ACM-18908


pkg/dryrun/dryrun_utils.go
```diff
@@ -317,3 +318,22 @@ func PathExists(path string) bool {
 
 	return err == nil
 }
+
+func addColorToDiff(bareDiff string, noColor bool) string {
+	lines := strings.Split(bareDiff, "\n")
+	for i, line := range lines {
+		if strings.HasPrefix(line, "-") && !strings.HasPrefix(line, "---") {
```
dhaiducek, line 325: (Optional) I'd personally be fine if the header were also colored...

```suggestion
		if strings.HasPrefix(line, "-") {
```

yiraeChristineKim: Screenshot 2025-04-11 at 12 24 33 PM
git diff doesn't add color to "---" so. If we add the color? same color red and green?

dhaiducek: Oh, we can leave it then--I didn't recall that git handled it this way also. 🙂

JustinKuli: I like it as-is

JustinKuli (approved):

PR #358, opened 2025-04-14
Add a flag to show the complete diff during dry run

yiraeChristineKim:
https://issues.redhat.com/browse/ACM-18907

### DIFF
--- a/pkg/dryrun/cmd.go
+++ b/pkg/dryrun/cmd.go
@@ -20,6 +20,7 @@ type DryRunner struct {
 	mappingsPath  string
 	logPath       string
 	noColors      bool
+	fullDiffs     bool
 }
 
 var ErrNonCompliant = errors.New("policy is NonCompliant")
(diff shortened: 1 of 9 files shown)


pkg/dryrun/cmd.go
```diff
@@ -77,6 +78,13 @@ func (d *DryRunner) GetCmd() *cobra.Command {
 		"Disables colored output. By default, output is printed with colors.",
 	)
 
+	cmd.Flags().BoolVar(
+		&d.fullDiffs,
+		"full-diffs",
+		false,
+		"Enable this to display the full diff in the output. By default, the maximum number of diff lines shown is 53.",
```
JustinKuli, line 85: It's 50 lines of diff, plus a 3 line header... So I'd rather we say 50 lines, or just something like "By default, the number of lines are limited for readability" (and not put a specific number)

JustinKuli (changes requested):
Looks good overall!

JustinKuli (approved):

PR #366, opened 2025-05-28
Always do a server-side dry-run check to know if resource updates are necessary

JustinKuli:
See https://issues.redhat.com/browse/ACM-19156 (or just the test case) for the situation that prompted this. Basically:
```
Previously, a server-side dry-run was only done after `handleKeys`
indicated that an update was necessary, to deal with cases where fields
were not returned by the api server and an update wasn't actually
needed.

Now, the server-side dry-run can also catch situations where
`handleKeys` incorrectly indicates an update is not necessary: in
particular it was found that some "empty" fields in the desired state
might need to be set on the object.
```

The commits are split up to help in review: removing some now-unused code and pre-factoring some conditions helped make the actual change more straightforward.


dhaiducek (commented):
Looks like my diff PR caused a conflict--sorry about that!

Overall looks good! I have some comments about updating function descriptions but realized late the `missingKeys` is propagated pretty deep. Somewhat tedious, but I think it'd be good to add everywhere for completeness, but I'll leave it up to your discretion how many places it needs to be. 🙂

test/e2e/case12_list_compare_test.go
```diff
@@ -96,6 +95,7 @@ var _ = Describe("Test list handling for musthave", func() {
 			}
 
 			deleteConfigPolicies(policies)
+			utils.KubectlDelete("role", "--namespace=default", "role-policy-e2e")
```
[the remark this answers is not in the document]
JustinKuli: It probably could, but other tests in this file manually clean things up. I'm not particularly interested in changing all of them :sweat_smile:

controllers/operatorpolicy_controller.go
```diff
@@ -2790,7 +2790,8 @@ func (r *OperatorPolicyReconciler) mergeObjects(
 	existingObjectCopy := existing.DeepCopy()
 	removeFieldsForComparison(existingObjectCopy)
 
-	_, errMsg, updateNeeded, _ := handleKeys(desiredObj, existing, existingObjectCopy, complianceType, "")
+	//nolint:dogsled
+	_, errMsg, updateNeeded, _, _ := handleKeys(desiredObj, existing, existingObjectCopy, complianceType, "")
```
yiraeChristineKim, line 2794: Do we have a unit test for missingkey?

JustinKuli: A unit test for this piece specifically would be nice... right now we only have the e2e tests which verify it works as expected.

I have a bit of a wishlist forming for refactors here, but I don't want all of them to clutter up this PR...

controllers/configurationpolicy_utils.go
```diff
@@ -146,57 +146,68 @@ func addOrUpdateRelatedObject(
...
-// equalObjWithSort is a wrapper function that calls the correct function to check equality depending on what
-// type the objects to compare are
-func equalObjWithSort(mergedObj interface{}, oldObj interface{}, zeroValueEqualsNil bool) (areEqual bool) {
+// deeplyEquivalent deeply compares the first two inputs, considering them
+// equivalent even if they have lists which are in different orders. When the
+// `zeroValueEqualsNil` parameter is true, it will allow nested maps to have
+// omitted fields that are "zero values" to be considered equivalent.
+//
+// It returns whether the first two inputs are equivalent, and if any of the
+// nested maps were considered equivalent dues to "zero values".
```
dhaiducek, line 155: nit: grammar
```suggestion
// nested maps were considered equivalent due to "zero values".
```

dhaiducek (dismissed):
/hold

I think this looks good. Evaluating ConfigPolicy flows is hard, but in the end the tests passed. 😄 I've held it because I have some minor questions about grammar and adding `recordDiff: log` in manifests that appear unrelated to the PR.

controllers/configurationpolicy_controller.go
```diff
@@ -2811,30 +2804,37 @@ func mergeSpecsHelper(
...
-// different in the template.
+// mergeArrays performs a deep merge operation, combining the given lists. It
+// specially considers items with "names", which it will merge from each list.
+// It preserves duplicates that are already in `existingArr`, but does not
+// introduce *new* duplicate items if they are in `desiredArr`. When merging
+// items or nested items which are maps, the `zeroValueEqualsNil` parameter
+// determines how to handle certain "zero value" cases (see `deeplyEquivalent`).
+//
+// It returns the merged list, and indicates whether any of the nested maps were
+// considered equivalent dues to "zero values".
```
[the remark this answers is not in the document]
JustinKuli: Screenshot From 2025-06-12 09-08-53

dhaiducek (approved):
LGTM!

PR #368, opened 2025-05-29
fix: address `.Object` bugs

dhaiducek:
- Expand types allowed in `.Object`
  - Originally, the Go template library limited the allowed types to structs and string maps. This pulls in an update that expands the allowed types.
- Fetch objects with `namespaceSelector`
  - Objects were not being fetched when the `namespaceSelector` was in use. This adds the fetch and also adds a log and modifies the compliance message to aid in identifying the object that caused the problem in case it was specific to the object being parsed instead of the template syntax.


controllers/configurationpolicy_controller.go
```diff
@@ -1387,9 +1390,33 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
...
+		// - Object var is used
+		// - Name is provided and not templated
+		// - Namespace is empty and not templated
+		// (Fetching for the object selector is handled later on)
+		fetchObject := needsObject && desiredName != "" && parsedMinMetadata.Metadata.Namespace == ""
+
 		if len(selectedNamespaces) != 0 {
 			for _, ns := range selectedNamespaces {
-				relevantNsNames[ns] = defaultNamesPerNs
+				switch {
```
JustinKuli, line 1402: Is there a linter that was complaining about an `if` statement here? Using a `switch` seems odd for a boolean.

test/dryrun/context_vars/object_pod_nsselector/error.txt
```diff
@@ -0,0 +1,7 @@
+Error:
+unable to complete the dryrun reconcile:
+failed to resolve the template {"apiVersion":"v1","kind":"Pod","metadata":{"labels":{"image":"{{ index .Object.spec.containers 0 \"image\" }}"},"name":"nginx-pod"}}:
+template:
+tmpl:5:28:
+executing "tmpl" at <.Object.spec.containers>:
+can't evaluate field Object in type struct { ObjectNamespace string; ObjectName string }
```
JustinKuli, line 7: I suppose it makes sense to have an error if you use `.Object` in a template when the resource doesn't exist yet... but it also feels like a bit of a major limitation. Is there a way for the user to provide a default value when the resource doesn't exist yet?

dhaiducek: I've updated the logic to populate `.Object` in all cases except for unnamed `objectDefinition`s. Please take another look when you have a minute!

JustinKuli (approved):
LGTM overall!

PR #369, opened 2025-06-04
fix: Record the diff for enforce

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-19111

Investigating this, the console (to be confirmed) appears to have conditionals around only displaying the diff link with `remediationAction: inform` and on violations. It seems we can persist the diff in that case. When flipping from `enforce` to `inform`, the diff is cleared if the objects match. I think this flow makes sense (persist for `enforce` to show the historical diff but don't persist for `inform`).


dhaiducek:
It's a relatively small number of line changes, but I'm not 100% certain the change should be backported?

yiraeChristineKim (approved):

JustinKuli (approved):

PR #370, opened 2025-06-05
Add helpful errors for unwatchable resources

JustinKuli:
Some resources like PackageManifests from OLM do not support the `watch` API, and would previously cause various errors in policies. Now, we handle these errors specially, and can advise the user to set the `evaluationInterval` on the policy in order for the lookups to work properly and as expected.

Refs:
 - https://issues.redhat.com/browse/ACM-19965


dhaiducek (approved):
LGTM!

PR #371, opened 2025-06-11
Use JSON instead of YAML for hub templates

JustinKuli:
For some reason, linebreaks in the YAML encoding can break the `copySecretData` template function.

Refs:
 - https://issues.redhat.com/browse/ACM-21394


dhaiducek (approved):
LGTM!

PR #373, opened 2025-07-01
fix: Allow templated namespace without nsSelector

dhaiducek:
When a `namespaceSelector` was not provided and a namespace was templated, the controller considered this invalid and didn't attempt to resolve the template, returning errors about a missing namespace.

ref: https://issues.redhat.com/browse/ACM-21804


JustinKuli (approved):
LGTM!

PR #377, opened 2025-07-17
Add RelatedObject property when compliant dry run overrides policy mismatch

jan-law:
When configuration policies detect desired vs actual object mismatches, the dry run update validation sometimes responds with a "no-op", aka "no changes required" due to apiserver webhooks or mutating admission controllers modifying requests. The policy is assumed to be compliant without indicating any potential mismatch to users. 

Since the requests are modified at the apiserver, we do not have other means of validating the policy before making the dry run request. To improve user experience when troubleshooting this scenario, this PR adds the property ~~`DryRunNoOpOverride`~~ (edit: `MatchesAfterDryRun`) when the policy detects a mismatch AND the no-op dry run update overrode the policy assessment.

The `RelatedObject` `Properties` object will appear for policies in any remediation mode. Previously, all properties were hidden for all compliant objects in `Inform` mode.

ref: https://issues.redhat.com/browse/ACM-14577

Screenshot From 2025-07-28 14-01-02

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -3575,6 +3575,10 @@ func removeFieldsForComparison(obj *unstructured.Unstructured) {
 	)
 	// The generation might actually bump but the API output might be the same.
 	unstructured.RemoveNestedField(obj.Object, "metadata", "generation")
+
+	if len(obj.GetAnnotations()) == 0 {
+		unstructured.RemoveNestedField(obj.Object, "metadata", "annotations")
+	}
 }
 
 // setEvaluatedObject updates the cache to indicate that the ConfigurationPolicy has evaluated this
--- a/test/resources/case8_status_check/case8_service.yaml
+++ b/test/resources/case8_status_check/case8_service.yaml
@@ -3,8 +3,6 @@ kind: Service
 metadata:
   name: grc-policy-propagator-metrics
   namespace: managed
-  annotations:
-    test: test
 spec:
   internalTrafficPolicy: Cluster
   ipFamilies:


yiraeChristineKim:
What is no-op mean?

test/e2e/case20_delete_objects_test.go
```diff
@@ -71,6 +71,18 @@ var _ = Describe("Test Object deletion", Ordered, func() {
...
+				managedPlc := utils.GetWithTimeout(clientManagedDynamic, gvrConfigPolicy,
+					case20ConfigPolicyNameCreate, testNamespace, true, defaultTimeoutSeconds)
+				relatedObj := managedPlc.Object["status"].(map[string]interface{})["relatedObjects"].([]interface{})[0]
+				properties := relatedObj.(map[string]interface{})["properties"].(map[string]interface{})
+
+				if dryRunNoOpOverride, exists := properties["dryRunNoOpOverride"]; exists {
+					return dryRunNoOpOverride.(bool)
+				}
+
+				return false
```
yiraeChristineKim, line 84: Is it ok dryRunNoOpOverride property missed?

yiraeChristineKim:
Are you going to add e2e test?

controllers/configurationpolicy_controller.go
```diff
@@ -2393,8 +2393,9 @@ func (r *ConfigurationPolicyReconciler) handleSingleObj(
 
 		created := false
 		uid := string(obj.existingObj.GetUID())
+		dryRunNoOpOverride := false
 
-		if evaluated, compliant, cachedMsg := r.alreadyEvaluated(obj.policy, obj.existingObj); evaluated {
+		if evaluated, compliant, cachedMsg, dryRunNoOp := r.alreadyEvaluated(obj.policy, obj.existingObj); evaluated {
```
JustinKuli, line 2398: Is it necessary to store `dryRunNoOp` in the cache of already evaluated things? It seems like it could be pulled from the status, like `Properties.Diff` is.

test/e2e/case20_delete_objects_test.go
```diff
@@ -142,14 +178,34 @@ var _ = Describe("Test Object deletion", Ordered, func() {
...
+				properties := relatedObj.(map[string]interface{})["properties"].(map[string]interface{})
 
-				return properties
-			}, defaultTimeoutSeconds, 1).Should(BeNil())
+				if dryRunNoOpOverride, exists := properties["dryRunNoOpOverride"]; exists {
+					return dryRunNoOpOverride.(bool)
+				}
+
+				return false
+			}, defaultTimeoutSeconds, 1).Should(BeFalse())
```
JustinKuli, line 208: Is there a test that checks when this new status field is supposed to be `true`?

I guess it might be hard to find a field that triggers this for a test in KinD: the example in the bug is openshift-specific.

In general, I think these `delete_objects_test` scenarios aren't exactly the right spot for verifying this field.

JustinKuli (changes requested):

controllers/configurationpolicy_controller.go
jan-law: Good point, I can remove it from the cache

controllers/configurationpolicy_controller.go
```diff
@@ -3227,7 +3229,7 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
 			diff = handleDiff(log, recordDiff, existingObjectCopy, mergedObjCopy, r.FullDiffs)
 		}
 
-		r.setEvaluatedObject(obj.policy, obj.existingObj, !throwSpecViolation, "")
+		r.setEvaluatedObject(obj.policy, obj.existingObj, !throwSpecViolation, "", false)
```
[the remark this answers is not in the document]
jan-law: `throwSpecViolation` is initially true when `handleKeys()` notices a difference between the policy fields and the actual object fields.

Afterwards, the dry run validation handles edge cases where `updateNeeded` found a difference between unset and empty fields. If the dry run Update did not update any fields, the policy is assumed to be compliant, and `throwSpecViolation` is inverted.

https://github.com/open-cluster-management-io/config-policy-controller/blob/b390ed7f20c7af78c017716d805bd2d031483305/controllers/configurationpolicy_controller.go#L3252 

In this case ACM-14577, the bad scc annotation in the policy is technically invalid, but the dry run validation can’t detect the error. During the dry run update request, an admission controller intercepts the request body and silently fixes the invalid annotation before the api server returns a response. This means the dry run update thinks the object matches the policy.

test/e2e/case20_delete_objects_test.go
jan-law: Let me see if Kind has any default admission controllers to reproduce this behaviour

controllers/configurationpolicy_controller.go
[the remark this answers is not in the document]
yiraeChristineKim: Also can you add comment

test/e2e/case20_delete_objects_test.go
yiraeChristineKim: if properties["dryRunNoOpOverride"] not exist.. the test is passed? I think we should throw error

jan-law: Sounds good, I'll give the field a default value

controllers/configurationpolicy_controller.go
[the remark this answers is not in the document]
yiraeChristineKim: okie Can you add comment for this

yiraeChristineKim:
Amazing, @JustinKuli! Let’s try it — my only idea was Gatekeeper, but yours is much better!

jan-law:
Setting to draft while running tests. Will re-open when ready

controllers/configurationpolicy_controller.go
```diff
@@ -3323,9 +3324,12 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
...
+			// Assume object is compliant by inverting the value of throwSpecViolation
 			r.setEvaluatedObject(obj.policy, obj.existingObj, !throwSpecViolation, "")
 
-			return throwSpecViolation, "", diff, updateNeeded, updatedObj
+			matchesAfterDryRun = false
+
+			return throwSpecViolation, "", diff, updateNeeded, updatedObj, matchesAfterDryRun
 		}
 
 		diff = handleDiff(log, recordDiff, existingObjectCopy, dryRunUpdatedObj, r.FullDiffs)
```
JustinKuli, line 3335: I feel like we only want to write `properties.matchesAfterDryRun` in some relatively specific cases. So usually we want this function to return false with that, and then the field will be omitted from the status (because false is the "empty" state for booleans).

So I think the `else` case here is when we're in those "relatively specific cases", so this might work:

```suggestion
			return throwSpecViolation, "", diff, updateNeeded, updatedObj, false
		}
		
		matchesAfterDryRun = true

		diff = handleDiff(log, recordDiff, existingObjectCopy, dryRunUpdatedObj, r.FullDiffs)
```

And then all returns lower than this can use
```
		return ..., matchesAfterDryRun
```

JustinKuli (changes requested):
Excellent tests, and nice work adding that field! I'm suggesting changing many of the `return` cases in `checkAndUpdateResource` to default to "false" for the new field, but maybe I'm missing a reason for it to be true?

test/resources/case8_status_check/case8_service_with_type.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: grc-policy-propagator-metrics
+  namespace: managed
+  annotations:
+    test: test
```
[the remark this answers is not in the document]
JustinKuli: Thanks to your test, I was able to narrow down what seems to be happening. When the service annotations are empty, the `reflect.DeepEqual(dryRunUpdatedObj.Object, existingObjectCopy.Object)` is coming back false because one of the objects has an empty annotations map, and the other has *no* annotations map. I don't exactly know why they are different, and especially why the same situation is *not* happening with the empty labels.

But adjusting this existing function seems like it could handle it (I've added the last 3 lines):
```
func removeFieldsForComparison(obj *unstructured.Unstructured) {
	unstructured.RemoveNestedField(obj.Object, "metadata", "managedFields")
	unstructured.RemoveNestedField(
		obj.Object, "metadata", "annotations", "kubectl.kubernetes.io/last-applied-configuration",
	)
	// The generation might actually bump but the API output might be the same.
	unstructured.RemoveNestedField(obj.Object, "metadata", "generation")

	if len(obj.GetAnnotations()) == 0 {
		unstructured.RemoveNestedField(obj.Object, "metadata", "annotations")
	}
}
```

controllers/configurationpolicy_controller.go
```diff
@@ -3188,7 +3189,7 @@ func (r *ConfigurationPolicyReconciler) checkAndUpdateResource(
 	if obj.existingObj == nil {
 		log.Info("Skipping update: Previous object retrieval from the API server failed")
 
-		return false, "", "", false, nil
+		return false, "", "", false, nil, true
```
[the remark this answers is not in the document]
jan-law: Sounds good, `true` was picked arbitrarily for those cases. Setting to false

controllers/configurationpolicy_controller.go
jan-law: Originally I was thinking about the bool from the perspective of the policy, as in "return true if the dry run matches the policy assessment". Since the property is in the RelatedObject properties, "return true if the object matches after the dry run update" makes more sense.

JustinKuli (approved):
LGTM! Thanks for the diligent work here!

PR #379, opened 2025-07-28
fix: Properly handle `objectSelector` with `namespaceSelector`

dhaiducek:
The `relevantNsNames` variable is a map to a map. However, in Golang a map is a pointer, so setting one `relevantNsNames[ns]` resulted in setting all of the namespaces to the same value. This instead instantiates a new value each time.

Followup to
- https://github.com/open-cluster-management-io/config-policy-controller/pull/361

ref: https://issues.redhat.com/browse/ACM-22676

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -1391,13 +1391,16 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
 
 	// Set up default name array to be used for each namespace. If no
 	// objectSelector is provided, add the desired name as the default.
-	defaultNamesPerNs := map[string]unstructured.Unstructured{}
 	objectSelector := objectT.ObjectSelector
 
-	if desiredName != "" || objectSelector == nil {
-		defaultNamesPerNs = map[string]unstructured.Unstructured{
-			desiredName: {},
+	getDefaultNamesPerNs := func() map[string]unstructured.Unstructured {
+		if desiredName != "" || objectSelector == nil {
+			return map[string]unstructured.Unstructured{
+				desiredName: {},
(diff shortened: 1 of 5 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -1391,13 +1391,16 @@ func (r *ConfigurationPolicyReconciler) determineDesiredObjects(
 
 	// Set up default name array to be used for each namespace. If no
 	// objectSelector is provided, add the desired name as the default.
-	defaultNamesPerNs := map[string]unstructured.Unstructured{}
 	objectSelector := objectT.ObjectSelector
 
-	if desiredName != "" || objectSelector == nil {
-		defaultNamesPerNs = map[string]unstructured.Unstructured{
-			desiredName: {},
+	getDefaultNamesPerNs := func() map[string]unstructured.Unstructured {
```
JustinKuli, line 1396: That's a clean way to do it!

test/dryrun/context_vars/objectname_nsselector/policy.yaml
```diff
@@ -0,0 +1,18 @@
...
+  object-templates:
+    - complianceType: musthave
+      recordDiff: InStatus
+      objectSelector: {}
+      objectDefinition:
+        apiVersion: v1
+        kind: ConfigMap
+        metadata:
+          name: '{{ not (hasPrefix "inv" .ObjectName) | skipObject }}'
+        data: '{{ set .Object.data "hocus" "pocus" | toJSON | toLiteral }}'
```
JustinKuli, line 18: Hocus Pocus

JustinKuli (approved):

PR #380, opened 2025-07-29
Implement a per-item rate-limiter

JustinKuli:
ConfigurationPolicy already had something for this, under the "evaluation-backoff" flag. However, this was implemented *inside* the reconcile, and could often be overriden by other logic. The new approach can more easily be implemented on multiple controllers.

This also tweaks (and explains) the default RateLimiters that were being used.

Refs:
 - https://issues.redhat.com/browse/ACM-22682

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -8,7 +8,6 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
-	"math"
 	"reflect"
 	"regexp"
 	"sort"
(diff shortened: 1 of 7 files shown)


controllers/configurationpolicy_controller.go
```diff
@@ -8,7 +8,6 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
-	"math"
```
dhaiducek: That's right, we don't need you...MATH 😂

PR #381, opened 2025-07-31
Remove unused serverVersion field

JustinKuli:
This field is no longer used, but could cause a panic if the target kubeconfig was removed and the container restarted (which could happen in a hosted mode scenario).

Refs:
 - https://issues.redhat.com/browse/ACM-22679


dhaiducek (approved):
LGTM!

Strange that we set it and then don't use it for anything 🤔 Must be a pretty old field? 😅

dhaiducek:
Oh I see it now. It was used to detect whether dry run is supported, but it's supported for everything now and that check was removed but not this one. 🙂

PR #382, opened 2025-08-01
Add repo-level architecture diagram

jan-law:
Looking for a faster workflow to generate and update team architecture diagrams. Inkscape has a steep learning curve.

This diagram was generated with Cursor after asking it to study the [diagram](https://github.com/stolostron/governance-policy-framework/blob/main/images/policy-framework-architecture-diagram.png) from the governance policy framework. The time spent prompting the AI to fix the positions of all the elements was roughly equal to the time it would take a beginner (myself) to manually arrange the elements on Inkscape.


JustinKuli (dismissed):
LGTM! Should we also add it to the README or another doc?

/hold

jan-law:
Good idea, I can add it to the README

PR #383, opened 2025-08-07
Adjust cleanup to avoid panics and getting stuck

JustinKuli:
In one situation, an unresolved hub template was getting stuck on a ConfigurationPolicy and prevented the finalizer from being removed. While testing that case, it was found that the reconciler's DynamicWatcher could be nil in some cases and would thus cause a panic.

Both of these situations are adressed by adjusting the cleanup logic so that it is clear that the only actions performed while cleaning up are related to the required removal of the finalizers.

Refs:
 - https://issues.redhat.com/browse/ACM-22713


JustinKuli:
/cc @jan-law

Since Dale is out, if you have a chance, can you review this? Let me know if any bits are confusing - I think I combined a few refactor steps.

controllers/configurationpolicy_controller.go
```diff
@@ -247,39 +258,45 @@ func (r *ConfigurationPolicyReconciler) Reconcile(ctx context.Context, request c
...
+			}
 
-		r.addForUpdate(policy, statusChanged)
+			r.addForUpdate(policy, statusChanged)
 
+			return reconcile.Result{}, err
+		}
+	}
+
+	if err := r.manageDeletionFinalizer(policy, cleanup); err != nil {
```
yiraeChristineKim, line 291: Question If cleanup is false then do we need to delete finalizer?

JustinKuli: When cleanup is false, it needs to add the finalizer if it's using those prune options, or remove them if it is not.

> Oh now understand... because policy addon (template controller) uninstalled so policy cannot be deleted because of the template didn't resolved.. is this correct?

Yeah: when the controller is being uninstalled, *if* there's some weird template error, we need to ignore it and just clean up the policies.

PR #437, opened 2026-02-02
refactor: pass context with operator policy logger

jan-law:
Refactors operator policy logs to use the same logger from context, with `common.LogConstructor`. Pass context to operator policy API calls in the e2e tests. I didn't see any other API calls in the controller that were missing a context.

ref: https://issues.redhat.com/browse/ACM-26536


JustinKuli (approved):
LGTM!

PR #451, opened 2026-02-26
chore: bump to k8s v1.35

dhaiducek:


controllers/configurationpolicy_controller.go
```diff
@@ -3884,18 +3882,20 @@ func (r *ConfigurationPolicyReconciler) updatePolicyStatus(
...
 		eventMessage := fmt.Sprintf("%s: %s", policy.Status.ComplianceState, strings.Join(condMessages, "; "))
 		log.Info("Policy status message", "policy", policy.GetName(), "status", eventMessage)
 
-		r.Recorder.Event(
+		r.Recorder.Eventf(
 			policy,
+			nil,
 			eventType,
-			"Policy updated",
+			"PolicyUpdate",
```
dhaiducek: There's actually some conflicting information out there (controller-runtime says one thing while the Events API has reason and action flipped from controller-runtime to say that Action should be machine readable), but doing `oc get events` does show the reason but not the action, so it looks cleaner and more consistent to have this camel cased like the other events in the list.

JustinKuli (approved):
Looks good! Thanks for the bit of investigation into the event format

PR #454, opened 2026-03-05
Generate and save JUnit reports for all tests

JustinKuli:
Previously, we have been generating ginkgo JSON reports for only our e2e tests, but I don't believe they're used for anything specific. Now the e2e tests and the unit tests will generate JUnit XML reports, which can be used by a variety of tools. One such tool was failing when a test "suite" had no tests, so the "empty_test.go" file has been replaced by a "noop_test.go" file with a similar purpose.

The upload tasks in the workflow should run even when the tests fail, so that we can track flaky tests.

At some point in the future, some other automated process may pull the JUnit reports from the GitHub artifacts.

### DIFF
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -44,12 +44,6 @@ jobs:
         run: |
           make test-coverage
 
-      - name: Upload Unit Test Coverage
-        uses: actions/upload-artifact@v7
-        with:
-          name: coverage_unit
-          path: coverage_unit.out
-
       - name: Bootstrap K8s KinD Cluster and Deploy Controller
         run: |
           make kind-bootstrap-cluster-dev
@@ -62,12 +56,15 @@ jobs:
           export GOPATH=$(go env GOPATH)
           KUBECONFIG=${PWD}/kubeconfig_managed make e2e-test-running-in-cluster
 
-      - name: Upload Uninstall Test Coverage
+      - name: Upload Test Info
+        if: ${{ !cancelled() }}
         uses: actions/upload-artifact@v7
         with:
-          name: coverage_e2e_uninstall
-          path: coverage_e2e_uninstall.out
-      
+          name: latest-preflight-tests
+          path: |
+            coverage_*.out
+            report_*.xml
+
       - name: Debug
         if: ${{ failure() }}
         run: |
@@ -113,12 +110,14 @@ jobs:
           export GOPATH=$(go env GOPATH)
           KUBECONFIG=${PWD}/kubeconfig_managed make e2e-test-coverage
 
-      - name: Upload E2E Test Coverage
+      - name: Upload Basic e2e Test Info
+        if: ${{ !cancelled() }}
         uses: actions/upload-artifact@v7
-        if: ${{ matrix.kind == 'latest' }}
         with:
-          name: coverage_e2e_basic
-          path: coverage_e2e_basic.out
+          name: "${{matrix.kind}}-basic-tests"
+          path: |
+            coverage_*.out
+            report_*.xml
 
       - name: Debug
         if: ${{ failure() }}
@@ -166,12 +165,14 @@ jobs:
           export GOPATH=$(go env GOPATH)
           KUBECONFIG=${PWD}/kubeconfig_managed make e2e-test-hosted-mode-coverage
 
-      - name: Upload Hosted Mode Coverage
+      - name: Upload Hosted Mode e2e Test Info
+        if: ${{ !cancelled() }}
         uses: actions/upload-artifact@v7
-        if: ${{ matrix.kind == 'latest' }}
         with:
-          name: coverage_e2e_hosted_mode
-          path: coverage_e2e_hosted_mode.out
+          name: "${{matrix.kind}}-hosted-tests"
+          path: |
+            coverage_*.out
+            report_*.xml
 
       - name: Debug
         if: ${{ failure() }}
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -59,11 +59,23 @@ fmt:
...
+
+.PHONY: gotestsum
+gotestsum:
+	$(call go-get-tool,gotest.tools/gotestsum@$(GOTESTSUM_VERION))
+
+GOTST_FMT := --format=pkgname-and-test-fails --format-icons=text
+ifeq ($(GITHUB_ACTIONS), true)
+  GOTST_FMT := --format=github-actions
+endif
+
```
dhaiducek: This look like things that could potentially go into the common Makefile if we're using them elsewhere?

Makefile
```diff
@@ -59,11 +59,23 @@ fmt:
 # test section
 ############################################################
 
+GOTESTSUM_VERION := v1.13.0
```
dhaiducek: An annotation here would make discovering available versions more straightforward.

```suggestion
# https://github.com/gotestyourself/gotestsum/releases/latest
GOTESTSUM_VERION := v1.13.0
```

.github/workflows/kind.yml
```diff
@@ -67,7 +74,14 @@ jobs:
...
           name: coverage_e2e_uninstall
           path: coverage_e2e_uninstall.out
-      
+
+      - name: Upload Uninstall Test Report
+        if: ${{ !cancelled() }}
+        uses: actions/upload-artifact@v7
+        with:
+          name: uninstall-junit-report
+          path: report_e2e_uninstall.xml
```
dhaiducek: Any particular reason these are separate uploads? They could even combine the `*.out` and `*.xml` files into a single upload.

```suggestion
      - name: Upload Test Reports
        if: ${{ !cancelled() }}
        uses: actions/upload-artifact@v7
        with:
          name: preflight-tests
          path: *.xml
```

Makefile
```diff
@@ -250,30 +262,35 @@ install-resources:
...
+E2E_JUNIT = --junit-report=report_e2e.xml
 
 .PHONY: e2e-test
 e2e-test: e2e-dependencies
-	$(GINKGO) -v --procs=$(E2E_PROCS) $(E2E_TEST_ARGS) test/e2e -- -is_hosted=$(IS_HOSTED)
+	$(GINKGO) -v --procs=$(E2E_PROCS) $(E2E_TEST_ARGS) --output-dir=. $(E2E_JUNIT) test/e2e -- -is_hosted=$(IS_HOSTED)
 
 .PHONY: e2e-test-coverage
-e2e-test-coverage: E2E_TEST_ARGS = --json-report=report_e2e.json --label-filter='!hosted-mode && !running-in-cluster && !hub-templates-enabled' --output-dir=.
+e2e-test-coverage: E2E_TEST_ARGS = --label-filter='!hosted-mode && !running-in-cluster && !hub-templates-enabled'
```
dhaiducek, line 260: I think migrating away from the JSON report should be fine. Once it hits `stolostron`, you'll also have to update the SonarQube configuration to match:
https://github.com/stolostron/config-policy-controller/blob/d6321bdb68e702140e23e316e870d403d6f5cd91/sonar-project.properties#L9-L10

.github/workflows/kind.yml
```diff
@@ -256,6 +291,7 @@ jobs:
         uses: actions/download-artifact@v8
         with:
           merge-multiple: true
+          pattern: coverage*
```
dhaiducek: If the uploads were combined into a single step, the Makefile also has file filtering, so it wouldn't be necessary to distinguish between the uploads like it does here.

.github/workflows/kind.yml
dhaiducek: I might prefer combining since they're archived by default anyway. But if not combined, perhaps we should use `compression-level: 0`?

dhaiducek (commented):
Overall LGTM! I have some thoughts on the flows and notes on the origins of some of these original configurations.

Makefile
JustinKuli: Yeah! I was thinking we'd try it out here for a week or so, then if we want to do the same thing in other repos, we'd add it to the common Makefile.

dhaiducek: Nice! If it's going in the common Makefile, might as well put it there now? If we don't use it, we take it out, and if we do use it, it's already there and we don't have to take it out of this Makefile. 🙂

.github/workflows/kind.yml
JustinKuli: I'll try combining - I need to investigate exactly what it looks like to download these anyway. Initially I felt like having different artifact names for all of them would be nice, but now that doesn't seem as helpful as I thought...

I'll check if `compression-level: 0` skips the zipping completely, although I suspect it won't? But the reports will potentially have a lot of plain text logs, so I think they're pretty compressible - what made you think about the compression?

Makefile
JustinKuli: Ah, Sonar uses the JSON reports, of course... I should probably leave both then.

JustinKuli: Looking at it again, I'm not sure Sonar actually does use the JSON reports, since they don't seem to be saved in any artifacts... but I will leave them there anyway for now. Maybe they will be useful in the future!

Makefile
JustinKuli: https://github.com/stolostron/governance-policy-framework/pull/1018

Makefile
dhaiducek: My suspicion is that Sonar can consume either type--if that's true, it'd just be a matter of swapping out the json for the xml in the properties file.

.github/workflows/kind.yml
dhaiducek: I was thinking that the benefit to a lack of archive would be that the file could be looked at directly, potentially entirely in the browser but definitely without uncompressing. But given that it's always named, you're probably right that it's always an archive. And in that case, I figure we might as well cram as much as we can into a single archive so there are more files available there in a single download.

.github/workflows/kind.yml
JustinKuli: The changes now have `pattern: latest-` to avoid name collisions from the matrixed tests. (And I added `latest-` to the "unit" test section, to include them here :laughing: )

dhaiducek (approved):

JustinKuli:
Latest update moves the gotestsum format config to the common makefile: https://github.com/stolostron/governance-policy-framework/pull/1026

dhaiducek (approved):

PR #456, opened 2026-03-12
feat: detect new minor operator version on another channel

jan-law:
Added `spec.ComplianceConfig.MinorChannelUpgradeAvailable` to check if a newer minor version of the operator exists on a different channel. `MinorChannelUpgradeAvailable` only affects compliance when `InstallPlanApproval` is `Automatic` and the operator channels include the minor version number in the channel names.

Moved duplicate code to fetch packagemanifest and channels into helpers.

ref: https://issues.redhat.com/browse/ACM-30371

Channel name parsing assisted by Cursor/Claude 4.5

## Summary by CodeRabbit

* **New Features**
  * Added a compliance option to detect recommended newer minor release channels and a corresponding status condition; CRD schema updated with the new field (default: Compliant).

* **Behavior**
  * Policy status now reports a minor-channel upgrade condition when automatic upgrades and the new setting are applicable; checks are skipped if channel data is unavailable.

* **Tests**
  * Added and expanded unit and e2e tests covering channel selection, minor-channel detection, status transitions, and updated test fixtures/versions.

### DIFF
--- a/api/v1beta1/operatorpolicy_types.go
+++ b/api/v1beta1/operatorpolicy_types.go
@@ -141,6 +141,12 @@ type ComplianceConfig struct {
 	//
 	// +kubebuilder:default=Compliant
 	DeprecationsPresent ComplianceConfigAction `json:"deprecationsPresent,omitempty"`
+	// MinorChannelUpgradeAvailable specifies how the availability of a newer minor version channel
+	// should affect overall policy compliance when spec.UpgradeApproval is Automatic. The default
+	// value is `Compliant`.
+	//
+	// +kubebuilder:default=Compliant
+	MinorChannelUpgradeAvailable ComplianceConfigAction `json:"minorChannelUpgradeAvailable,omitempty"`
 }
 
 // OperatorPolicySpec defines the desired state of a particular operator on the cluster.
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -396,7 +396,7 @@ func (r *OperatorPolicyReconciler) handleResources(ctx context.Context, policy *
 		return earlyComplianceEvents, condChanged || changed, err
 	}
 
-	desiredSub, desiredOG, changed, err := r.buildResources(ctx, policy)
+	desiredSub, desiredOG, packageManifest, changed, err := r.buildResources(ctx, policy)
 	condChanged = condChanged || changed
 
 	if err != nil {
@@ -449,7 +449,14 @@ func (r *OperatorPolicyReconciler) handleResources(ctx context.Context, policy *
 		return earlyComplianceEvents, condChanged, err
 	}
 
-	err = r.updateDeprecationStatus(ctx, policy, subscription, csv)
+	err = r.updateDeprecationStatus(ctx, policy, subscription, csv, packageManifest)
+	if err != nil {
+		return earlyComplianceEvents, condChanged, err
+	}
+
+	changed, err = r.updateMinorChannelUpgradeStatus(policy, subscription, packageManifest)
+	condChanged = condChanged || changed
+
 	if err != nil {
 		return earlyComplianceEvents, condChanged, err
 	}
@@ -489,12 +496,13 @@ func (r *OperatorPolicyReconciler) handleResources(ctx context.Context, policy *
 // checks if the policy's spec is valid. It returns:
 //   - the built Subscription
 //   - the built OperatorGroup
+//   - the PackageManifest
 //   - whether the status has changed
 //   - an error if an API call failed
 //
 // The built objects can be used to find relevant objects for a 'mustnothave' policy.
 func (r *OperatorPolicyReconciler) buildResources(ctx context.Context, policy *policyv1beta1.OperatorPolicy) (
-	*operatorv1alpha1.Subscription, *operatorv1.OperatorGroup, bool, error,
+	*operatorv1alpha1.Subscription, *operatorv1.OperatorGroup, *unstructured.Unstructured, bool, error,
 ) {
 	opLog := ctrl.LoggerFrom(ctx)
 	disableTemplates := false
@@ -514,14 +522,14 @@ func (r *OperatorPolicyReconciler) buildResources(ctx context.Context, policy *p
 		if err != nil {
 			newError := fmt.Errorf("unable to create template resolver: %w", err)
 
-			return nil, nil, updateStatus(policy, validationCond([]error{newError})), nil
+			return nil, nil, nil, updateStatus(policy, validationCond([]error{newError})), nil
 		}
 
 		err = resolveVersionsTemplates(policy, tmplResolver)
 		if err != nil {
 			newError := fmt.Errorf("unable to create template resolver: %w", err)
 
-			return nil, nil, updateStatus(policy, validationCond([]error{newError})), nil
+			return nil, nil, nil, updateStatus(policy, validationCond([]error{newError})), nil
 		}
 	} else {
 		opLog.V(1).Info("Templates disabled by annotation")
@@ -532,30 +540,37 @@ func (r *OperatorPolicyReconciler) buildResources(ctx context.Context, policy *p
 	var returnedErr error
 
 	sub, subErr := buildSubscription(policy, tmplResolver)
-	if subErr == nil {
-		err := r.applySubscriptionDefaults(ctx, policy, sub)
-		if err != nil {
-			// If it's a PackageManifest API error, then that means it should be returned for the Reconcile method
-			// to requeue the request. This is to workaround the PackageManifest API not supporting watches.
-			if errors.Is(err, ErrPackageManifest) {
-				returnedErr = err
-			}
-
-			return nil, nil, updateStatus(policy, validationCond([]error{err})), returnedErr
-		}
-
-		if sub != nil && sub.Namespace == "" {
-			if r.DefaultNamespace != "" {
-				sub.Namespace = r.DefaultNamespace
-			} else {
-				newError := errors.New("namespace is required in spec.subscription")
-
-				return sub, nil, updateStatus(policy, validationCond([]error{newError})), nil
-			}
-		}
-	} else {
+	if subErr != nil {
 		// Invalid subscription spec - mark status and return without an API error
-		return sub, nil, updateStatus(policy, validationCond([]error{subErr})), nil
+		return sub, nil, nil, updateStatus(policy, validationCond([]error{subErr})), nil
+	}
+
+	packageManifest, err := r.getPackageManifest(ctx, sub.Spec.Package)
+	if err != nil {
+		returnedErr = fmt.Errorf("%wfailed to get the PackageManifest", ErrPackageManifest)
+
+		// If it's a PackageManifest API error, then that means it should be returned for the Reconcile method
+		// to requeue the request. This is to workaround the PackageManifest API not supporting watches.
+		return nil, nil, nil, updateStatus(policy, validationCond([]error{err})), returnedErr
+	}
+
+	err = r.applySubscriptionDefaults(ctx, policy, sub, packageManifest)
+	if err != nil {
+		if errors.Is(err, ErrPackageManifest) {
+			returnedErr = err
+		}
+
+		return nil, nil, nil, updateStatus(policy, validationCond([]error{err})), returnedErr
+	}
+
+	if sub != nil && sub.Namespace == "" {
+		if r.DefaultNamespace != "" {
+			sub.Namespace = r.DefaultNamespace
+		} else {
+			newError := errors.New("namespace is required in spec.subscription")
+
+			return sub, nil, nil, updateStatus(policy, validationCond([]error{newError})), nil
+		}
 	}
 
 	opGroupNS := r.DefaultNamespace
@@ -566,20 +581,20 @@ func (r *OperatorPolicyReconciler) buildResources(ctx context.Context, policy *p
 	opGroup, ogErr := buildOperatorGroup(policy, opGroupNS, tmplResolver)
 	if ogErr != nil {
 		// OperatorGroup spec invalid - mark status, don't requeue
-		return sub, nil, updateStatus(policy, validationCond([]error{ogErr})), nil
+		return sub, nil, nil, updateStatus(policy, validationCond([]error{ogErr})), nil
 	}
 
 	watcher := opPolIdentifier(policy.Namespace, policy.Name)
 
 	gotNamespace, err := r.DynamicWatcher.Get(watcher, namespaceGVK, "", opGroupNS)
 	if err != nil {
-		return sub, opGroup, false, fmt.Errorf("error getting operator namespace: %w", err)
+		return sub, opGroup, packageManifest, false, fmt.Errorf("error getting operator namespace: %w", err)
 	}
 
 	if gotNamespace == nil && policy.Spec.ComplianceType.IsMustHave() {
 		newError := fmt.Errorf("the operator namespace ('%v') does not exist", opGroupNS)
 
-		return sub, opGroup, updateStatus(policy, validationCond([]error{newError})), nil
+		return sub, opGroup, packageManifest, updateStatus(policy, validationCond([]error{newError})), nil
 	}
 
 	changed, overlapErr, apiErr := r.checkSubOverlap(ctx, policy, sub)
@@ -591,12 +606,12 @@ func (r *OperatorPolicyReconciler) buildResources(ctx context.Context, policy *p
 		// When an overlap is detected, the generated subscription and operatorgroup
 		// will be considered to be invalid to prevent creations/updates.
 		// sub and opgroup should be nil to prevent creation/update.
-		return nil, nil, updateStatus(policy, validationCond([]error{overlapErr})), returnedErr
+		return nil, nil, nil, updateStatus(policy, validationCond([]error{overlapErr})), returnedErr
 	}
 
 	changed = updateStatus(policy, validationCond([]error{})) || changed
 
-	return sub, opGroup, changed, returnedErr
+	return sub, opGroup, packageManifest, changed, returnedErr
 }
 
 func (r *OperatorPolicyReconciler) checkSubOverlap(
@@ -677,7 +692,10 @@ func (r *OperatorPolicyReconciler) checkSubOverlap(
 // utilizing the PackageManifest API. If the PackageManifest can not be found, and a subscription already exists for the
 // operator, then information from the found subscription will be used.
 func (r *OperatorPolicyReconciler) applySubscriptionDefaults(
-	ctx context.Context, policy *policyv1beta1.OperatorPolicy, subscription *operatorv1alpha1.Subscription,
+	ctx context.Context,
+	policy *policyv1beta1.OperatorPolicy,
+	subscription *operatorv1alpha1.Subscription,
+	packageManifest *unstructured.Unstructured,
 ) error {
 	opLog := ctrl.LoggerFrom(ctx)
 	subSpec := subscription.Spec
@@ -691,27 +709,13 @@ func (r *OperatorPolicyReconciler) applySubscriptionDefaults(
 
 	opLog.V(1).Info("Determining defaults for the subscription based on the PackageManifest")
 
-	// PackageManifests come from an API server and not a Kubernetes resource, so the DynamicWatcher can't be used since
-	// it utilizes watches. The namespace doesn't have any meaning but is required.
-	packageManifest, err := r.DynamicClient.Resource(packageManifestGVR).Namespace("default").Get(
-		ctx, subSpec.Package, metav1.GetOptions{},
-	)
-	if err != nil {
-		if k8serrors.IsNotFound(err) {
-			if !r.usingExistingSubIfFound(policy, subscription) {
-				return fmt.Errorf(
-					"%wthe subscription defaults could not be determined because the PackageManifest was not found",
-					ErrPackageManifest,
-				)
-			}
-
+	if packageManifest == nil {
+		if r.usingExistingSubIfFound(policy, subscription) {
 			return nil
 		}
 
-		opLog.Error(err, "Failed to get the PackageManifest", "name", subSpec.Package)
-
 		return fmt.Errorf(
-			"%wthe subscription defaults could not be determined because the PackageManifest API returned an error",
+			"%wthe subscription defaults could not be determined because the PackageManifest was not found",
 			ErrPackageManifest,
 		)
 	}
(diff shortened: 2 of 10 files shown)


test/resources/case38_operator_install/operator-policy-manual-upgrades.yaml
```diff
@@ -20,12 +20,12 @@ spec:
     targetNamespaces:
       - operator-policy-testns
   subscription:
-    channel: strimzi-0.36.x
+    channel: strimzi-0.50.x
```
jan-law, line 23: Now that this test expects the install plan to complete, I noticed 0.36.0 failed to upgrade to 0.36.1 on newer versions of k8s. Upgraded to fix the install

controllers/operatorpolicy_status.go
```diff
@@ -307,6 +307,19 @@ func calculateComplianceCondition(policy *policyv1beta1.OperatorPolicy) metav1.C
 		}
 	}
 
+	if !policy.Spec.ComplianceType.IsMustNotHave() &&
+		policy.Spec.ComplianceConfig.MinorChannelUpgradeAvailable == "NonCompliant" {
+		idx, cond = policy.Status.GetCondition(minorChannelConditionType)
+		if idx != -1 {
+			if cond.Status != metav1.ConditionTrue {
+				messages = append(messages, cond.Message)
```
jan-law, line 315: Although the deprecation condition message appears regardless of compliance, I remember there was an effort to declutter the message. What do you think of only adding the minor channel message when the condition is false?

controllers/operatorpolicy_controller.go
```diff
@@ -1647,6 +1638,72 @@ func (r *OperatorPolicyReconciler) updateDeprecationStatus(ctx context.Context,
...
+	policy *policyv1beta1.OperatorPolicy,
+	desiredSub *operatorv1alpha1.Subscription,
+) error {
+	opLog := ctrl.LoggerFrom(ctx)
+	channelOKCond := metav1.Condition{
+		Type:    minorChannelConditionType,
+		Status:  metav1.ConditionTrue,
+		Reason:  "Recommended",
+		Message: "the subscription is on the recommended channel",
+	}
```
JustinKuli: Nit: this could be defined in `operatorpolicy_status.go` near all the other conditions.

controllers/operatorpolicy_controller.go
```diff
@@ -3009,6 +3066,93 @@ func (r *OperatorPolicyReconciler) mergeObjects(
...
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return nil, err
+		}
+
+		return nil, fmt.Errorf("failed to get the PackageManifest %q %w", packageName, err)
+	}
+
+	return packageManifest, nil
+}
```
JustinKuli, line 3078: Can this result be re-used somehow? Since PackageManifest isn't a "normal" kubernetes object, I don't think controller-runtime or anything else will help us. So potentially the API is called 3 times during the evaluation of this policy. It might be especially important because if the same package is added by multiple catalogs, the one that is returned by the API is inconsistent. If different calls during the same Reconcile get different results, it could cause a very difficult to understand situation.

controllers/operatorpolicy_controller.go
```diff
@@ -3009,6 +3066,93 @@ func (r *OperatorPolicyReconciler) mergeObjects(
...
+	}
+
+	return selectedChannelName, nil
+}
+
+// getNextMinorChannel checks if there is a newer minor channel available,
+// assuming channel naming patterns similar to:
+// selected-channel-name-1.2, selected-channel-namev1.2.x, 3.17, 2.2.x.
+// Returns the newer channel name if found, along with a boolean indicating success.
+func getNextMinorChannel(channels []interface{}, selectedChannelName string) (string, bool) {
```
JustinKuli: This is pretty good, and the tests really help understand which things it supports. But I think there are cases it might handle incorrectly, and I wonder if overall it's too flexible. I think it might be better to just specify a small number of supported regexes for the versions, and ignore channels that don't match any of those.

A case that I don't think would be properly handled now: `openshift-virt-1.2`. The `v` in the channel would mess up the major version: I think it would try to use `irt-1`.

I'm looking through what channel names I can find in an OpenShift cluster to get an idea of what we want to support, and I'll comment here with more info soon, but I'm curious if you already tried regex and had any specific problems?

controllers/operatorpolicy_status.go
JustinKuli: Yes - I think what you have here is good

controllers/operatorpolicy_controller.go
```diff
@@ -1647,6 +1638,72 @@ func (r *OperatorPolicyReconciler) updateDeprecationStatus(ctx context.Context,
...
+	if err != nil || !ok {
+		return fmt.Errorf("failed to retrieve channels in PackageManifest %q: %w", desiredSub.Spec.Package, err)
+	}
+
+	if nextChannelName, found := getNextMinorChannel(channels, selectedChannelName); found {
+		updateStatus(policy, metav1.Condition{
+			Type:    minorChannelConditionType,
+			Status:  metav1.ConditionFalse,
+			Reason:  "UpgradeAvailable",
+			Message: "a newer version of the operator is available in another channel: " + nextChannelName,
```
JustinKuli: I wonder if we should list *all* possible "next" channels, if there are multiple. Maybe it's not worth it...

JustinKuli (commented):
Overall I think this is looking pretty good. I bet this would handle the common cases we want, I'm just hoping to make the version handling a little more robust for odd cases.

controllers/operatorpolicy_controller.go
jan-law: Regex would work too. Originally I was unsure if the list of regexes would get too long, so I decided to ignore any channels that didn't match the unit tests with `_, errMajorVer := strconv.Atoi(versionNums[0])  ... if errMajorVer != nil { return "", false } `. 

Will update with regex. Let me know if you see any more channel names to support

JustinKuli: Here's the command I used, in case you want to look through a bunch of possible inputs:
```
❯ oc get packagemanifest -o jsonpath='{.items[*].status.channels[*].name}' | tr ' ' '\n' | sort | uniq
```

Here some examples of patterns I think we want to match and handle:
```
1.10
1.10.x
2022.4
amq-streams-2.2.x
candidate-3.10
ocm-2.13
stable-1.33
strimzi-0.50.x
stable-v1.18
v0.2
v1.0.x
```

And here's some I think we want to ignore:
```
1.29-nightly
1.x
2022.4.0
alpha-0.3.3
amq-streams-2.x
latest
lts
v1
v2-stable
v42.10.0
```

Finally [a regex](https://regex101.com/r/7UsnBY/2) I came up with: `(?:v|-|^)(\d+)\.(\d+)(?:.x)?$`. I'll break it down by groups:
- `(?:v|-|^)` a non-capturing group to find the end of the "prefix", either the last `v`, the last `-` or just the very beginning of the input (for channels like `2.10`)
- `(\d+).(\d+)` capturing groups (with a `.` between them) to get the major and minor version
- `(?:.x)?$` a non-capturing group to allow a suffix `.x`, but not require it, right at the end of the input.

What do you think? Does that hit the inputs you were also expecting, did I miss anything or handle any badly?

jan-law: Thanks! After writing this regex out I realized we could treat the ending `v` or `-` as part of the "prefix", to simplify a bit further. Added openshift-virt to the unit tests too.

Also, the current logic checking for the next `minor version + 1` won't work if a channel is missing, eg `v1.0, v1.2, v1.3`. Do we want to support this case as well?

JustinKuli: > Also, the current logic checking for the next minor version + 1 won't work if a channel is missing, eg v1.0, v1.2, v1.3. Do we want to support this case as well?

I meant to mention this; yes, I think we should support that kind of case as well, and mention all of those "newer" channels.

controllers/operatorpolicy_controller.go
jan-law: Package manifests are now fetched once in `buildResources()`, then passed to `applySubscriptionDefaults()`, `updateDeprecationStatus()` and `updateMinorChannelStatus()`. Let me know what you think of this implementation

controllers/operatorpolicy_controller.go
jan-law: All "newer" channels are listed in the condition message now

jan-law:
@coderabbitai review

controllers/operatorpolicy_status.go
```diff
@@ -307,6 +307,19 @@ func calculateComplianceCondition(policy *policyv1beta1.OperatorPolicy) metav1.C
...
+				foundNonCompliant = true
+			}
+		} else {
+			foundNonCompliant = true
+		}
```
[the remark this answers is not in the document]
jan-law: In this case, I think we still want to set foundNonCompliant = true when the condition is not found. Following the pattern with `policy.Spec.ComplianceConfig.DeprecationsPresent` above the highlighted code, users expect to see a condition when they specifically enable `MinorChannelUpgradeAvailable == "NonCompliant"`

api/v1beta1/operatorpolicy_types.go
```diff
@@ -141,6 +141,12 @@ type ComplianceConfig struct {
 	//
 	// +kubebuilder:default=Compliant
 	DeprecationsPresent ComplianceConfigAction `json:"deprecationsPresent,omitempty"`
+	// MinorChannelUpgradeAvailable specifies how the availability of a newer minor version channel
+	// should affect overall policy compliance when installPlanApproval is Automatic. The default
+	// value is `Compliant`.
+	//
+	// +kubebuilder:default=Compliant
+	MinorChannelUpgradeAvailable ComplianceConfigAction `json:"minorChannelUpgradeAvailable,omitempty"`
```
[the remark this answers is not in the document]
jan-law: Fixed the wording to use spec.UpgradeApproval

controllers/operatorpolicy_controller.go
```diff
@@ -1647,6 +1636,63 @@ func (r *OperatorPolicyReconciler) updateDeprecationStatus(ctx context.Context,
...
+	if desiredSub == nil || !policy.Spec.ComplianceType.IsMustHave() {
+		return nil
+	}
+
+	isComplianceConfigCompliant := policy.Spec.ComplianceConfig.MinorChannelUpgradeAvailable == "Compliant"
+	isManualUpgradeApproval := policy.Spec.UpgradeApproval != "Automatic"
+
+	if isComplianceConfigCompliant || isManualUpgradeApproval {
+		updateStatus(policy, minorChannelOKCond())
+
```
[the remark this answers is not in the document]
jan-law: The upgradeApproval condition is correct here. The comment description incorrectly used installPlanApproval, which has been fixed

test/e2e/case38_install_operator_test.go
```diff
@@ -1641,7 +1642,86 @@ var _ = Describe("Testing OperatorPolicy", Label("supports-hosted"), func() {
...
+				opPolName,
+				true,
+				[]policyv1.RelatedObject{},
+				metav1.Condition{
+					Type:   "MinorChannelUpgradeAvailable",
+					Status: metav1.ConditionFalse,
+					Reason: "UpgradeAvailable",
+					Message: "a newer version of the operator is available in channel(s): " +
+						strings.Join(laterChannels, ", "),
+				},
```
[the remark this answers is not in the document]
jan-law: Suggestion committed

controllers/operatorpolicy_controller_test.go
```diff
@@ -701,3 +719,258 @@ func TestCanonicalizeVersions(t *testing.T) {
...
+		},
+	}
+
+	// condition should not be updated
+	err := r.updateMinorChannelUpgradeStatus(t.Context(), policy, sub, nil)
+	assert.NoError(t, err)
+
+	idx, _ := policy.Status.GetCondition(minorChannelConditionType)
+	assert.Equal(t, -1, idx)
+}
```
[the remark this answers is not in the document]
jan-law: > update the controller to overwrite it with an explicit unavailable/unknown result

Done

test/e2e/case38_install_operator_test.go
```diff
@@ -1641,7 +1642,89 @@ var _ = Describe("Testing OperatorPolicy", Label("supports-hosted"), func() {
...
+				`[{"op": "replace",
+				"path": "/spec/complianceConfig/minorChannelUpgradeAvailable",
+				"value": "NonCompliant"}]`)
+
+			By("Fetching the latest channel names")
+			pm := utils.GetWithTimeout(
+				targetK8sDynamic, gvrPackageManifest, "strimzi-kafka-operator", testNamespace, true, eventuallyTimeout)
+			Expect(pm).ToNot(BeNil())
+			channels, _, err := unstructured.NestedSlice(pm.Object, "status", "channels")
+			Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
jan-law: package manifest is a special case - the namespace is ignored here

controllers/operatorpolicy_controller.go
```diff
@@ -1536,19 +1539,11 @@ func (r *OperatorPolicyReconciler) updateDeprecationStatus(ctx context.Context,
...
-				"PackageManifestName", desiredSub.Spec.Package)
-
-			return nil
-		}
+	if packageManifest == nil {
+		opLog.Info("PackageManifest not found; skipping DeprecationStatus check",
+			"PackageManifestName", desiredSub.Spec.Package)
 
-		return fmt.Errorf("failed to get the PackageManifest %q %w", desiredSub.Spec.Package, err)
+		return nil
```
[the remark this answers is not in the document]
jan-law: I'd like to preserve the existing logic, as package manifests disappearing should be a transient condition

controllers/operatorpolicy_controller_test.go
```diff
@@ -701,3 +719,259 @@ func TestCanonicalizeVersions(t *testing.T) {
...
+		},
+	}
+
+	err := r.updateMinorChannelUpgradeStatus(policy, sub, nil)
+	assert.NoError(t, err)
+
+	_, cond := policy.Status.GetCondition(minorChannelConditionType)
+	assert.Equal(t, metav1.ConditionTrue, cond.Status)
+	assert.Equal(t, "UpgradeAvailable", cond.Reason)
+	assert.Equal(t, "Skipping check for minor channel upgrades; PackageManifest not found", cond.Message)
```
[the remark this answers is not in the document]
jan-law: done

controllers/operatorpolicy_controller_test.go
```diff
@@ -701,3 +719,259 @@ func TestCanonicalizeVersions(t *testing.T) {
...
+		},
+	}
+
+	err := r.updateMinorChannelUpgradeStatus(policy, sub, pkgManifest)
+	assert.NoError(t, err)
+
+	_, cond := policy.Status.GetCondition(minorChannelConditionType)
+	assert.Equal(t, metav1.ConditionTrue, cond.Status)
+	assert.Equal(t, "Recommended", cond.Reason)
+	assert.Equal(t, "the subscription is on the recommended channel", cond.Message)
```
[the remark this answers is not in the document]
jan-law: not an issue. the test name is "InstallPlanApprovalNotAutomatic"

deploy/crds/policy.open-cluster-management.io_operatorpolicies.yaml
```diff
@@ -81,6 +81,16 @@ spec:
                     - Compliant
                     - NonCompliant
                     type: string
+                  minorChannelUpgradeAvailable:
+                    default: Compliant
+                    description: |-
+                      MinorChannelUpgradeAvailable specifies how the availability of a newer minor version channel
+                      should affect overall policy compliance when spec.UpgradeApproval is Automatic. The default
+                      value is `Compliant`.
```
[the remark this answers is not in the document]
jan-law: By definition, all minor version channels include the minor version in the name. Other types of channels like stable do not follow this naming convention. https://olm.operatorframework.io/docs/best-practices/channel-naming/#example-3-recommended-option-is-the-most-common-scenarios

controllers/operatorpolicy_controller.go
```diff
@@ -1647,6 +1636,64 @@ func (r *OperatorPolicyReconciler) updateDeprecationStatus(ctx context.Context,
...
+// updateMinorChannelUpgradeStatus checks if a newer minor version channel is available for the operator
+// when using automatic install plan approval, and updates the condition
+func (r *OperatorPolicyReconciler) updateMinorChannelUpgradeStatus(
+	policy *policyv1beta1.OperatorPolicy,
+	desiredSub *operatorv1alpha1.Subscription,
+	packageManifest *unstructured.Unstructured,
+) error {
+	if desiredSub == nil || !policy.Spec.ComplianceType.IsMustHave() {
+		return nil
+	}
```
JustinKuli: If the policy was previously `musthave`, and had its `MinorChannelUpgradeAvailable` condition populated at that point, but then was changed to `mustnothave`, then I think that old condition could cause an unexpected overall compliance? Is that right, or am I missing where that might be handled?

Maybe we want to set the condition to some skipped state, or clear it out somehow.

controllers/operatorpolicy_controller.go
```diff
@@ -1647,6 +1636,64 @@ func (r *OperatorPolicyReconciler) updateDeprecationStatus(ctx context.Context,
...
+			Status: metav1.ConditionFalse,
+			Reason: "UpgradeAvailable",
+			Message: "a newer version of the operator is available in channel(s): " +
+				strings.Join(nextChannelNames, ", "),
+		})
+
+		return nil
+	}
+
+	updateStatus(policy, minorChannelOKCond())
```
JustinKuli: If this returns true, then shoudn't a new compliance event should be emitted? But if this isn't returned, I think that it might get missed in some case.

Should this function use the `changed` and `condChanged` pattern?

controllers/operatorpolicy_controller.go
JustinKuli: It's great! Thanks for that update.

controllers/operatorpolicy_controller.go
```diff
@@ -3009,6 +3056,112 @@ func (r *OperatorPolicyReconciler) mergeObjects(
...
+// getNextMinorChannel checks if there are newer minor channels available,
+// assuming channel naming patterns similar to unit tests
+// Returns the newer channel name if found, along with a boolean indicating success.
+func getNextMinorChannel(channels []interface{}, selectedChannelName string) ([]string, bool) {
+	if selectedChannelName == "" || len(channels) == 0 {
+		return nil, false
+	}
+
+	// find the version numbers in the selected channel name using regex
+	pattern := `^([a-z-]*)(\d+)\.(\d+)(\.x)?$`
```
JustinKuli: Nit: there are some channels that use capital letters, I found `Fast` and `Stable`. I didn't find any that use an capital letters and have the version info we want, so it might not matter, but I think the first group could be updated to include capital letters: `([a-zA-Z-]*)`

controllers/operatorpolicy_controller.go
```diff
@@ -3009,6 +3056,112 @@ func (r *OperatorPolicyReconciler) mergeObjects(
...
+	minorVersion := matches[3]
+	suffix := matches[4] // ".x" or ""
+
+	selectedMinorVersion, err := strconv.Atoi(minorVersion)
+	if err != nil {
+		return nil, false
+	}
+
+	nextChannelPattern := fmt.Sprintf(`^%s%s\.(\d+)%s$`, prefix, majorVersion, suffix)
+	nextRe := regexp.MustCompile(nextChannelPattern)
```
JustinKuli: I think it would be fine because channel names will have *some* restrictions, and/or the first pattern might prevent anything too weird, but using `MustCompile` on something that potentially has arbitrary inputs feels wrong. I think it's easy enough to just check that the pattern compiles without an error, rather than risk (even the remote chance of) a panic.

controllers/operatorpolicy_controller.go
JustinKuli: Ok, I see that the overall compliance only considers this condition if the policy is `musthave`. But this stale condition could still be in the status, and possibly be confusing if someone noticed.

It's acceptable either way - you don't have to fix this if there's other work you want to get to.

controllers/operatorpolicy_controller.go
JustinKuli: I know it's outside of the current changes, but I noticed this typo 😆 
```suggestion
		return fmt.Errorf("failed to get packageName in the packageManifest %q %w", desiredSub.Spec.Package, err)
```

controllers/operatorpolicy_controller.go
```diff
@@ -1647,6 +1636,64 @@ func (r *OperatorPolicyReconciler) updateDeprecationStatus(ctx context.Context,
...
+func (r *OperatorPolicyReconciler) updateMinorChannelUpgradeStatus(
+	policy *policyv1beta1.OperatorPolicy,
+	desiredSub *operatorv1alpha1.Subscription,
+	packageManifest *unstructured.Unstructured,
+) error {
+	if desiredSub == nil || !policy.Spec.ComplianceType.IsMustHave() {
+		return nil
+	}
+
+	isComplianceConfigCompliant := policy.Spec.ComplianceConfig.MinorChannelUpgradeAvailable == "Compliant"
```
JustinKuli: I'm not sure I like the `complianceConfig` changing the details of the specific Condition. The test where the complianceConfig is switched from Compliant to NonCompliant to Compliant made me notice this: I feel like only the overall compliance should change in that situation, not this specific condition. What do you think? I could possibly be convinced.

JustinKuli (commented):
This is looking quite good overall. There are just a few specific situations I think need looking into

controllers/operatorpolicy_controller.go
jan-law: Your comment makes sense. I was confused by the conditions defined in operatorpolicy_status.go and thought conditions were supposed to change details when the compliance changed. I can keep the details the same regardless of compliance

controllers/operatorpolicy_controller.go
jan-law: Thanks, updated.

Aside: Does that mean `err = r.updateDeprecationStatus(ctx, policy, subscription, csv, packageManifest)` does not emit compliance events intentionally?

controllers/operatorpolicy_controller.go
jan-law: I like the idea of a Skipped condition. Added it in

controllers/operatorpolicy_controller.go
JustinKuli: I don't know about "intentionally" - we probably just overlooked it.

JustinKuli (approved):
LGTM! Thanks for taking on all the review comments, I think this is looking great

PR #465, opened 2026-04-21
test: replace quay-operator tests

jan-law:
The operator created in some OperatorPolicy tests was removed from OperatorHub. Replace with grc example operator.

It turns out the cluster-scoped test that deployed `quay-operator` was already running as `Serial`. I used `example-operator` for the replacement.

## Summary by CodeRabbit

* **Tests**
  * Updated test fixtures and scenarios to use updated operator package and catalog source references across multiple test cases.


JustinKuli (approved):
Thanks!

PR #469, opened 2026-05-07
fix: ignore startingCSV only in subscription

JustinKuli:
Previously, the `startingCSV` in a Subscription that was adopted by an OperatorPolicy could be incorrectly considered an allowed version, even if that version was not specified in the `versions` or `startingCSV` of the OperatorPolicy. Now, when determining which versions should be considered compliant for the operator, only the versions actually in the policy are used.

Refs:
 - https://issues.redhat.com/browse/ACM-32977

Assisted-by: Cursor, claude-4.6-opus-high

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed operator policy approval logic to correctly consider starting CSV values provided in raw subscription data when evaluating approved CSVs.

* **Tests**
  * Added a regression test to verify approval behavior when adopting subscriptions whose starting CSV is not listed in the policy's versions.

* **Chores**
  * Enhanced e2e-debug output to list and describe all pods before collecting controller logs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -300,6 +300,11 @@ e2e-stop-instrumented:
 
 .PHONY: e2e-debug
 e2e-debug:
+	@echo running pods:
+	-kubectl get pods -A
+	@echo "::group::describe pods"
+	-kubectl describe pods -A
+	@echo "::endgroup::"
 	@echo local controller log:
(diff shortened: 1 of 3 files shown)


jan-law (approved):
LGTM, sorry for the delay

PR #471, opened 2026-05-13
Preserve createdByPolicy through recreates

JustinKuli:
Previously, when a configurationpolicy would recreate an object in order to enforce the policy, the controller would always mark that object in the status as having been created by the policy. While this is technically true, it meant that a resource created *before* the policy might be deleted if `pruneObjectBehavior: DeleteIfCreated` was set, and that was often surprising to users.

Now, when an object is recreated by a policy, the previous setting for that object in the status is preserved.

Refs:
 - https://issues.redhat.com/browse/ACM-26300

Assisted-by: Cursor, claude-4.6-opus-high

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ConfigurationPolicy handling of object recreation scenarios to properly preserve object ownership information when objects are recreated during policy enforcement and reconciliation.

* **Tests**
  * Extended test suite for ConfigurationPolicy recreation functionality with comprehensive coverage for ownership state tracking behavior when policies trigger object recreation, including verification of ownership state preservation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/open-cluster-management-io/config-policy-controller/pull/471)

### DIFF
--- a/controllers/configurationpolicy_controller.go
+++ b/controllers/configurationpolicy_controller.go
@@ -2335,6 +2335,24 @@ type objectTmplEvalResultWithEvent struct {
 	event  objectTmplEvalEvent
 }
 
+// uidWasCreatedByPolicy reports whether the configuration policy's status lists the given object UID
+// with createdByPolicy set. The UID should be the cluster object's metadata.uid from before an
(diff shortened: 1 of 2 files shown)


jan-law (approved):
LGTM!

PR #480, opened 2026-06-24
Cleaning up removed CustomResources

JustinKuli:
Previously, if a custom resource endpoint was no longer available on a
cluster, but a policy still referenced an object of that type, the
policy would become stuck. The failed lookups would prevent the
finalizer from being removed.

Now, when the API response indicates there is no match for a resource
version, the controller will consider the child object as properly
removed.

Refs:
 - https://issues.redhat.com/browse/ACM-31401

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup behavior when a managed object or its API is no longer available, reducing unnecessary delete-failure reports.
  * Policy deletion now handles missing child resource types more reliably and can complete even when related objects point to removed kinds.

* **Tests**
  * Added coverage for unavailable child-object error cases.
  * Expanded end-to-end coverage for policy deletion scenarios involving missing child resource definitions.


jan-law (approved):
LGTM

PR #488, opened 2026-07-13
Detect AllNamespaces operator in another namespace

### LINKED ISSUE #479: OperatorPolicy reports NonCompliant when operator is already installed in a different namespace
When an operator is already installed in namespace A by another controller (e.g., the OCP ingress operator auto-installing `servicemeshoperator3` for Gateway API), an OperatorPolicy targeting namespace B for the same operator stays permanently NonCompliant, even though the operator is running and the Subscription in namespace B reports `installedCSV` and `state: AtLatestKnown`.

## Root cause

OLM copies the CSV from namespace A to namespace B but labels the copy with `operators.coreos.com/<pkg>.<namespace-A>` (the original namespace) and `olm.copiedFrom: <namespace-A>`.

The OperatorPolicy's `handleCSV()` uses `subLabelSelector()` which filters by `operators.coreos.com/<pkg>.<namespace-B>`.
The copied CSV does not match this selector, so `csvList` is empty, and the policy reports "the ClusterServiceVersion required by the policy was not found".

Relevant code: `controllers/operatorpolicy_controller.go` lines 2457-2459, 3186-3198.

## Why this matters

Using a custom namespace is the natural way to avoid conflicts with pre-existing operator installations.
Without it, `enforce` + `musthave` overwrites the pre-existing Subscription's fields (e.g., changes `installPlanApproval` from `Manual` to `Automatic`), and `mustnothave` removes the operator unconditionally regardless of who created it.

A custom namespace cleanly isolates cleanup (`mustnothave` only removes resources in the policy's namespace), but the permanent NonCompliant status makes it unusable in mixed environments where some clusters have a pre-existing operator and others don't.

## Steps to reproduce

1. Install an operator via OLM Subscription in namespace `sail-operator` with `installPlanApproval: Manual` (simulating what the OCP ingress operator does for Gateway API)
2. Create an OperatorPolicy with `enforce` + `musthave` targeting namespace `mesh-operator` for the same operator package
3. The Subscription is created in `mesh-operator`, OLM resolves it, InstallPlan completes, CSV appears in `mesh-operator` with `olm.copiedFrom: sail-operator`
4. OperatorPolicy reports NonCompliant: "the ClusterServiceVersion required by the policy was not found"

## Expected behavior

The OperatorPolicy should recognize that the operator is installed and running, either by falling back to `sub.Status.InstalledCSV` when the label selector finds no results, or by also matching CSVs with the `olm.copiedFrom` label.

mkolesnik:
OperatorPolicy reported NonCompliant for operators already available in the target namespace via an AllNamespaces installation elsewhere. In enforce mode it created a duplicate Subscription that would never produce a real installation and become permanently orphaned if the original was removed (OLM bug operator-framework/operator-lifecycle-manager#3866).

Detect the existing installation via OLM's copied CSV in the target namespace and report Compliant without creating any resources. When the original is removed, OLM deletes the copied CSV, the detection stops, and normal flow takes over.

Clean up a stale Subscription if enforce mode created one before the detection could fire. OperatorGroup cleanup is deferred until ownership tracking is available (#489).

Closes #479

## Summary by CodeRabbit

- **New Features**
  - Operator policies now detect existing AllNamespaces installations created in other namespaces and treat them as compliant.
  - When an existing copied CSV is found, policies update conditions and avoid creating duplicate local Subscription/OperatorGroup/InstallPlan/CSV resources.

- **Bug Fixes**
  - Version-allowance behavior is consistent and properly marks unapproved versions as not allowed.
  - Catalog source unhealthy polarity is handled correctly in status conditions.
  - Stale subscriptions are cleaned up while preserving pre-existing operator resources.

- **Tests**
  - Added unit and end-to-end coverage for existing and competing AllNamespaces scenarios.

### DIFF
--- a/controllers/operatorpolicy_controller.go
+++ b/controllers/operatorpolicy_controller.go
@@ -405,6 +405,41 @@ func (r *OperatorPolicyReconciler) handleResources(ctx context.Context, policy *
 		return earlyComplianceEvents, condChanged, err
 	}
 
+	// If an AllNamespaces operator is already installed elsewhere, OLM copies its CSV into this
+	// namespace. Short-circuit to avoid creating a duplicate Sub that would become orphaned.
+	// See https://github.com/operator-framework/operator-lifecycle-manager/issues/3866
+	if desiredSub != nil && desiredOG != nil &&
+		policy.Spec.ComplianceType.IsMustHave() &&
+		len(desiredOG.Spec.TargetNamespaces) == 0 {
+		copiedCSV, err := r.findCopiedCSV(ctx, policy, desiredSub)
+		if err != nil {
+			return earlyComplianceEvents, condChanged, err
+		}
+
+		if copiedCSV != nil {
+			condChanged = r.handleExistingInstallation(policy, copiedCSV) || condChanged
+
+			// Compliance config checks apply to the CSV regardless of whether the installation
+			// is local or existing, but need the packageManifest and desiredSub context.
+			if err := r.updateDeprecationStatus(ctx, policy, desiredSub, copiedCSV, packageManifest); err != nil {
+				return earlyComplianceEvents, condChanged, err
+			}
+
+			changed, err = r.updateMinorChannelUpgradeStatus(policy, desiredSub, packageManifest)
+			condChanged = condChanged || changed
+
+			if err != nil {
+				return earlyComplianceEvents, condChanged, err
+			}
+
+			if err := r.cleanupStaleInstallation(ctx, policy, desiredSub); err != nil {
+				return earlyComplianceEvents, condChanged, err
+			}
+
+			return earlyComplianceEvents, condChanged, nil
+		}
+	}
+
 	desiredSubName := ""
 	if desiredSub != nil {
 		desiredSubName = desiredSub.Name
@@ -920,6 +955,26 @@ func policyStartingCSV(policy *policyv1beta1.OperatorPolicy) string {
 	}
 
 	return sub.StartingCSV
+}
+
+// isCSVVersionAllowed assumes the Versions list has been canonicalized already,
+// usually done by buildResources (via resolveVersionsTemplates and canonicalizeVersions).
+func isCSVVersionAllowed(policy *policyv1beta1.OperatorPolicy, csvName string) bool {
+	if len(policy.Spec.Versions) == 0 {
+		return true
+	}
+
+	for _, version := range policy.Spec.Versions {
+		if version == csvName {
+			return true
+		}
+	}
+
+	if startingCSV := policyStartingCSV(policy); startingCSV != "" && startingCSV == csvName {
+		return true
+	}
+
+	return false
 }
 
 // buildSubscription bootstraps the subscription spec defined in the operator policy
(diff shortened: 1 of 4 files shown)


test/e2e/case38_install_operator_test.go
```diff
@@ -4545,4 +4578,211 @@ var _ = Describe("Testing OperatorPolicy", Label("supports-hosted"), func() {
...
+			By("Waiting for the copied CSV to appear in the target namespace")
+			preFunc()
+			waitForCSVField(ctx, csvName, opPolTestNS, preInstNS, "metadata", "labels", "olm.copiedFrom")
+
+			By("Creating the OperatorPolicy targeting " + opPolTestNS)
+			setupPolicy(opPolYAML, opPolName, parentPolicyName)
+
+			DeferCleanup(func() {
+				KubectlTarget("delete", "ns", preInstNS, "--ignore-not-found", "--wait")
+			})
```
JustinKuli: Nit: this `DeferCleanup` could be right after the namespace is created, and then it's clear that it will be cleaned up, even if one of the `waitForCSVField` checks (or something else) fails unexpectedly.

test/e2e/case38_install_operator_test.go
```diff
@@ -4545,4 +4578,211 @@ var _ = Describe("Testing OperatorPolicy", Label("supports-hosted"), func() {
...
+			By("Creating third OperatorPolicy targeting " + tertiaryNS)
+			setupPolicy(opPolYAML, tertiaryPolName, parentPolicyName, tertiaryNS)
+			utils.EnforceOperatorPolicy(tertiaryPolName, testNamespace)
+
+			checkCompliance(ctx, tertiaryPolName, testNamespace, eventuallyTimeout, policyv1.Compliant)
+
+			// The stale Sub should be cleaned up.
+			Eventually(func(g Gomega, ctx context.Context) {
+				expectNoResources(g, ctx, tertiaryNS, gvrSubscription)
+			}, eventuallyTimeout, 5).WithContext(ctx).Should(Succeed())
```
JustinKuli, line 4780: Can you explain some more about where a stale Subscription might come from? The `cleanupStaleInstallation` function mentions that it might be from "before the existing AllNamespaces installation was detected" but I don't follow how the controller would have missed it... Maybe I'm missing something without playing with it in a cluster.

controllers/operatorpolicy_controller.go
```diff
@@ -405,6 +405,39 @@ func (r *OperatorPolicyReconciler) handleResources(ctx context.Context, policy *
...
+
+		if copiedCSV != nil {
+			condChanged = r.handleExistingInstallation(policy, copiedCSV) || condChanged
+
+			if err := r.updateDeprecationStatus(ctx, policy, desiredSub, copiedCSV, packageManifest); err != nil {
+				return earlyComplianceEvents, condChanged, err
+			}
+
+			changed, err = r.updateMinorChannelUpgradeStatus(policy, desiredSub, packageManifest)
+			condChanged = condChanged || changed
```
JustinKuli, line 429: Can you add a brief comment explaining why (only) these additional details in the status need to be updated?

controllers/operatorpolicy_controller.go
```diff
@@ -922,6 +955,24 @@ func policyStartingCSV(policy *policyv1beta1.OperatorPolicy) string {
 	return sub.StartingCSV
 }
 
+func isCSVVersionAllowed(policy *policyv1beta1.OperatorPolicy, csvName string) bool {
```
JustinKuli, line 962: Can we add a comment here that this should only be used after `resolveVersionsTemplates` and `canonicalizeVersions` have been used to get the Versions list into the nice simple format? It could just say something like "this function assumes the Versions list has been canonicalized already, usually done by `buildResources`"

controllers/operatorpolicy_controller.go
```diff
@@ -2442,6 +2493,73 @@ func (r *OperatorPolicyReconciler) mustnothaveInstallPlan(
...
+		opLog.Info("Deleting stale Subscription from previous reconcile",
+			"subName", foundSub.GetName(), "subNamespace", foundSub.GetNamespace())
+
+		if err := r.TargetClient.Delete(ctx, foundSub); err != nil {
+			return fmt.Errorf("error deleting stale Subscription: %w", err)
+		}
+	}
+
+	// FUTURE: revisit OG cleanup when #489 (OG ownership tracking) is implemented.
+	// Until then, we can't distinguish stale OGs from pre-existing ones.
```
JustinKuli, line 2565: Good plan! Thanks for opening the additional issues!

JustinKuli (changes requested):
This looks really good overall - thanks for the contribution! I think everything is very understandable in this PR, I only requested some additional comments to clarify things for when the topic is less fresh in our minds.

LMK if you need any help rebasing the commit or anything else.

test/e2e/case38_install_operator_test.go
mkolesnik: There are two scenarios:

1. **Race on first reconcile**: The policy is set to enforce and the controller runs before OLM has copied the CSV into the target namespace.
   The copied CSV hasn't appeared yet, so `findCopiedCSV` returns nil and the controller falls through to the normal flow, creating a Subscription (and OG).
   On the next reconcile, the copied CSV is there, the detection fires, and now the Subscription from the previous reconcile is stale.

2. **Upgrade path**: A cluster running the controller without this feature has already created a Subscription in enforce mode.
   After upgrading the controller, the new code detects the existing AllNamespaces installation and the previously created Subscription is now stale.

The tertiary namespace test exercises scenario 1: the OG already exists, the policy is enforced, and we verify the stale Sub gets cleaned up while the pre-existing OG survives.

controllers/operatorpolicy_controller.go
mkolesnik: Will add a comment noting the canonicalization prerequisite.

controllers/operatorpolicy_controller.go
mkolesnik: Will add a comment.

These are the compliance config checks (deprecation, minor channel upgrade) that apply to the CSV regardless of whether the installation is local or existing.
The core conditions (OG, Sub, InstallPlan, etc.) are set by `handleExistingInstallation`, but these two need the packageManifest and desiredSub context to evaluate.

test/e2e/case38_install_operator_test.go
mkolesnik: Good catch.
I'll move the `DeferCleanup` right after the namespace creation so it's clear the cleanup is tied to the resource acquisition, not the end of setup.

JustinKuli (approved):
LGTM!

mkolesnik:
@JustinKuli thanks for the ack, I had to squash the commit because tide refused to merge it with the default rebase strategy.

Nothing changed otherwise, but now it shouldn't complain

JustinKuli (approved):

